PR #35, opened 2019-09-02
Clicking on club card/row tag updates selected tag filters

sigmachirality:
There's a question mark after this on the Trello board so maybe this was a waste of time...

On the bright side, I've setup my local dev environment so I can help out with this from now on if need be!

### DIFF
--- a/components/ClubTableRow.js
+++ b/components/ClubTableRow.js
@@ -40,7 +40,7 @@ class ClubTableRow extends React.Component {
   }
 
   render() {
-    const { club, openModal, updateFavorites, favorite } = this.props
+    const { club, openModal, updateFavorites, favorite, selectedTags, updateTag } = this.props
     const { name, tags } = club
 
     return (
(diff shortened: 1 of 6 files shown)


components/ClubTableRow.js
```diff
@@ -50,7 +50,11 @@ class ClubTableRow extends React.Component {
             <div className="columns is-gapless" style={{ padding: 10 }}>
               <div className="column is-4-desktop is-12-mobile">
                 <b className="is-size-6" style={{ color: CLUBS_GREY }}> {name} </b>
-                <TagGroup tags={tags} />
+                <TagGroup 
+                  tags={tags}
+                  selectedTags={selectedTags}
+                  updateTag={updateTag}
+                  />
```
ccabo1:
```suggestion
                />
```

components/common/TagGroup.js
```diff
@@ -1,14 +1,41 @@
...
-      <BlueTag key={id} className="tag is-rounded has-text-white">
-        {name}
+    tags.map(tag => {
+      const matchedTag = selectedTags.find(({ value }) => value === tag.id)
+      if (matchedTag) {
+        return <SelectedTag
+          key={tag.value}
+          className="tag is-rounded has-text-white"
+          onClick={(e) => {
+            e.stopPropagation()
```
ccabo1: Maybe put a comment as to why we need to stop propagation

sigmachirality: done

PR #37, opened 2019-10-04
Frontend refactor

ccabo1:
• Add prettier
• Make prettier and eslint agree, make eslint rules more similar to basics
• Lint all files
• Fix some issues with keys
• Add some PropTypes
• Convert some class components to functional components 
• Add some hooks where easy to move away from state
• Use destructuring more on objects for more readable code w/ less side effects

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,14 @@
   "plugins": ["react"],
   "rules": {
     "comma-dangle": ["error", "always-multiline"],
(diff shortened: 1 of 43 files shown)


ccabo1:
Will have to rebase on master...

ccabo1:
In the future will keep PR's smaller since this is effectively impossible to review 🌵

ccabo1:
Just skimmed over everything one more time and it looks pretty good; seemed to be running well in dev on my machine

PR #39, opened 2019-10-05
Implement "change modal "add favorites" button to "remove from favorites" when it is a favorite"

sigmachirality:
See Trello card 'change modal "add favorites" button to "remove from favorites" when it is a favorite': the title is exactly what this is.

### DIFF
--- a/renderPage.js
+++ b/renderPage.js
@@ -174,7 +174,7 @@ export function renderListPage(Page) {
             tags={tags}
             closeModal={this.closeModal.bind(this)}
             updateFavorites={this.props.updateFavorites}
-            favorite={favorites.includes(modalClub.id)}
+            favorite={favorites.includes(modalClub.code)}
           />
           <Page
             clubs={clubs}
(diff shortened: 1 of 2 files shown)


renderPage.js
```diff
@@ -174,7 +174,7 @@ export function renderListPage(Page) {
             tags={tags}
             closeModal={this.closeModal.bind(this)}
             updateFavorites={this.props.updateFavorites}
-            favorite={favorites.includes(modalClub.id)}
+            favorite={favorites.includes(modalClub.code)}
```
ccabo1: What is the difference between `code` and `id`? Can you explain why we need this?

ccabo1:
Resolve merge conflicts, answer my q, then we should be gtg

renderPage.js
sigmachirality: the id is a numerical id assigned to each club, the code for each club is something like "pennlabs" or "penn-dems". I think it might be the url slug? Whatever the case, the favorites array contains string codes, not numerical ids.

sigmachirality: we good to yeet now?

ccabo1:
ye

PR #40, opened 2019-10-06
add no favs image, refactor component

ccabo1:


ccabo1:
This is p vanilla, tested it and it works fine; from now forwards, with PR's that change UI, we should include screenshots in the PR

PR #43, opened 2019-10-09
Add some hooks

ccabo1:
* Hooks
* Default Props
* JSX cleanup

### DIFF
--- a/components/ClubModal/Details.js
+++ b/components/ClubModal/Details.js
@@ -50,10 +50,15 @@ const Details = ({ size, applicationRequired, acceptingMembers }) => (
   </DetailsTag>
 )
 
+Details.defaultProps = {
+  applicationRequired: null,
+  acceptingMembers: false,
+}
+
 Details.propTypes = {
   size: PropTypes.string.isRequired,
-  applicationRequired: PropTypes.number.isRequired,
-  acceptingMember: PropTypes.bool.isRequired,
+  applicationRequired: PropTypes.number,
+  acceptingMembers: PropTypes.bool,
 }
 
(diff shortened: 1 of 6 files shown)


components/ClubModal/Details.js
```diff
@@ -50,10 +50,15 @@ const Details = ({ size, applicationRequired, acceptingMembers }) => (
...
+  applicationRequired: null,
+  acceptingMembers: false,
+}
+
 Details.propTypes = {
   size: PropTypes.string.isRequired,
-  applicationRequired: PropTypes.number.isRequired,
-  acceptingMember: PropTypes.bool.isRequired,
+  applicationRequired: PropTypes.number,
+  acceptingMembers: PropTypes.bool,
```
LeoNero: `isAcceptingMembers` is a better name I think

components/DropdownFilter.js
```diff
@@ -121,89 +121,69 @@ const Chevron = () => (
...
+  const [drop, setDrop] = useState(false)
 
-  toggleDrop() {
-    const { drop } = this.state
-    this.setState({ drop: !drop })
-  }
+  const toggleDrop = () => setDrop(!drop)
 
-  isSelected(tag) {
+  const isSelected = tag => {
```
LeoNero: The name of the function could be changed to `getSelected` or something like that.

components/ClubModal/Details.js
ccabo1: This is the way it is stored in the backend to my understanding—good flag which might be worth the refactor but I don't really want to do it in this PR if that's ok

PR #44, opened 2019-10-18
Exit club modal when the escape key is pressed

sigmachirality:


ccabo1:
I think we need to remove the event listener when the modal is closed, else we are building up a lot of them

ccabo1:
Maybe take a look at what we do here: https://github.com/pennlabs/pennbasics/blob/master/src/frontend/components/shared/Modal.js

LeoNero:
I don't think doing things directly on `document` is recommended. But there's been some time since I used React for the last time, so I'm not entirely sure :P

sigmachirality:
rip modal

PR #45, opened 2019-10-18
Format links and icons better on club show page

cx-lu:
Screen Shot 2019-10-20 at 1 23 31 PM


components/ClubPage/SocialLink.js
```diff
@@ -0,0 +1,15 @@
...
+      if (social == 'email') {
+        return (
+            club[item.name]
+          );
+      } else {
+        return (
+            <a href={club[item.name]}>{club.name}</a>
+          );
+      }
+  }
```
ccabo1: There should be a newline at the end of the file. Can you install [prettier](https://prettier.io) on your text editor? It will make it ST we don't run into issues with formatting going forwards

ccabo1:
Let me know if you want help on anything or if something isn't clear!

cx-lu:
Made changes - let me know what you think!

components/ClubPage/SocialLink.js
```diff
@@ -0,0 +1,11 @@
+export default props => {
+  let club = props.club;
```
LeoNero: I think const is better than let here. Also, you can do `const { club, item, type: social } = props` (or `social: type`, I don't remember exactly lol

Also, the if-else can be simplified to
```
if (social == "email") {
    return club[item.name];
  } 
  return <a href={club[item.name]}>{club.name}</a>;
```

LeoNero: Or even better, you can do 
```
export default ({ club, item, social: type }) => {
if (social == "email") {
    return club[item.name];
  } 
 return <a href={club[item.name]}>{club.name}</a>;
}
```

ccabo1:
Hey, in light of #47 I think we'll have to make some changes / port these changes over on top of the new code. @cx-lu lmk what you think or if you need help.

components/ClubPage/SocialLink.js
```diff
@@ -0,0 +1,11 @@
...
+  let item = props.item;
+  let social = props.type;
+
+  if (social == "email") {
```
LeoNero: Oh, also, use "===" in js, not "=="

cx-lu:
Screen Shot 2019-11-03 at 1 04 10 PM

Made changes with feather icons!

PR #46, opened 2019-10-25
Refactor cards

ccabo1:
**Updates**
* Add shade to make selection more intuitive
* Better align margins
* Remove default club image on card view
* Remove unnecessary divs, state, etc.
* Add details at bottom of card
* Kill modal 🍾 (now when you click on the card it goes straight to the club in a new tab)

image

---

**Old stuff (during WIP of this PR)**

image

image

image

image


ccabo1:
Works well on all screen sizes

ccabo1:
Let's hold off on merging until I kill the modal

components/ClubCard.js
```diff
@@ -72,38 +73,39 @@ const CardTitle = s.strong`
...
 `
 
-class ClubCard extends React.Component {
-  constructor(props) {
-    super(props)
-    this.state = {
-      modal: '',
-    }
-  }
+const shorten = desc => {
```
ezwang: I don't know if it applies in this case, but you could try using [text-overflow: ellipsis](https://www.w3schools.com/cssref/css3_pr_text-overflow.asp) to show exactly the amount of text that can fit inside the box.

ccabo1: Issue with the CSS solution is it only works for one line of text—there's no easy way to have like 3 lines of text where the last line ends in a '...' if needed

ezwang: Looks like [css-line-clamp](https://caniuse.com/#feat=css-line-clamp) is a thing but not supported well on all browsers. Ah well, maybe sometime in the future.

PR #50, opened 2019-11-03
show page refactor

cacabo:
image


cacabo:
Yeah there's way more work to be done on this page (@valenciafu and @jtlieb are working on designs) I more so just wanted to clean it up and make sure we were using shared components as much as possible (cc @ezwang)

PR #51, opened 2019-11-15
Org vertical

haoyxc:

PR #55, opened 2019-11-30
OSA Reports

davish:
This PR will have a frontend for demonstrating OSA reporting.

Screen Shot 2019-11-30 at 6 20 49 PM

The current draft allows selecting which columns you want in the output Excel file. To-do is taking advantage of the back-end to get a list of all possible fields instead of hard-coding them into the front-end, and also adding filters.

- [x] Basic functionality for importing/exporting
- [x] Remove hardcoded list of fields
- [ ] Implement category ordering
- [ ] Implement owner/officer name/email checkboxes
- [ ] Implement using report name/description fields on backend
- [x] Fix style issues on smaller screens/mobile
- [x] Fix remaining style issues with `yarn lint`


components/common/Checkbox.js
```diff
@@ -0,0 +1,42 @@
...
+    <CheckboxContainer className={className}>
+      <HiddenCheckbox checked={checked} onChange={onChange} {...props} />
+      <StyledCheckbox onClick={onChange} checked={checked}>
+        <Icon name={checked ? 'check-box-red' : 'box-red'} /> 
+      </StyledCheckbox>
+    </CheckboxContainer>
+  )
+}
+
+export default Checkbox
```
cacabo:
```suggestion
export default Checkbox

```

components/common/Sidebar.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+`
+
+export const Sidebar = ({ children }) => (
+  <Wrapper>
+    <Content>
+      { children }
+    </Content>
+  </Wrapper>
+)
```
cacabo:
```suggestion
)

```

package-lock.json
```diff
@@ -0,0 +1,14671 @@
+{
```
cacabo: We need to either go all in on yarn on npm. seems like it will be best if we move everything to npm, but that shouldn't happen in this PR.

pages/reports.js
```diff
@@ -0,0 +1,210 @@
...
+    'Contact Information': [
+      'Owner name(s)',
+      'Owner email(s)',
+      'Listserv',
+      'Officer name(s)',
+      'Officer email(s)'
+    ]
+  }
+
+  const nameToCode = {
```
cacabo: we should probably modularize this more. Have a folder for components relating to `reports` then we can have a `constants.js` file where we import these things. IMO having them in the file doesn't improve readability and increases the amount ppl have to scroll.

cacabo: Also this constant is presumably useful / already used elsewhere? not sure

pages/reports.js
```diff
@@ -0,0 +1,210 @@
...
+
+  const query = {
+    format: 'xlsx',
+    fields: Object.keys(includedFields)
+                  .filter(field => includedFields[field])
+                  .map(name => nameToCode[name])
+                  .filter(e => e !== undefined)
+  }
+
+  console.log(query)
```
cacabo:
```suggestion
```

pages/reports.js
```diff
@@ -0,0 +1,210 @@
...
+
+  const generateCheckboxGroup = (groupName, fields) => {
+    return (
+      <div style={{flexBasis: '50%', flexShrink: 0}}>
+        <GroupLabel key={groupName} className="subtitle is-4" style={{ color: CLUBS_GREY }}>{groupName}</GroupLabel>
+        { fields.map((field, idx) => (
```
cacabo: Make sure you're using `prettier` for code formatting

pages/reports.js
```diff
@@ -0,0 +1,210 @@
...
+      <Container>
+        <WideContainer>
+          <h1 className="title" style={{ color: CLUBS_GREY }}>
+            Create a new report 
+          </h1>
+          <div className="box">
+            <h3 className="title is-4" style={{ color: CLUBS_GREY }}>
+              Report Details
+            </h3>
+            <div>
```
cacabo: indenting or something is off here

package-lock.json
ezwang: Let's switch over to npm at a later separate PR, removed the `package-lock.json` for now.

pages/reports.js
ezwang: This has been (partially) changed to be loaded from the API server, so that when we update backend fields we don't have to remember to update frontend too.

The groups of fields constant has not been changed yet.

ezwang:
Replaced by https://github.com/pennlabs/penn-clubs/pull/51.

ezwang:
Changes are a bit less fancy than the original design, but everything works and is usable.

image

frontend/pages/reports.js
```diff
@@ -0,0 +1,262 @@
...
+            <CheckboxLabel htmlFor={field}>{field}</CheckboxLabel>
+          </div>
+        ))}
+      </div>
+    )
+  }
+
+  return (
+    <Container>
+      <WideContainer>
```
cacabo: `WideContainer` already has a `Container` inside of it. If it alone is too wide, use `MediumContainer` (I think is what it's called) or just use `Row` and `Col` directly

frontend/pages/reports.js
```diff
@@ -0,0 +1,262 @@
...
+                          const params = JSON.parse(report.parameters)
+                          const paramString = Object.keys(params).reduce(
+                            (acc, cur) =>
+                              `${acc}&${encodeURIComponent(
+                                cur
+                              )}=${encodeURIComponent(params[cur])}`,
+                            ''
+                          )
+                          return `${API_BASE_URL}/clubs/?existing=true${paramString}`
+                        })()}
```
cacabo: would be good to split things like this into helper functions. I prefer render methods being exclusively rendering information + calling helper functions.

frontend/pages/reports.js
```diff
@@ -0,0 +1,259 @@
+import s from 'styled-components'
+import { useEffect, useState } from 'react'
+import renderPage from '../renderPage.js'
+import { Icon, Flex, Empty } from '../components/common'
+import { doApiRequest, API_BASE_URL } from '../utils'
+import Checkbox, { CheckboxLabel } from '../components/common/Checkbox'
+import { Container, WideContainer } from '../components/common/Container'
```
cacabo: Ideally we export everything in `common/index.js` so we can just say
```js
import { Checkbox, CheckboxLabel, Container, ... } from '../components/common'
```

frontend/components/common/Checkbox.js
```diff
@@ -0,0 +1,47 @@
...
+    <CheckboxContainer className={className}>
+      <HiddenCheckbox checked={checked} onChange={onChange} {...props} />
+      <StyledCheckbox onClick={onChange} checked={checked}>
+        <Icon alt={checked ? 'checked' : 'unchecked'} name={checked ? 'check-box' : 'box'} />
+      </StyledCheckbox>
+    </CheckboxContainer>
+  )
+}
+
+export default Checkbox
```
cacabo: Would prefer all named exports (export const ...) for all things in the common directory

frontend/pages/reports.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+`
+
+const SelectedManager = ({ value, onClick }) => (
+  <HoverListElement>
+    {value}
+
+    <span style={{ marginLeft: '1em' }} onClick={onClick}>
+      X
```
cacabo:
```suggestion
      &times;
```

PR #56, opened 2019-12-02
Render Club Page Server Side

davish:
This PR moves the fetch for club info on the club detail page into the `getInitialProps` call. Now, nextjs will render all club details server-side and send the page to the client. This is preferable for a few reasons:
1. SEO -- web crawlers will now get all the HTML for the page without having to run JS (which they don't do very well)
2. Speed -- before, the page needed to load + render before the backend was queried for club data. Now, the browser gets the HTML right away in the original request.

Relevant documentation for reading more: https://nextjs.org/docs#fetching-data-and-component-lifecycle

### DIFF
--- a/components/TabView.js
+++ b/components/TabView.js
@@ -44,7 +44,8 @@ const Tabs = s.div``
 
 const TabView = ({ tabs, tabClassName, background }) => {
   // the server side rendering does not have a window object
-  const hashString = typeof window !== 'undefined' ? window.location.hash.substring(1) : null
+  const hashString =
+    typeof window !== 'undefined' ? window.location.hash.substring(1) : null
   const [currentTab, setCurrentTab] = useState(hashString || tabs[0].name)
 
   const getTabContent = () =>
--- a/pages/club/[club]/index.js
+++ b/pages/club/[club]/index.js
@@ -1,5 +1,5 @@
 import s from 'styled-components'
-
+import { useState, useEffect } from 'react'
 import renderPage from '../../../renderPage.js'
 import { doApiRequest } from '../../../utils'
 import Tabs from '../../../components/ClubPage/Tabs'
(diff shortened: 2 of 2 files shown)


pages/club.js
```diff
@@ -79,7 +70,9 @@ const Club = ({ query, userInfo, favorites, updateFavorites }) => {
 
 Club.getInitialProps = async props => {
   const { query } = props
-  return { query }
+  const resp = await doApiRequest(`/clubs/${query.club}/?format=json`)
```
cacabo: Not that we have it right now, but should error handling be happening here? what if we don't get a response / we get a 500?

davish: I think it'll throw an exception, which we won't catch and a 500 error (probably an ugly one) will be shown – I'll add a try/catch so that we show our error page.

ezwang:
The issue with this change is that the client's authentication credentials are not passed into the request with the API server. Certain features, like showing emails to only people who are in the club, will not work anymore.

You will need to figure out a way to pass the client's authentication credentials to the backend server if you want to do this. @davish

davish:
@ezwang oh that's a good point i hadn't realized... do you think it would work if we fetched the club on the server-side once, and then again on the client side with proper user credentials? So that the authenticated details will be filled in after a small wait, but we still get the speed improvement and SEO benefits if users aren't authenticated / aren't running javascript

davish:
Just added a change so there's a client-side load. I'm pretty sure React handles rendering in such a way that it won't re-render things it doesn't need to, so the only cost is the second HTTP request.

cacabo:
@davish is this gtg?

cacabo:
@ezwang  or @sigmachirality can you guys resolve conflicts and get this merged? lgtm I'm just moving around a bit today and don't have time for resolving

cacabo:
@davish @sigmachirality I might have misinterpreted your work but got it working on my end and did the rebase on top of master. Pls review my changes to make sure this is right.

ezwang:
@cacabo Fixed the merging issues, now only 2 files should be edited instead of 91.

PR #60, opened 2020-01-02
Strip Social Urls to Decrease Their Length

sigmachirality:
Uses a regex to strip "https" and "www" from social urls.

Maybe we could format social urls in specific ways (for example, displaying twitter page links as @twitterusername) instead?

### DIFF
--- a/components/ClubPage/SocialLink.js
+++ b/components/ClubPage/SocialLink.js
@@ -1,5 +1,5 @@
 const stripUrl = url => {
-  return url.replace(/^(?:https?:\/\/)?(?:www\.)?/i, '')
+  return url.replace(/^(?:https?:\/\/)?(?:www\.)?/i, '').replace(/\/$/, '')
 }
 
 const SocialLink = ({ club, item, type }) => {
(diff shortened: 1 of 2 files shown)


components/ClubPage/SocialLink.js
```diff
@@ -1,14 +1,17 @@
+const stripUrl = url => {
+  return url.replace(/^(?:https?:\/\/)?(?:www\.)?/i, '')
```
cacabo: what about `http`?

cacabo: also you probably want to make sure that this is at the start of the url if there's an edge case where http is somewhere later like in a query param? might be better off not using a regex

ezwang: The question mark makes the `s` optional (matches both `http` and `https`). The `^` ensures that this regex only matches the start of the url.

sigmachirality: Screenshot from 2020-01-02 17-42-25

I'm no regex expert but it seems to work for me

sigmachirality:
going to yeet

PR #61, opened 2020-01-02
Various improvements

ezwang:
- Don't merge until backend `profile` branch is merged first.
- Implement subscription button (needs styling improvements)
    - image
    - image (old)
- Add placeholder for empty profile fields
    - image
- Fix button icons with correct coloring and spacing
    - image
- Implement allowing club owners to specify schools/majors to target
    - image
- Various small bug fixes
- Implement better invitation page (with public toggling)
    - image
- Fix club link appearance on settings page
    -  image

PR #63, opened 2020-01-05
Bookmarks Persist After Toggle in Profile

sigmachirality:
this PR does the title and also removes localstorage usage since users can't see bookmarks unless they're logged in anyway.

### DIFF
--- a/components/ClubPage/Header.js
+++ b/components/ClubPage/Header.js
@@ -42,7 +42,7 @@ const Header = ({
     (inClub && inClub.role <= ROLE_OFFICER) ||
     (userInfo && userInfo.is_superuser)
 
-  const [favCount] = useState(club.favorite_count || 0)
+  const favCount = club.favorite_count || 0
 
   const { active, code, name, tags } = club
 
--- a/components/ClubPage/Header.js
+++ b/components/ClubPage/Header.js
@@ -42,7 +42,7 @@ const Header = ({
     (inClub && inClub.role <= ROLE_OFFICER) ||
     (userInfo && userInfo.is_superuser)
 
-  const favCount = club.favorite_count || 0
+  const [favCount, setFavCount] = useState(club.favorite_count || 0)
 
   const { active, code, name, tags } = club
 
--- a/components/ClubPage/Header.js
+++ b/components/ClubPage/Header.js
@@ -42,7 +42,7 @@ const Header = ({
     (inClub && inClub.role <= ROLE_OFFICER) ||
     (userInfo && userInfo.is_superuser)
 
-  const [favCount, setFavCount] = useState(club.favorite_count || 0)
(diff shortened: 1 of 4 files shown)


components/ClubPage/Header.js
```diff
@@ -27,7 +27,7 @@ const Header = ({ club, userInfo, favorites, style, updateFavorites }) => {
     (inClub && inClub.role <= ROLE_OFFICER) ||
     (userInfo && userInfo.is_superuser)
 
-  const [favCount, setFavCount] = useState(club.favorite_count || 0)
+  const [favCount] = useState(club.favorite_count || 0)
```
cacabo: If not using the setter, why not just have
```suggestion
  const favCount = club.favorite_count || 0
```

components/Settings/FavoritesTab.js
```diff
@@ -1,9 +1,17 @@
+import { useState } from 'react'
 import ClubTableRow from '../ClubTableRow'
 
-export default (props) => {
-  const { updateFavorites, favorites, clubs } = props
-
+export default ({ clubs, favorites, updateFavorites }) => {
+  const [table, setTable] = useState(() => {
+    const ret = {}
+    favorites.forEach(favorite => { ret[favorite] = true })
```
cacabo: I think prettier would recommend this but I might be wrong
```suggestion
    favorites.forEach(favorite => {
      ret[favorite] = true
    })
```

components/ClubPage/Header.js
sigmachirality: I think this lint warning is being handled in another PR

ezwang: I believe that #61 might use the setter, we can handle this during the merge process.

components/Settings/FavoritesTab.js
```diff
@@ -1,9 +1,19 @@
...
-
+export default ({ keyword, clubs, favorites, updateFavorites }) => {
+  const [table, setTable] = useState(() => {
+    const ret = {}
+    favorites.forEach(favorite => {
+      ret[favorite] = true
+    })
+    return ret
+  })
   const findClub = clubs ? code => clubs.find(club => club.code === code) : () => {}
```
cacabo: Feel like there's a better way to do this—even just splitting into lines and adding a comment that "if there is no code, there is no club to be found" would be a lot more readable than this ternary operator

cacabo:
Other than that one comment this lgtm

components/Settings/FavoritesTab.js
sigmachirality: rewritten in the commit "rewrite findClub to be clearer"

sigmachirality:
@ezwang I fixed the bug, was using the list of favorites updated by the user instead of the list of favorites cached in state to check whether to display the placeholder text

PR #64, opened 2020-01-05
Logout Button

sigmachirality:
Right now it's in the navbar, but it can be put pretty much anywhere

where put tho

### DIFF
--- a/components/Header/Links.js
+++ b/components/Header/Links.js
@@ -1,7 +1,7 @@
 import s from 'styled-components'
 
 import { Icon } from '../common'
-import { LOGIN_URL, LOGOUT_URL } from '../../utils'
+import { LOGIN_URL } from '../../utils'
 import { mediaMaxWidth, MD } from '../../constants/measurements'
 import { MEDIUM_GRAY, DARK_GRAY, BORDER } from '../../constants/colors'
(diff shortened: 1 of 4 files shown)


cacabo:
From slack:

> Daniel 4  1 hour ago
> where should the logout button go?
> 
> valencia:mushroom:  40 minutes ago
> good question. could put a power button icon in the nav bar (right side of name)?
> 
> valencia:mushroom:  39 minutes ago
> or to keep the navbar uncluttered, have a logout button on the profile page
> 
> valencia:mushroom:  39 minutes ago
> @Cameron ?
> 
> Daniel 4  32 minutes ago
> I'm going to put it in the navbar for now, it's not hard to move it around
> :+1:
> 1
> 
> Cameron:gloves:  2 minutes ago
> don't think it's needed in the navbar
> 
> Cameron:gloves:  1 minute ago
> b/c it's an action which is so rarely carried out and which we don't necessarily want users to do...
> 
> Cameron:gloves:  1 minute ago
> I think the convention for sites like this is putting it at the bottom of settings...

sigmachirality:
Screenshot from 2020-01-05 18-18-15

cacabo:
Yep lgtm

PR #66, opened 2020-01-07
Add Login Prompt Modal

sigmachirality:
This modal should be opened whenever an authenticated user attempts and action that requires authentication, instead of the action.

To do that, this PR adds a higher order function, checkAuth, that takes in a function and then its parameters. For example, console.log("meme") would become checkAuth(console.log, "meme").

The Modal looks something like this:
Screenshot from 2020-01-07 20-43-24


components/LoginModal.js
```diff
@@ -0,0 +1,99 @@
+import s from 'styled-components'
+
+import { Icon } from './common/Icon'
+import { ALLBIRDS_GRAY, LIGHT_GRAY, MEDIUM_GRAY, DARK_GRAY } from '../constants/colors'
+import { BORDER_RADIUS_LG, MD, SM, mediaMaxWidth } from '../constants/measurements'
+import { LOGIN_URL } from '../utils'
+
+const Logo = s.img`
+  width: 100px
+  transform: scale(1);
```
ezwang: What is this for?

components/LoginModal.js
```diff
@@ -0,0 +1,99 @@
...
+const CloseModalIcon = s(Icon)`
+  position: absolute;
+  right: 20px;
+  top: 20px;
+  cursor: pointer;
+  color: ${LIGHT_GRAY};
+
+  &:hover {
+    color: ${MEDIUM_GRAY};
+  }
```
ezwang: Note that this currently won't actually change the color of the icons, the current way we handle svgs are resistant to simple css `color` rules. :(

components/LoginModal.js
sigmachirality: nothing really, I'll take it out.

components/LoginModal.js
sigmachirality: I'll remove it then, this is probably left over from the font awesome days.

cacabo:
Do we not already have a modal component in this codebase? If not, what you have looks pretty good, but the modal itself should be its own component in the shared directory. You can take some things (like animations) from the Penn Basics codebase.

https://github.com/pennlabs/pennbasics/blob/master/src/frontend/components/shared/Modal.js

Would like to see this changed made before we merge this puppy in, but this looks good otherwise

components/LoginModal.js
```diff
@@ -0,0 +1,94 @@
+import s from 'styled-components'
+
+import { Icon } from './common/Icon'
+import { ALLBIRDS_GRAY, LIGHT_GRAY, DARK_GRAY } from '../constants/colors'
+import { BORDER_RADIUS_LG, MD, SM, mediaMaxWidth } from '../constants/measurements'
+import { LOGIN_URL } from '../utils'
+
+const Logo = s.img`
+  width: 100px
```
cacabo:
```suggestion
  width: 100px;
```

sigmachirality:
> Do we not already have a modal component in this codebase? If not, what you have looks pretty good, but the modal itself should be its own component in the shared directory. You can take some things (like animations) from the Penn Basics codebase.
> 
> https://github.com/pennlabs/pennbasics/blob/master/src/frontend/components/shared/Modal.js
> 
> Would like to see this changed made before we merge this puppy in, but this looks good otherwise

There used to be a Modal in the codebase back when clubs were shown in a Modal, but it was taken out when we switched to the current way of displaying clubs. I actually went back in the commit history and tried to extract the old Modal for usage here. I'll abstract the Modal out into a separate component.

cacabo:
Cool sounds good pls re-request review when that's done. Def take a look at the basics example because it has an animation + some accessibility concerns addressed (like pressing the escape key (I think))

sigmachirality:
@ezwang The bug where clicking the bookmarks button when not logged decreases the count no matter what is back, I think it has to do with how checkAuth returns false if the Modal is opened. What should checkAuth return if the modal is opened? I don't want to run the function passed in if the user isn't authenticated, but I'm not sure how handling checkAuth would work here otherwise.

ezwang:
@sigmachirality Bug should be fixed! Error was bad logic, probably introduced accidentally during the merging process.

frontend/components/LoginModal.js
```diff
@@ -0,0 +1,45 @@
+import { useState, useEffect } from 'react'
+import s from 'styled-components'
+
+import Modal from './common/Modal'
```
cacabo: can we make this a named export and export it from `common/index.js` so we can say `import { Modal } from './common'`?

frontend/components/LoginModal.js
```diff
@@ -0,0 +1,45 @@
...
+
+export default props => {
+  const [newlyMounted, setNewlyMounted] = useState(true)
+  useEffect(() => {
+    newlyMounted && setNewlyMounted(false)
+  })
+  return <Modal {...props} >
+    {
+      newlyMounted ? (
+        <div className="loader is-loading"></div>
```
cacabo: I think we have a `<Loading />` component in `common`

frontend/components/common/Modal.js
```diff
@@ -0,0 +1,125 @@
...
+  top: 0;
+  left: 0;
+  height: 100%;
+  width: 100%;
+  overflow-x: hidden;
+  overflow-y: auto;
+  z-index: 1002;
+  text-align: center;
+  animation-name: ${({ show }) => (show ? fadeIn : fadeOut)};
+  animation-duration: ${LONG_ANIMATION_DURATION};
```
cacabo: Just want to make sure that we don't need `ms` here—do the animations work locally for you?

frontend/components/common/Modal.js
```diff
@@ -0,0 +1,125 @@
...
+const ModalCard = s.div`
+  border-radius: ${BORDER_RADIUS_LG};
+  border: 0 !important;
+  box-shadow: none !important;
+  height: auto;
+  width: 35%;
+  animation-name: ${({ show }) => (show ? fadeIn : fadeOut)};
+  animation-duration: ${LONG_ANIMATION_DURATION};
+
+  ${mediaMaxWidth(MD)} {
```
cacabo: Also just want to make sure did you check that the modal is mobile responsive

frontend/components/LoginModal.js
sigmachirality: Yup!

frontend/components/common/Modal.js
sigmachirality: Yes, I did!

ezwang:
There's also an [AuthPrompt](https://github.com/pennlabs/pennclubs/blob/master/frontend/components/common/AuthPrompt.js) class that you may be able to take advantage of, it's primarily designed for using the whole page instead of being a modal so there may not be a lot of overlap.

frontend/components/common/Modal.js
sigmachirality: Added ms.

sigmachirality:
AuthPrompt doesn't seem to overlap here too much, but I'll keep it in mind for things in the future!

PR #69, opened 2020-01-13
Add Close Icon When Filtering by Tags on Mobile

sigmachirality:
Looks like this:
Screenshot from 2020-01-13 02-17-51


sigmachirality:
> LGTM! Consider only making the close button show up when there is something to close? May be less confusing to the user on what it does that way.

Done.

PR #70, opened 2020-01-13
Enable Submit Button Only If Changes Were Made

sigmachirality:
The submit button in the Form component is disabled until a change is made.

### DIFF
--- a/frontend/components/Form.js
+++ b/frontend/components/Form.js
@@ -349,6 +349,15 @@ class Form extends Component {
     this.state.edited || this.setState({ edited: true })
   }
 
+  handleSubmit() {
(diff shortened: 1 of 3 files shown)


ezwang:
Note that the `ProfileForm` also implements this feature already, we should make sure that we don't duplicate code.

You may also want to call `checkChange` inside `onChange` and register the `onChange` handler everywhere to prevent code duplication! The `onChange` handler just allows parent components to pass a function in case they want to perform some event when the child changes the form (could be useful later for auto-saving forms or something).

sigmachirality:
I think I'll rewrite the Form component and components using the Form component to reduce code duplication.

PR #72, opened 2020-01-13
Improve import script and club name matching

ezwang:
- Add shared club name matching function to aggressively fuzzy search for a matching club.
- Improve GO@P import script behavior when merging clubs.

### DIFF
--- a/backend/clubs/management/commands/goap_import.py
+++ b/backend/clubs/management/commands/goap_import.py
@@ -30,6 +30,8 @@ class Command(BaseCommand):
     def handle(self, *args, **kwargs):
         self.count = 1
         self.club_count = 0
+        self.create_count = 0
+        self.update_count = 0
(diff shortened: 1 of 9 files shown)


backend/clubs/management/commands/goap_import.py
```diff
@@ -43,7 +53,11 @@ def handle(self, *args, **kwargs):
         if self.skip_tags:
             self.stdout.write("Skipping tag imports...")
         self.process_url(self.START_URL)
-        self.stdout.write("Imported {} clubs!".format(self.club_count))
+        self.stdout.write(
+            "Imported {} clubs! {} created, {} updated".format(
```
sigmachirality: Now that we're on 3.6 and up, should we start using format strings for string interpolation?

ezwang: I like that, I'll switch it over!

PR #73, opened 2020-01-13
Merge domains

ArmaanT:
This PR sets up django and next so that they run on the same domain. This reduces the django settings complexity and makes our set up more secure.

For frontend development, you need to set `API_BASE_URL` to http://localhost:3000/api and `LABS_REDIRECT_URI` to http://localhost:3000/api/accounts/callback/

Backend only development doesn't need to set any environment variables.

@ezwang Definitely take a look at the frontend directory and make any changes you need to. @davish can help answer questions for how everything is set up.

### DIFF
--- a/backend/pennclubs/settings/base.py
+++ b/backend/pennclubs/settings/base.py
@@ -131,7 +131,7 @@ STATIC_ROOT = os.path.join(BASE_DIR, "static")
 
 PLATFORM_ACCOUNTS = {
     "REDIRECT_URI": os.environ.get(
-        "LABS_REDIRECT_URI", "http://localhost:8000/api/accounts/callback/"
+        "LABS_REDIRECT_URI", "http://localhost:3000/api/accounts/callback/"
     ),
     "CLIENT_ID": "clientid",
     "CLIENT_SECRET": "supersecretclientsecret",
--- a/backend/pennclubs/settings/base.py
+++ b/backend/pennclubs/settings/base.py
@@ -15,8 +15,7 @@ import os
 import dj_database_url
 
 
-FRONTEND_DOMAIN = os.environ.get("FRONTEND_DOMAIN", "example.com")
-BACKEND_DOMAIN = os.environ.get("BACKEND_DOMAIN", "api.example.com")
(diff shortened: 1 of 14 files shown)


frontend/utils.js
```diff
@@ -29,9 +29,9 @@ function removeEndingSlash(val) {
   return val
 }
 
-export const API_BASE_URL =
-  removeEndingSlash(getConfig().publicRuntimeConfig.API_BASE_URL) ||
-  'https://pennclubs.com/api'
+export const DEFAULT_SITE_ORIGIN = process.env.NODE_ENV === 'production' ? 'https://pennclubs.com' : `http://localhost:${process.env.PORT || 3000}`
```
ArmaanT: Can we use the `DOMAIN` env variable here? It will be set to a domain without the http(s) ex. `pennclubs.com`

frontend/utils.js
```diff
@@ -29,9 +29,9 @@ function removeEndingSlash(val) {
...
+export const SITE_ORIGIN = removeEndingSlash(getConfig().publicRuntimeConfig.API_BASE_URL) || DEFAULT_SITE_ORIGIN
```
ArmaanT: Do we still need to check `API_BASE_URL` with the above changes?

ezwang: We don't, do you think we should just remove it?

ArmaanT: I think we could. We would lose the ability to set an arbitrary url for the backend, but I don't know if we need to

backend/pennclubs/settings/base.py
```diff
@@ -131,7 +131,7 @@
 
 PLATFORM_ACCOUNTS = {
     "REDIRECT_URI": os.environ.get(
-        "LABS_REDIRECT_URI", "http://localhost:8000/api/accounts/callback/"
+        "LABS_REDIRECT_URI", "http://localhost:3000/api/accounts/callback/"
```
ArmaanT: So setting the port to 3000 will break the login when doing backend-only development, but if you set it to 8000 it will break frontend dev by default. Not sure what we should go with

PR #74, opened 2020-01-13
Show Badges on Club Page

ezwang:
- Show umbrella club badges on club page.
- Invert given tag color to ensure text visibility.
  - May look better to play around with `lighten` and `darken` depending on the brightness value of the color.

image

image

### DIFF
--- a/frontend/components/common/TagGroup.js
+++ b/frontend/components/common/TagGroup.js
@@ -1,17 +1,31 @@
 import PropTypes from 'prop-types'
-import { BlueTag } from './Tags'
+import { BlueTag, Tag } from './Tags'
 
 export const TagGroup = ({ tags = [] }) => {
   if (!tags || !tags.length) return null
+
   return <>
-    {tags.map(tag => (
-      <BlueTag
-        key={`${tag.id}-${tag.value}`}
-        className="tag is-rounded has-text-white"
-      >
-        {tag.name}
-      </BlueTag>
-    ))}
+    {tags.map(tag => {
+      if (tag.color) {
+        return (
+          <Tag
+            key={`${tag.id}-${tag.value}`}
+            color={tag.color}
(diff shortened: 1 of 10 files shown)


frontend/components/common/Tags.js
```diff
@@ -11,8 +12,15 @@ import {
...
   margin: 0 4px 4px 0;
-  fontSize: .7em;
   font-weight: 600;
+  ${({ color }) =>
+    color &&
+    `
+    background-color: #${color} !important;
+    color: ${Color(`#${color}`)
+      .negate()
+      .hex()} !important;
```
cacabo: Yeah I think it would be good to use lighten and darken—can you maybe give a screenshot example of when the color is not neutral (in which case negating doesn't really do anything). As I can see, this will give us like orange text on blue background or something. If we use lighten and darken and do it enough ST it is an accessible contrast between the two colors, I think that will be more in line with how our other tags are styled.

frontend/components/common/TagGroup.js
```diff
@@ -40,14 +52,14 @@ export const TagGroup = ({ tags = [], selectedTags = [], updateTag }) => {
             updateTag(
               {
                 value: tag.id,
-                label: tag.name,
+                label: tag.name || tag.label,
                 name: 'Type',
               },
               'Type'
             )
```
cacabo: Do we need updateTag? Did we make a mistake with https://github.com/pennlabs/pennclubs/pull/76?

ezwang: We actually don't need updateTag here, I just added the extra statement because it seemed like a reasonable alternative in case thee backend uses `label` instead of `name`. I will remove this.

frontend/components/common/Tags.js
ezwang: Changed!

PR #77, opened 2020-01-14
[HACKY] Confirm/Prompt If User Has Unsaved Changes + Routes Constants

sigmachirality:
You may have noticed that I throw an Error in this code. You probably don't like that. Trust me, I don't either, but according to this [PR](https://github.com/zeit/next.js/issues/2476), Next's developers don't intend on implementing an API to abort route changes ever even though react-router has this feature, meaning that I resorted to using this solution which I also found in the PR.

This PR also converts some <a> components to <Link> components in order to better take advantage of Next's built in code splitting/single page webapp optimizations. Additionally, this PR adds some new route constants to use instead of hardcoding routes in Links.


frontend/pages/_error.js
```diff
@@ -26,9 +27,11 @@ const Error = ({ statusCode = 500, message = 'Something went wrong' }) => (
       <p className="is-size-5" style={{ marginBottom: '1rem' }}>
         {message}
       </p>
-      <a className="button is-medium is-primary" href={HOME_ROUTE}>
-        Back to home
-      </a>
+      <Link href={HOME_ROUTE}>
```
ezwang: The link above uses `href="/"`, while this Link uses `HOME_ROUTE`, which is equivalent to `/`. We should standardize on either not using the constant or using it.

ezwang:
One possible idea is to compare the user entered values with the `defaults` to determine if a field is actually edited. The user could have accidentally changed something and then changed it back, which should not be counted as having edits.

frontend/pages/_error.js
sigmachirality: I'm for replacing everything with HOME_ROUTE, since we could decide to change what route is the landing page in the future.

sigmachirality:
> compare the user entered values with the `defaults`
I was worried about the performance cost, but this would work as long as forms don't get too big.

ezwang:
> > compare the user entered values with the `defaults`
> 
> I was worried about the performance cost, but this would work as long as forms don't get too big.

This should be `O(n)` on the number of fields, and will only be checked when you navigate away from the page. It should always take no longer than actually submitting the form.

sigmachirality:
>This should be O(n) on the number of fields, and will only be checked when you navigate away from > the page. It should always take no longer than actually submitting the form.

I was thinking about this.state.edited, which is also used to determine if the submit button should be rendered. Right now this PR uses this.state.edited, but I feel like if we switch to doing this comparison for the warning that we should also do it for the submit button, in which case the cost would be incurred every time the component is rendered.

frontend/components/Form.js
```diff
@@ -29,110 +33,114 @@ class Form extends Component {
...
+    const { router } = Router
+    if (edited && !confirm(UNSAVED_MESSAGE)) {
+      router.abortComponentLoad()
+      router.events.emit('routeChangeError')
+      throw 'Abort link navigation - ignore this error.' // eslint-disable-line
+    }
+  }
+
+  confirmExit(e) {
+    if (this.state.edited) {
```
cacabo:
```suggestion
    const { edited } = this.state
    if (edited) {
```

frontend/components/Form.js
```diff
@@ -152,14 +160,24 @@ class Form extends Component {
...
 
+    const {
+      [`field-${name}`]: value,
+      [`editorState-${name}`]: editorState,
+    } = this.state
+
+    const isHorizontal =
+      typeof this.props.isHorizontal !== 'undefined'
+        ? this.props.isHorizontal
+        : true
```
cacabo: Why can't we just have like
```js
const isHorizontal = Boolean(this.props.isHorizontal)
```

frontend/components/Form.js
```diff
@@ -366,7 +380,7 @@ class Form extends Component {
     // other value, it resolves with that value.
     const { onSubmit } = this.props
     onSubmit &&
```
cacabo: Would prefer return early:
```js
if (!onSubmit) return
return Promise...
```

sigmachirality:
@ezwang Yeah, I think we can do stuff with comparing the form's current value and the default values later, for now this.state.edited works well enough.

frontend/components/Form.js
sigmachirality: actually why don't we just use the prop itself? Seems like it should be a boolean

sigmachirality: I tried destructuring the prop and the suggestion above, both break the form for some reason

sigmachirality: fixed, used default arg in destructure to get the same behavior

PR #81, opened 2020-01-16
Recruitment events

cphalen:
New functionality for Events.

On the `Membership` tab for in `Edit Club` you can create and delete new events which are then visible to the public on the clubs page.

While right now this is only used for recruitment events, we would just have to add new `type` choices in `models.py` to allow for other types of events to be made.


frontend/pages/club/[club]/edit.js
```diff
@@ -262,6 +272,39 @@ class ClubForm extends Component {
...
+        resp.json().then(err => {
+          this.notify(formatResponse(err))
+        })
+      }
+    })
+  }
+
+  deleteEvent(id) {
+    console.log(this.state.club.code)
+    console.log(`/clubs/${this.state.club.code}/events/${id}`)
```
Comment (ezwang) [inline frontend/pages/club/[club]/edit.js]: Remove all of these `console.log`s when you're ready to merge! Run `npm run lint --fix` to check for style issues similar to this one.

backend/clubs/migrations/0014_event_type.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.0.2 on 2020-01-15 00:02
```
ezwang: After you're ready to merge, merge all of these migrations into one! You can do so by deleting all of them and running `./manage.py makemigrations` again.

sigmachirality:
iirc we're switching to npm, so remove the yarn.lock before merging this PR?

davish:
Just a comment, but seems as though the git history got messed up at some point -- looks like it's counting all the commits since this branch diverged as part of this branch. Would recommend merging in master one more time to see if that fixes it -- it normally does.

frontend/components/ClubPage/Events.js
```diff
@@ -0,0 +1,53 @@
+import { Icon, Text, StrongText, Card } from '../common'
+import { useState, useEffect } from 'react'
+
+const Events = props => {
```
cacabo: Would destructure props to start out
```suggestion
const Events =({ data }) => {
```

frontend/components/ClubPage/Events.js
```diff
@@ -0,0 +1,53 @@
...
+  const data = props.data
+
+  const iconStyles = {
+    float: "left",
+    marginRight: "7px"
+  }
```
cacabo: Generally don't want float, would be good if we could do the same thing with flexbox or `margin: auto`

frontend/components/ClubPage/Events.js
```diff
@@ -0,0 +1,53 @@
...
+
+  const bigStyles = {
+    fontSize: "13px",
+    fontWeight: "bold"
+  }
+
+  const smallStyles = {
+    fontSize: "13px"
+  }
```
cacabo: Also would be good to use `styled-components` for these styles. We also tend to define font sizes in terms of `rem` an normally a multiple of `0.2rem`. Maybe something like a font size of `0.8rem` would work in this case?

frontend/components/ClubPage/Events.js
```diff
@@ -0,0 +1,53 @@
...
+
+  if (!data || !data.length) {
+    return <></>
```
cacabo:
```suggestion
    return null
```

frontend/components/ClubPage/Events.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return (
+    <Card bordered style={{ marginBottom: '1rem' }}>
+      <div style={{marginLeft: '7px'}}>
```
cacabo: What does this div do?

frontend/components/ClubPage/Events.js
```diff
@@ -0,0 +1,53 @@
...
+            <div key={index} style={{ marginBottom: '0.5rem' }}>
+              <Icon name='calendar' style={iconStyles} size="32px" alt='Calendar icon' />
+              <p style={bigStyles}>
+                {new Intl.DateTimeFormat("en-US", {
+                  year: "numeric",
+                  month: "long",
+                  day: "2-digit",
+                  hour: "numeric",
+                  minute: "numeric"
+                }).format(new Date(entry.start_time))} | {entry.location}
```
cacabo: Can you put this logic higher up in the component? Ideally the render part of the function is just taking in stuff we've computed and rendering it

cacabo: Also would be more readable since this could be a variable name like `eventTime`

frontend/components/ClubPage/Events.js
```diff
@@ -0,0 +1,53 @@
...
+              </p>
+              <p style={smallStyles}>
+                {entry.name}
```
cacabo: would prefer destructuring name higher up as well:
```js
const { name } = entry
```
or
```js
const { name: eventName } = entry
```

frontend/components/EventTableRow.js
```diff
@@ -0,0 +1,124 @@
...
+    border-color: rgba(0, 0, 0, 0.1);
+    box-shadow: 0 1px 4px rgba(0, 0, 0, 0.15);
+  }
+
+  ${mediaMaxWidth(MD)} {
+    margin: 0;
+    width: 100%;
+    border-radius: 0;
+  }
+`
```
cacabo: It seems like a lot of this is duplicated from the Row component in shared, why can't we reuse that? If you need to make some modifications/overrides, you can say:
```js
const StyledRow = s(Row)`
  yourStylesHere
`
```

frontend/components/EventTableRow.js
```diff
@@ -0,0 +1,124 @@
...
+
+const Content = s.div`
+  ${mediaMinWidth(LG)} {
+    padding: 0 0.75rem;
+  }
+`
+
+const Subtitle = s.p`
+  color: ${CLUBS_GREY_LIGHT};
+  font-size: .8rem;
```
cacabo:
```suggestion
  font-size: 0.8rem;
```

frontend/components/EventTableRow.js
```diff
@@ -0,0 +1,124 @@
...
+`
+
+const time_format = new Intl.DateTimeFormat("en-US", {
```
cacabo: Would prefer camelCase in JS
```suggestion
const timeFormat = new Intl.DateTimeFormat("en-US", {
```

frontend/components/EventTableRow.js
```diff
@@ -0,0 +1,124 @@
...
+  year: "numeric",
+  month: "long",
+  day: "2-digit",
+  hour: "numeric",
+  minute: "numeric"
+})
+
+class EventTableRow extends React.Component {
```
cacabo: Can this be made a functional component?

frontend/components/EventTableRow.js
```diff
@@ -0,0 +1,124 @@
...
+
+  render() {
+    const {
+      club,
+      event,
+      deleteEvent
+    } = this.props
+
```
cacabo:
```js
const { name, location } = event
```

frontend/components/EventTableRow.js
```diff
@@ -0,0 +1,124 @@
...
+                      day: "2-digit",
+                      hour: "numeric",
+                      minute: "numeric"
+                    }).format(new Date(event.start_time))} - {new Intl.DateTimeFormat("en-US", {
+                      year: "numeric",
+                      month: "long",
+                      day: "2-digit",
+                      hour: "numeric",
+                      minute: "numeric"
+                    }).format(new Date(event.end_time))}
```
cacabo: Would prefer having this logic before the render

frontend/pages/club/[club]/edit.js
```diff
@@ -262,6 +251,24 @@ class ClubForm extends Component {
     })
   }
 
+
+  deleteEvent(id) {
+    console.log(this.state.club.code)
+    console.log(`/clubs/${this.state.club.code}/events/${id}`)
+    doApiRequest(`/clubs/${this.state.club.code}/events/${id}`, {
```
Comment (cacabo) [inline frontend/pages/club/[club]/edit.js]: Would prefer destructuring out the `code` before using it in the string interpolation

frontend/pages/club/[club]/edit.js
```diff
@@ -262,6 +251,24 @@ class ClubForm extends Component {
...
+        resp.json().then(err => {
+          this.notify(formatResponse(err))
+        })
+      }
+    })
+  }
+
   saveMember(username, data) {
     doApiRequest(
       `/clubs/${this.state.club.code}/members/${username}/?format=json`,
```
Comment (cacabo) [inline frontend/pages/club/[club]/edit.js]: Same here

frontend/pages/club/[club]/index.js
```diff
@@ -100,6 +101,17 @@ const Club = ({ club: initialClub, userInfo, favorites, updateFavorites, subscri
             <StrongText>Social</StrongText>
             <SocialIcons club={club} />
           </Card>
+          {club.how_to_get_involved != "" ? (
```
Comment (cacabo) [inline frontend/pages/club/[club]/index.js]:
```suggestion
          {club.how_to_get_involved ? (
```

frontend/pages/club/[club]/index.js
```diff
@@ -100,6 +101,17 @@ const Club = ({ club: initialClub, userInfo, favorites, updateFavorites, subscri
...
+            <Card bordered style={{ marginBottom: '1rem' }}>
+              <div style={{marginLeft: '6px'}}>
```
Comment (cacabo) [inline frontend/pages/club/[club]/index.js]: What does this div/margin do?

frontend/components/ClubPage/Events.js
cphalen: I actually cannot remember why I added this so I've taken it out. Thanks for pointing it out.

frontend/pages/club/[club]/index.js
Comment (cphalen) [reply frontend/pages/club/[club]/index.js]: Removed this one as well

frontend/components/ClubPage/Events.js
cphalen: I think I got this resolved using flexboxes. Thanks for the heads up

PR #82, opened 2020-01-21
Update DLA backend

ArmaanT:
This PR updates the custom django backend to log in users to conform to the [current version provided by DLA](https://github.com/pennlabs/django-labs-accounts/blob/master/accounts/backends.py)

### DIFF
--- a/backend/Pipfile.lock
+++ b/backend/Pipfile.lock
@@ -48,18 +48,18 @@
         },
         "boto3": {
             "hashes": [
-                "sha256:3fa0cd930c4f920b358ed3493102866c96bef2049448ca34c60dcbf9108d8674",
-                "sha256:595c8059320816e7ea6760247076e7462b88212d0bf76d0bb068ec16ae555231"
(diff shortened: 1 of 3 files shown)


ezwang:
@ArmaanT would we also need to update the `Pipfile.lock` to use version 0.5.4?

ezwang:
Also is there a way we could abstract out the parts of DLA where clubs backend differs into methods and then override it from clubs? That way when DLA is updated, we don't have to also update the clubs backend.

ArmaanT:
@ezwang I'll have to take another look, but I'm pretty sure that would be possible under the `post_authenticate` function. I can't believe I didn't think about that. I'll try to make the changes tonight.

ArmaanT:
@ezwang After looking at the code, I don't think it would be possible to subclass the DLA backend and modify the `post_authenticate` method. The main issue is we need custom logic to handle the shuffling of usernames to be able to create a user with the correct pennid and pennkey and that can't be done in `post_authenticate`

PR #85, opened 2020-01-21
Make Login Button More Noticeable

sigmachirality:
The Login Button is styled using CLUBS_RED, so it should change color when the colors are switched over.

Screenshot from 2020-01-21 14-31-29
Screenshot from 2020-01-21 14-31-46


frontend/components/Header/Links.js
```diff
@@ -4,14 +4,39 @@ import s from 'styled-components'
...
   opacity: 0.5;
   margin-right: 4px;
 `
 
+const LoginButton = s.a`
+  border: 0;
+  background-color: ${CLUBS_RED};
+  padding: 14px 20px;
+  margin: auto;
+  color: ${SNOW} !important;
```
cacabo: Would prefer `WHITE_ALPHA(0.8)` or just `WHITE`

frontend/components/Header/Links.js
```diff
@@ -4,14 +4,39 @@ import s from 'styled-components'
...
+
+  &:hover {
+    color: ${SNOW_ALPHA} !important;
```
cacabo: Generally want to keep font color the same but then make the background color a slightly darker version. So feel free to make a color called CLUBS_RED_DARK or CLUBS_RED_DEEP. A good way to do this is to google the hex number then click a little bit to the bottom right of the color in the color picker until there is a noticeable but not extreme difference between the two colors. 

lmk if you need help with this.

frontend/components/Header/Links.js
```diff
@@ -4,14 +4,39 @@ import s from 'styled-components'
...
+  }
+
+  ${mediaMaxWidth(MD)} {
+    padding: 8px 0;
+    padding-top: .4rem;
```
cacabo:
```suggestion
    padding-top: 0.4rem;
```

frontend/components/Header/Links.js
```diff
@@ -4,14 +4,39 @@ import s from 'styled-components'
...
+    width: 5rem !important;
```
cacabo: Why is fixing the width necessary? Can't we just use wider padding on the sides?

frontend/components/Header/Links.js
sigmachirality: Done.

frontend/components/Header/Links.js
sigmachirality: Done, though check out the hex number I chose and make sure I didn't go too far with the difference.

frontend/components/Header/Links.js
```diff
@@ -4,25 +4,57 @@ import s from 'styled-components'
...
   opacity: 0.5;
   margin-right: 4px;
 `
 
+const LoginButton = s.a`
+  border: 0;
+  background-color: ${CLUBS_RED};
+  padding: 14px 20px;
+  margin: auto;
+  color: ${WHITE_ALPHA(0.8)} !important;
```
cacabo:
```suggestion
  color: ${WHITE_ALPHA(0.8)} !important;
  transition: color ${SHORT_ANIMATION_DURATION}ms ease,
              background ${SHORT_ANIMATION_DURATION}ms ease;
```

sigmachirality: lgtm!

PR #86, opened 2020-01-24
Delete old csrftoken cookie

ezwang:
Delete old csrftoken cookie on every page load. Fixes bug with old token shadowing the current working cookie. This code should be removed in a year or so.


ArmaanT:
Can we also delete sessionid? Would also be great to delete all cookies on `.pennclubs.com` but those won't impact anything.

PR #87, opened 2020-01-24
Add Edit Block Screen for Ppl Without Access

ezwang:
image


frontend/components/common/AuthPrompt.js
```diff
@@ -23,32 +23,36 @@ const TitleHeader = s.div`
...
-      </Margin>
-      <SmallText>
-        <i>(We're sorry, we hate two-step too.)</i>
-      </SmallText>
+      {hasLogin && (
+        <>
+          <Margin>
+            <a
+              href={`${LOGIN_URL}?next=${
+                typeof window !== 'undefined' ? window.location.href : '/'
```
sigmachirality:
```suggestion
               window ? window.location.href : '/'
```

might be simpler, not sure if it catches all the cases though.

sigmachirality: also might be able to make href=... one line if you make this change.

frontend/pages/club/[club]/edit.js
```diff
@@ -312,13 +312,28 @@ class ClubForm extends Component {
...
     }
 
+    if (
+      authenticated &&
+      club &&
+      !userInfo.is_superuser &&
+      !userInfo.membership_set.some(m => m.code === club.code && m.role <= 10)
+    ) {
+      return (
+        <AuthPrompt title="Oh no!" hasLogin={false}>
```
Comment (sigmachirality) [inline frontend/pages/club/[club]/edit.js]: Unless AuthPrompt has prop-types enforcing hasLogin be a boolean, you might be able to just delete hasLogin - when AuthPrompt tries to pull hasLogin, it will be undefined, which is falsey.

frontend/pages/club/[club]/edit.js
```diff
@@ -312,13 +312,28 @@ class ClubForm extends Component {
...
+          You do not have permission to edit the page for {club.name}. To get
+          access, contact{' '}
```
Comment (sigmachirality) [inline frontend/pages/club/[club]/edit.js]: Maybe try `&nbsp;` instead of {' '}

frontend/pages/club/[club]/edit.js
```diff
@@ -312,13 +312,28 @@ class ClubForm extends Component {
...
+          <a href="mailto:contact@pennclubs.com">contact@pennclubs.com</a>.
+        </AuthPrompt>
+      )
+    }
+
     if (this.state.isEdit && club === null) {
```
Comment (sigmachirality) [inline frontend/pages/club/[club]/edit.js]: Destructure isEdit on ln 316?

frontend/pages/club/[club]/edit.js
Comment (ezwang) [reply frontend/pages/club/[club]/edit.js]: Ah I set the default value of `hasLogin` to true, which displays the login button.

The login button does not apply in the case of access denied, since the user is logged in but does not have the right permissions.

frontend/pages/club/[club]/edit.js
Comment (ezwang) [reply frontend/pages/club/[club]/edit.js]: This is actually automatically formatted by [Prettier](https://prettier.io/), so I assume that this is the standard.

frontend/pages/club/[club]/edit.js
Comment (ezwang) [reply frontend/pages/club/[club]/edit.js]: Could you clarify this?

frontend/components/common/AuthPrompt.js
ezwang: This would actually result in an error instead of the intended behavior.

image

sigmachirality: sad, kk

frontend/pages/club/[club]/edit.js
Comment (sigmachirality) [reply frontend/pages/club/[club]/edit.js]:
```suggestion
    if (isEdit && club === null) {
```

and destructure isEdit from the statement on line 316

frontend/pages/club/[club]/edit.js
```diff
@@ -312,14 +312,29 @@ class ClubForm extends Component {
   }
 
   render() {
-    const { authenticated, schools, majors, years, tags } = this.props
+    const { authenticated, userInfo, schools, majors, years, tags } = this.props
     const { club, invites, editMember } = this.state
```
Comment (sigmachirality) [inline frontend/pages/club/[club]/edit.js]:
```suggestion
    const { club, invites, isEdit, editMember } = this.state
```
Probably need this as well

PR #88, opened 2020-01-25
Enforce prettier

ezwang:
- Complete and merge this when there are no other PRs, this will likely merge conflict with everything.
- Run `prettier --write ./{components,constants,pages}/**/*.js *.js` in the `frontend` folder before merging.


cacabo:
@ArmaanT how would going (back) to yarn fix this problem / what actually is the problem that we currently have?

ArmaanT:
@cacabo The yarn idea is pretty much unrelated to this PR, but this PR reminded me that it seemed people preferred using yarn to npm. If we think yarn's a better solution then we can switch to using that instead

PR #89, opened 2020-01-26
Update Color Scheme

sigmachirality:
The switch over isn't perfect, but it should be enough to get us started/hold us over for the SAC fair.
index1
index2
club
profile


cacabo:
@sigmachirality can you resolve merge conflicts and yeet?

PR #95, opened 2020-01-26
Member card width/padding bug

cx-lu:
Should fix the name overflow issue by truncating with ellipses - more sustainable solution is a WIP
Screen Shot 2020-01-26 at 5 59 32 PM

### DIFF
--- a/frontend/components/common/Grid.js
+++ b/frontend/components/common/Grid.js
@@ -54,6 +54,7 @@ const ColContainer = s.div`
   background: ${({ background }) => background || 'transparent'};
   overflow-x: visible;
   position: relative;
+  min-width: 0;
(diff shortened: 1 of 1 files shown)


cx-lu:
Also for some reason this doesn't work on mobile - any tips on how to fix?

cacabo:
If we used the `Col` and `Row` components it would probably be easier and more consistent:
```js
<Row margin={MARGIN}>
  {members.map(... => (
    <Col sm={12} md={6} lg={3} margin={MARGIN}>
      ...
    </Col>
  ))}
</Row>
```

cacabo:
Also the initials function we have doesn't work that well for this edge case, do you think you can resolve this?

image

PR #97, opened 2020-01-28
Regenerate Clubs List Cache on Club Modification

sigmachirality:
Club list is regenerated in another thread when a club is edited.

- [x] Make cache indefinite
- [x] Regenerate cache in background thread when a club is modified
- [x] Fix image_url serialization errors
- [x] Debug failing tests


ezwang:
@sigmachirality Instead of returning None, you want to return `image_url`!

PR #100, opened 2020-01-30
Embed svgs inline

ezwang:
- This allows for svgs to be styled by css rules, like `color`.
- [x] Some properties no longer exist, so icon styles are messed up on some pages. Need to fix.
- [x] Need to remove the colored version of some icons and use `color: color` to style instead.


sigmachirality:
Might want to merge this in before https://github.com/pennlabs/penn-clubs/pull/89 and then make sure to restyle all the icons to their new colors again

PR #105, opened 2020-02-01
Add frontend test cases

ezwang:
- Use [Cypress](https://www.cypress.io/) for testing


ArmaanT:
The actual testing makes sense to me, but maybe we should try to generalize some of the CI code and put it in a CircleCI orb. Since these are closer to integration tests, I wonder how feasible it would be to run these after building and publishing the docker images so that we don't have to worry about setting up each individual app

ezwang:
@ArmaanT that sounds like a great idea, we can definitely abstract out a lot of this code into orbs.

However, these tests should run before we publish the docker images, since if there are any errors we don't want to publish a docker image. I'm also thinking of making a `./manage.py populate` command to populate the backend with fake data, for both development and for testing.

ArmaanT:
I looked into it a bit more, and we can use a docker-compose file in CI to spin up multiple containers. We could very easily have those run before publishing the images. A good side effect is that we could use that same system to develop locally if we wanted to. We can merge now and figure that out later

PR #112, opened 2020-02-07
Better Tag Filter UI

sigmachirality:
- [x] Rename to Tags
- [x] use react-select to expose Multiselect to user for Tags

This PR reworks the Tags filtering UI by adding a multiselect that allows users to search for tags. Additionally, this PR shifts the color scheme a bit by changing Tags to blue, Size to navy and Application Needed to red, allowing the Tags section coloring to match the tags on the club cards.

Also could use some direction on whether I should factor out FilterHeader from the DropdownFilter component or move it to another component. @cacabo @ezwang thoughts? I left a TODO in the relevant spot in the codebase that I'm talking about,

Screenshot from 2020-02-09 04-05-14
Screenshot from 2020-02-09 19-45-46


frontend/components/FilterSearch.js
```diff
@@ -0,0 +1,153 @@
...
+} from '../constants/colors'
+
+const SearchWrapper = s.div`
+  margin-bottom: 30px;
+  max-height: 0;
+  opacity: 0;
+  overflow: visible;
+  transition: all ${ANIMATION_DURATION}ms ease;
+
+  ${({ active }) => active && 'max-height: 150vh; opacity: 1;'}
```
cacabo: Why 150vh?

frontend/components/FilterSearch.js
```diff
@@ -0,0 +1,153 @@
...
+const SearchIcon = s.span`
+  cursor: pointer;
+  color: ${MEDIUM_GRAY};
+  opacity: 0.5;
+  margin-right: 6px;
+
+  ${mediaMaxWidth(MD)} {
+    right: 24px;
+  }
+`
```
cacabo: Would be good to put a comment above this just explaining why we style things this way (I know I don't really do that but we should get in the habit of it lool)

frontend/components/FilterSearch.js
```diff
@@ -0,0 +1,153 @@
...
+
+const Search = ({ searchTags, recommendedTags, updateTag }) => {
+  // Custom styles for the react-select
+  const styles = {
+    control: ({ background, ...base }, { isFocused, isSelected }) => {
+      const isEmphasized = isFocused || isSelected
+      return {
+        ...base,
+        border: `1px solid ${ALLBIRDS_GRAY}`,
```
cacabo:
```suggestion
        border: `1px solid ${BORDER}`,
```

frontend/components/FilterSearch.js
```diff
@@ -0,0 +1,153 @@
...
+        background: isEmphasized ? FOCUS_GRAY : background,
+        color: CLUBS_GREY,
+      }
+    },
+  }
+
+  // Overriding specific components of the react-select
+  const components = {
+    IndicatorSeparator: () => null,
+    DropdownIndicator: () => <SearchIcon><Icon name="tag" alt="" /></SearchIcon>,
```
cacabo:
```suggestion
    DropdownIndicator: () => <SearchIcon><Icon name="tag" alt="" />,
```

frontend/components/FilterSearch.js
```diff
@@ -0,0 +1,153 @@
...
+    loadOptions={searchTags}
+    defaultOptions={recommendedTags}
+    cacheOptions
+    onChange={(_, selectEvent) => {
+      const { action, option } = selectEvent
+      action === 'select-option' && updateTag(option, 'Tags')
+    }}
+    placeholder="Search for tags"
+    value={null}
+  />
```
cacabo: Should be wrapped in parens make sure your prettier is working

frontend/components/FilterSearch.js
```diff
@@ -0,0 +1,153 @@
...
+  return <Select
+    isMulti
+    styles={styles}
+    components={components}
+    loadOptions={searchTags}
+    defaultOptions={recommendedTags}
+    cacheOptions
+    onChange={(_, selectEvent) => {
+      const { action, option } = selectEvent
+      action === 'select-option' && updateTag(option, 'Tags')
```
cacabo: What does this do?

frontend/components/FilterSearch.js
```diff
@@ -0,0 +1,153 @@
...
+    }}
+    placeholder="Search for tags"
+    value={null}
+  />
+}
+
+const selectRecommended = (tags = []) => {
+  const options = tags
+    .sort(() => 0.5 - Math.random())
+    .slice(0, Math.min(tags.length, 3))
```
cacabo: have a short comment explaining what this does

frontend/components/FilterSearch.js
```diff
@@ -0,0 +1,153 @@
...
+      name="Tags"
+      toggleActive={toggleActive}
+    />
+    <SearchWrapper active={active}>
+      <Search
+        searchTags={searchTags}
+        recommendedTags={recommendedTags}
+        updateTag={updateTag}
+      />
+    </SearchWrapper>
```
cacabo: Indent is off by 2 spaces

frontend/components/FilterSearch.js
sigmachirality: this onChange is provided by react-select, it handles when an option is selected. I check that the event is "select" here and then add the tag to the selected tags using updateTag

frontend/components/FilterSearch.js
sigmachirality: fixed prettier, should be fixed now

frontend/components/FilterSearch.js
sigmachirality: added

frontend/components/FilterSearch.js
sigmachirality: fixed prettier, should be good now.

frontend/components/FilterSearch.js
sigmachirality: I don't think this works

frontend/components/FilterSearch.js
sigmachirality: changed to 100%, needed some value to override max-height: 0 set above when active is true

frontend/components/FilterSearch.js
sigmachirality: actually I think since moving to feather from fa that we shouldn't have to style things this way anymore, lemme try some things out

sigmachirality: simplified styles, lmk if you still think it needs a comment

sigmachirality:
This is good to go imo

cacabo:
Not sure why the build failed @ezwang seems like we just need to kick it back off again?

sigmachirality:
> Not sure why the build failed @ezwang seems like we just need to kick it back off again?

I'm trying to fix the incompatibilities between the current eslint config and prettier, so when I commit that we can see what happens.

ezwang:
@sigmachirality we're currently using both `fuse.js` and `fuzzysort`. These libraries do the same thing, we should choose one and stick to it!

ezwang:
@sigmachirality also should we just render all of the tags anyways in the dropdown? It may be confusing to the user as to which tags exist, and having a list of all of them can help the user decide.

Unlike in PCR, there should be a small enough number of tags that there will not be any performance decrease.

ezwang:
@sigmachirality another possible suggestion would be to display the selected tags in the select input in addition to the tags selected list on the right. This can make the user experience less confusing for users.

sigmachirality:
@ezwang I'm going to switch the code to fuse from fuzzysort, but for the design questions I think you should work that out with @valenciafu @cacabo and the other designers after this PR is closed. I opened a story on clubhouse and assigned @cacabo to it.

sigmachirality:
Added selected tags and club counts to the multiselect, it's a bit weird that selectedtags are duplicated on mobile but that's a problem that can be handled in the mobile redesign of the filter.

PR #118, opened 2020-02-11
Relative requests to the backend

ArmaanT:
The latest version of DLA (0.6.0) requires all requests to use a relative `next` parameter. This PR updates clubs to do just that. I tested this locally and all the links seem to work, but someone else should double check this.


sigmachirality:
weow Armann does frontend???

frontend/pages/invite/[club]/[invite]/[token]/index.js
```diff
@@ -39,7 +39,7 @@ class Invite extends Component {
             resp.status === 403 &&
             data.detail === 'Authentication credentials were not provided.'
           ) {
-            window.location.href = `${LOGIN_URL}?next=${window.location.href}`
```
Comment (ezwang) [inline frontend/pages/invite/[club]/[invite]/[token]/index.js]: @ArmaanT note that this will drop any query parameters and hashes. For example, "/settings#Bookmarks" will simply get redirected to "/settings". For a better user experience, we should append `window.location.search` and `window.location.hash`. These two strings will be empty by default if there are no parameters/hash.

Comment (ArmaanT) [reply frontend/pages/invite/[club]/[invite]/[token]/index.js]: And this is where I'm totally out of my depth. I can take a stab at it if you want, but I'm sure you'd do it better/faster than I could

PR #119, opened 2020-02-17
Added back selected fields sidebar

davish:
Added back selected fields sidebar for the demo tomorrow

image

### DIFF
--- a/frontend/pages/reports.js
+++ b/frontend/pages/reports.js
@@ -8,6 +8,7 @@ import {
   Checkbox,
   CheckboxLabel,
 } from '../components/common'
+import { mediaMinWidth, mediaMaxWidth, MD, LG } from '../constants/measurements'
 import { doApiRequest, API_BASE_URL } from '../utils'
 import { Container } from '../components/common/Container'
 import { CLUBS_GREY } from '../constants/colors'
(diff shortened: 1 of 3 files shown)


frontend/pages/reports.js
```diff
@@ -99,130 +122,149 @@ const Reports = ({ nameToCode }) => {
...
+        <div className="box">
+          <h3 className="title is-4" style={{ color: CLUBS_GREY }}>
+            Select fields to include
+          </h3>
+          <div>
+            <Flex>
+              {Object.keys(fields).map(group =>
+                generateCheckboxGroup(group, fields[group])
+              )}
+            </Flex>
```
pawalt: 💪💪💪

frontend/pages/reports.js
```diff
@@ -28,6 +28,29 @@ const serializeParams = params => {
...
 }
 
+const HoverListElement = s.li`
+  span {
+    display: none;
+  }
+  &:hover span {
+    display: inline;
+    cursor: pointer;
+  }
```
cacabo: This doesn't really work with mobile—do we care about that use case?

Also could we maybe try like `margin-left: auto` so all the icons will be aligned on the right side of the sidebar? Might be better UI

ezwang:
@davish does this work on smaller screens? I left it out because of that, but we can add it back temporarily and fix the mobile issue after the demo!

davish:
Yeah, I don't think this will work on mobile. But I also don't see OSA generating reports on mobile, I think it's fine to not support mobile for this view, since you need to download the excel file.

ezwang:
@davish maybe a rule to just hide the sidebar if we're on mobile? That way the page is at least still functional. The ideal fix would just be to prevent the overlap by shifting the content of the page downwards.

PR #120, opened 2020-02-17
Add Colored Header and Moved Edit to Separate Component

mtsiang19:
EDIT - @ezwang 

image

-----

image


sigmachirality:
Can you add some screenshots?

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+
+const TransparentButton = s.button`
+  width: 190px;
+  height: 35px;
+  border-radius: 17px;
+  border: 0;
+  background: rgba(255,255,255,0.32);
+  float: right;
+  font-size: 15px;
+  font-family: Lato;
```
sigmachirality: Is the font-family being switched to Lato in general? Feel strange to have one font in just one place

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+  background: linear-gradient(to right, #ef4c5f, #4954f4);
+  height: 7em;
+`
+
+const TransparentButton = s.button`
+  width: 190px;
+  height: 35px;
+  border-radius: 17px;
+  border: 0;
+  background: rgba(255,255,255,0.32);
```
sigmachirality: Try to see if there's a constant in colors that you can use here, or make one?

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+import renderPage from '../../renderPage.js'
+import { Icon, Empty, Checkbox, CheckboxLabel } from '../../components/common'
+import { doApiRequest, API_BASE_URL } from '../../utils'
+import { Container } from '../../components/common/Container'
+import { CLUBS_GREY } from '../../constants/colors'
+import Edit from './edit'
+
+const GroupLabel = s.h4`
+  margin-bottom: 0em !important;
+  font-size: 32px;
```
sigmachirality: would prefer measurements in ems

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+  color: #626572;
+`
+
+const ColoredHeader = s.div`
+  background: linear-gradient(to right, #ef4c5f, #4954f4);
```
sigmachirality: same, use constant colors if they exist

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+  height: 7em;
+`
+
+const TransparentButton = s.button`
+  width: 190px;
+  height: 35px;
+  border-radius: 17px;
+  border: 0;
+  background: rgba(255,255,255,0.32);
+  float: right;
```
sigmachirality: float is jank, try to achieve the same positioning using margins maybe?

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+
+  const [reports, setReports] = useState([])
+  const [reportFlag, updateReportFlag] = useState(false)
+
+  const [isEdit, setIsEdit] = useState(false)
+
+  useEffect(() => {
+    doApiRequest('/reports/?format=json')
+      .then(resp => {
+        if (resp.ok) {
```
sigmachirality: prefer ternary statement

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+          return resp.json()
+        } else {
+          return []
+        }
+      })
+      .then(data => setReports(data))
+  }, [reportFlag])
+
+  const [includedFields, setIncludedFields] = useState(
+    (() => {
```
sigmachirality: you don't have to call the anonymous function manually here, useState accepts functions!

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+  return (
+    <div>
+      {isEdit ? (
+        <div>
+          <ColoredHeader>
+            <div style={{ padding: '2em' }}>
+              <span className="title is-3" style={{ color: 'white' }}>
+                Create a new report
+              </span>
+              <TransparentButton onClick={() => handleBack()}>
```
sigmachirality: could just be onClick={handleBack} I think, unless there's some reason you want to call handleBack with no args here?

sigmachirality:
I'm of the opinion that JSX should contain as little inline JS logic as possible (I think @cacabo is also in this camp). Too much inline JS logic can make it hard to read the code (pcr is suffering a bit from this rn lol)

frontend/pages/reports/index.js
cacabo: ^can also add the constant colors so other ppl can use them and they're more human readable

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+
+const GroupLabel = s.h4`
+  margin-bottom: 0em !important;
+  font-size: 32px;
+  color: #626572;
+`
+
+const ColoredHeader = s.div`
+  background: linear-gradient(to right, #ef4c5f, #4954f4);
+  height: 7em;
```
cacabo: Why 7em?

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+`
+
+const TransparentButton = s.button`
+  width: 190px;
+  height: 35px;
```
cacabo: What's the significance of these? For any seemingly arbitrary measurements first (1) make sure they work on a few screen sizes/resolutions/browsers and (2) add a comment above explaining why you chose those values. Else it's hard for other ppl to pick up where you left off. Generally fixing button width instead of adding padding is not good–what if the font size changes or the font changes?

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+  border-radius: 17px;
```
cacabo: Why 17px? We tend to use `BORDER_RADIUS` from `constants`

frontend/pages/reports/index.js
cacabo: this could be called `WHITE_ALPHA(0.32)`

frontend/pages/reports/index.js
cacabo: Yep. Either use margins or flexbox.

frontend/pages/reports/index.js
cacabo: Do not change font family. We are using`inter` per recent PR by @valenciafu .

frontend/pages/reports/index.js
```diff
@@ -0,0 +1,227 @@
...
+  float: right;
+  font-size: 15px;
+  font-family: Lato;
+  font-weight: 500;
+  font-stretch: normal;
+  font-style: normal;
+  line-height: normal;
+  letter-spacing: normal;
+  text-align: center;
+  color: rgba(255,255,255,1)
```
cacabo: this is just `WHITE`

frontend/pages/reports/index.js
cacabo: I've found in eslint things that this syntax is actually preferred. When/if we move to TypeScript the linter will prefer we use this style (deliberately not sending the event if we don't need it)

frontend/pages/reports/edit.js
```diff
@@ -0,0 +1,98 @@
...
+            <div className="field">
+              <label className="label">Description</label>
+              <div className="control">
+                <TallTextArea
+                  name="description"
+                  className="input"
+                  type="text"
+                  placeholder='e.g. "Pulls all clubs, the emails from club owners, and names of owners"'
+                  value={descInput}
+                  onChange={e => setDescInput(e.target.value)}
```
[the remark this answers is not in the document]
sigmachirality: I defer to resident JS god

frontend/pages/reports/index.js
sigmachirality: makes sense, () => handleBack() swallows the event so it makes the types nicer, and if handleBack checks to see if any meaningful params are passed in this lets handleBack know there isn't.

frontend/pages/reports/index.js
mtsiang19: Oops my bad, thought it was Lato

PR #123, opened 2020-02-19
Club Create Page

davish:
The `edit.js` page already had all the logic for club creation, so this PR factors out that file into a new `ClubEditPage` component. The `club/[club]/edit` and `/create` pages now use this component to render the right content.

This PR also has backend logic to prevent those without permission from marking a club as accepted. Clubs created through the API are always marked as not accepted / pending acceptance.

Filtering logic is also added to make sure that only those with permission can view non-approved clubs. Members of clubs can always view their clubs even if those clubs aren't approved.

TODO:

- [x] Indicate that a club as "pending approval" on the club page
- [x] Add button to approve club via a `PATCH` request on club page
- [ ] Tests for backend logic
- [x] Maybe attempt frontend tests?


ezwang:
Remember to approve all existing clubs before merging this PR, or otherwise no clubs will appear on the home page.

backend/clubs/views.py
```diff
@@ -233,6 +243,20 @@ def get_filename(self):
             return "{}.xlsx".format(slugify(name))
         return super().get_filename()
 
+    def partial_update(self, request, *args, **kwargs):
+        if request.data.get("accepted", False) is True and not request.user.has_perm(
+            "approve_club"
+        ):
+            return HttpResponseForbidden()
```
ezwang: We want to use [`PermissionDenied`](https://stackoverflow.com/a/50076664) instead of `HttpResponseForbidden` here. `HttpResponseForbidden` is used in Django, while `raise PermissionDenied` is used in DRF.

ezwang:
Note that the current design leaks information with the caching feature! Be sure to fix before merging.

cphalen:
Looks good to me! Happy to see the caching removed at least for now.

sigmachirality:
rip caching

PR #126, opened 2020-02-23
Membership Request

xiayufeilucy:


Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
ezwang: Remove this file!

backend/clubs/migrations/0018_auto_20200207_1737.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 3.0.2 on 2020-02-07 22:37
```
ezwang: Combine these migrations before merging!

PR #128, opened 2020-03-08
Event preview

ezwang:
- Implement event phone preview feature for Spring Fling
- Merge after #127 

image


sigmachirality:
rip, spring fling is canceled?

PR #129, opened 2020-03-12
Backspace Button Can Delete Tags

sigmachirality:
title


sigmachirality:
Changed to using the react-select functionality instead of writing my own key handler, seems to have fixed the issue. Take a look when you have time!

PR #133, opened 2020-04-08
Make Add Club Button Blue

sigmachirality:
The green on the home page doesn't feel right to me, this makes it the same color as the login button so we maintain visual consistency while also still making the button stand out to the user. Thoughts?

I think we should also change the color of the Edit Club button, or maybe move it under the ellipsis dropdown?

EDIT - @ezwang 

image

### DIFF
--- a/frontend/components/DisplayButtons.js
+++ b/frontend/components/DisplayButtons.js
@@ -2,7 +2,14 @@ import Link from 'next/link'
 import PropTypes from 'prop-types'
 import s from 'styled-components'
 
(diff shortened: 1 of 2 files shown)


valenciafu:
what do you think about making it clubs blue? create club is definitely not the same level of importance as log in. edit button can be blue too. 

in general, im gonna think more about color-function conventions

sigmachirality:
@valenciafu I'm down to do whatever, just don't like the current green!

ezwang:
@sigmachirality @valenciafu Did we ever come to a consensus for what the color of this button should be?

valenciafu:
@ezwang let's go w clubs blue

PR #134, opened 2020-04-18
Add Info About Symlinks to README

sigmachirality:
This repo contains several symlinks, which aren't reproduced on Windows by Git by default due to security reasons. Adds info to the Development section of the README on how to enable symlinks in development for Windows users.


ezwang:
Modified pull request to specify that this additional step is only necessary on Windows.

PR #136, opened 2020-05-09
Render authenticated data on first page load

davish:
This PR adds cookie forwarding to requests in `getInitialProps` to remove the need to send AJAX requests once the page loads on the client. User data and individual club data will be loaded in on the NextJS server, and sent in the initial HTML payload to the client.

This gets rid of any flicker on page load.


davish:
Hmm so fixed the build failure... but still one test passing. I tried running tests locally and they all passed. Going to try re-running the tests to see if it was a flaky thing

PR #142, opened 2020-07-25
Near finalized commit for backend searching

cphalen:
Mainly re-purposed the `resetDisplay` method in `index.tsx`. Lots of kinda convoluted logic for parsing the checkboxes and turning it into the correct backend query -- let me know if you have ideas on how to clean this up.

Also, made a few small changes to `views.py`. Namely, fixing a bug on parsing the operation, as well as adding supporting for parsing multiple ints (although I think you also added in a solution for this concurrently, when merging I just kept both for good measure because I didn't want to break anything else you might've been working on).

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -205,6 +205,11 @@ class ClubsFilter(filters.BaseFilterBackend):
             return {}
 
         def parse_int(field, value, operation, queryset):
+            if operation == "in":
+                values = value.strip().split(",")
+                sizes = [int(size) for size in values if size]
+                return {f"{field}__in": sizes}
+
             if "," in value:
                 values = [int(x.strip()) for x in value.split(",") if x]
                 if operation == "and":
@@ -212,6 +217,7 @@ class ClubsFilter(filters.BaseFilterBackend):
                         queryset = queryset.filter(**{field: value})
                     return queryset
                 return {f"{field}__in": values}
+                
             if value.isdigit():
                 suffix = ""
                 if operation in {"lt", "gt", "lte", "gte"}:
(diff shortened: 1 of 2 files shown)


backend/clubs/views.py
```diff
@@ -205,13 +205,19 @@ def parse_year(field, value, operation, queryset):
             return {}
 
         def parse_int(field, value, operation, queryset):
+            if operation == "in":
+                values = value.strip().split(",")
+                sizes = [int(size) for size in values if size]
+                return {f"{field}__in": sizes}
+
```
cphalen, line 212: My solution to the same problem that you solve below, instead I added a "in" operation. We can go with either solution if you wanna take mine out no worries.

frontend/pages/index.tsx
```diff
@@ -159,43 +159,61 @@ class Splash extends React.Component<SplashProps, SplashState> {
...
+        (tag) => tag.name === 'Application' && tag.value === 1,
+      ) !== -1
+    const noRequiredApplication =
+      selectedTags.findIndex(
+        (tag) => tag.name === 'Application' && tag.value === 2,
+      ) !== -1
+    const acceptingMembers =
+      selectedTags.findIndex(
+        (tag) => tag.name === 'Application' && tag.value === 3,
+      ) !== -1
```
cphalen, line 180: Kinda messy but I couldn't come up with a better way to do this

backend/clubs/views.py
ezwang: The default operation is actually "in" if you don't specify "and", see line 214/219 below.

I like how you made it explicit though! Definitely makes it less confusing for someone reading the code. One option could be to add a comment above line 214/219 saying that the default behavior is to do "in".

frontend/pages/index.tsx
ezwang: Hmm, I can't think of a better way to do this either without refactoring more broadly. The whole sidebar tag system could be refactored a bit to make it easier to extend, but that can be done in a later commit.

This works for me!

PR #144, opened 2020-08-10
[WIP] Rescue Club Tags

sigmachirality:
So I figured out why the tags disappeared, turns out there was a switch statement that expects the kind property of a tag object to be populated with either "tag" or "badge", so I added a default branch and now tags show up.

However, from looking at the TS types, I suspect that somewhere in the frontend or the backend the tag objects should be annotated with the kind property containing either the string value of "tag" or "badge". Should I annotate the objects, and should it be on the backend (in a Serializer or smth) or on the frontend?

### DIFF
--- a/frontend/components/ClubPage/Header.tsx
+++ b/frontend/components/ClubPage/Header.tsx
@@ -18,7 +18,6 @@ type HeaderProps = {
 
 const Header = ({ club, style }: HeaderProps): ReactElement => {
   const { active, name, tags, badges } = club
-
   return (
     <div style={style}>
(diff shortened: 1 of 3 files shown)


ezwang:
@sigmachirality You should be able to annotate them in the frontend, since they're two separate requests. This also saves us from sending over a useless field for each tag/badge request. I believe you can differentiate them using typescript using a type attribute. Depending on the type of the object, the `kind` field will have a different value.

https://github.com/pennlabs/penn-clubs/blob/master/frontend/types.ts#L54

PR #145, opened 2020-08-10
Add Button For Alphabetical Sort On Splashpage

sigmachirality:
reorderButton

Added "reorder" button on the splashpage. I don't like how cluttered the filtering buttons are with the addition of this button, and I also think it could have better wording, but I defer to the design team on those matters - the button is functional.

### DIFF
--- a/frontend/components/DisplayButtons.tsx
+++ b/frontend/components/DisplayButtons.tsx
@@ -47,6 +47,7 @@ const buttonStyles = {
 
 const DisplayButtons = ({
   switchDisplay,
+  switchSort,
   shuffle,
 }: DisplayButtonsProps): ReactElement => (
   <DisplayButtonsTag>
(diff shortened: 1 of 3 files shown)


sigmachirality:
I also don't like how I'm sorting displayClubs each time displayAlphabatized is toggled to true - ideally alphabatizedDisplayClubs would be generated each time displayClubs is updated in state, but this is not easy to achieve in stateful components. Ideally if/when this component is converted to a functional component, a hook is used to maintain state in this way.

frontend/components/DisplayButtons.tsx
```diff
@@ -72,6 +73,14 @@ const DisplayButtons = ({
       <Icon name="shuffle" alt="shuffle club order" style={iconStyles} />
       &nbsp;&nbsp; Shuffle
     </button>
+    <button
+      onClick={switchSort}
+      style={{ ...buttonStyles, color: DARK_GRAY, fontWeight: 600 }}
+      className="button is-small"
+    >
+      <Icon name="filter" alt="shuffle club order" style={iconStyles} />
```
ezwang, line 81: This could potentially be a drop down with different options, like:
- Recommended
- Bookmarks
- Name
- etc

sigmachirality:
Implemented already so I'm closing this

PR #146, opened 2020-08-10
Add renewal request card for all clubs 

snie11:
image

Added renewal card at the top of all club pages to gear up for renewal process. Left out some of the logic for @ezwang to look at re: renewal schema. 

Side note: am i a frontend dev now?????????????


ezwang:
@snie11 I added the logic to hide it if the club is already marked as active. I also added a link to the renew page if the user is an officer of that club.

image

The button could use some work style wise! I'll merge this in for now, feel free to upgrade the button in another pull request.

PR #147, opened 2020-08-10
Move Delete Profile Picture Button

sigmachirality:
Screenshot from 2020-08-10 14-14-55

One thing that's confusing with this change is that the upload image button requires you to press save, whereas the delete image button deletes the image straight up, without needing the user to press save. Maybe I should edit the Form component so that you can have a file upload field that shows two buttons when the field is uploaded, one of which sets [name]: null like the delete profile pic button was doing before, instead of "injecting' the button like I did here? lmk what y'all think.

### DIFF
--- a/frontend/components/Settings/ProfileForm.js
+++ b/frontend/components/Settings/ProfileForm.js
@@ -82,18 +82,22 @@ class ProfileForm extends React.Component {
         name: 'image',
         type: 'file',
         label: 'Profile Picture',
-        trivia: settings.image_url ? <button
-        onClick={this.deleteProfilePic}
-        className="button is-danger is-pulled-right"
-        style={{
-          marginLeft: M2
-        }}
-      >
-        <span className="file-icon"><Icon name="trash" /></span>
-        <span className="file-label">
-          Remove Image
-        </span>
-      </button> : <></>
+        trivia: settings.image_url ? (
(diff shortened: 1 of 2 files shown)


frontend/components/Settings/ProfileForm.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 
+import { M1, M2 } from '../../constants'
```
ezwang, line 3: What's this import for? Is it used anywhere in the code?

ezwang:
> One thing that's confusing with this change is that the upload image button requires you to press save, whereas the delete image button deletes the image straight up, without needing the user to press save. Maybe I should edit the Form component so that you can have a file upload field that shows two buttons when the field is uploaded, one of which sets [name]: null like the delete profile pic button was doing before, instead of "injecting' the button like I did here? lmk what y'all think.

I agree that this behavior is confusing, it would be great to have image fields auto upload since they require a separate request anyways.

frontend/components/Settings/ProfileForm.js
ezwang: Ah, I see that `M2` is used but `M1` is not. Remove `M1` before merging!

ezwang:
Wheee!

PR #148, opened 2020-08-10
Renewal backend

cphalen:
Added:

- `/renew` detail endpoint for clubs that sets `active=true
- `deactivate` command that sets `active=false` for all clubs (be careful with this one!)
- Renew form now posts to renewal endpoint to clubs are set to active after finishing the form

### DIFF
--- a/backend/clubs/management/commands/deactivate.py
+++ b/backend/clubs/management/commands/deactivate.py
@@ -32,6 +32,6 @@ class Command(BaseCommand):
                 self.stdout.write("Input:", ending=" ")
 
         # deactivate all clubs
-        Club.objects.all().update(active=False)
+        Club.objects.all().update(active=False, approved=False, approved_by=None)
 
         self.stdout.write("All clubs deactivated!")
--- a/backend/clubs/management/commands/deactivate.py
+++ b/backend/clubs/management/commands/deactivate.py
@@ -5,25 +5,33 @@ from clubs.models import Club
 
 class Command(BaseCommand):
     help = "Deactivates all clubs in the database. This should be used at \
-            the end of the school year when clubs must be renewed."
+            the beginning of the school year when clubs must be renewed."
 
-    def handle(self, *args, **kwargs):
-        print(
-            "\033[93m"
-            + "You are about to set all club status to inactive and will have"
-            + "to begin the renewal process. This should only happen at the end"
-            + " of the school year. Are you postive this is what you want"
-            + " to do? Type 'deactivate all clubs' to continue."
-            + "\033[0m"
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--force",
+            dest="force",
+            action="store_true",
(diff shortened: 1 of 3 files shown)


cphalen:
A few questions:

- Do we want the `deactivate` script also to set `approved=false` for all clubs?
- Do we want clubs to not be visible on the site when `active=false`? This is something we do when `approved=false` so if the script sets `active=false` and `approved=false` maybe this is not necessary.

backend/clubs/management/commands/deactivate.py
```diff
@@ -0,0 +1,29 @@
...
+
+from clubs.models import Club
+
+
+class Command(BaseCommand):
+    help = "Deactivates all clubs in the database. This should be used at \
+            the end of the school year when clubs must be renewed."
+
+    def handle(self, *args, **kwargs):
+        print(
```
ezwang, line 11: You want to use `self.stdout` instead of `print` when writing management commands. From the [Django documentation](https://docs.djangoproject.com/en/3.0/howto/custom-management-commands/):

image

frontend/pages/club/[club]/renew.tsx
```diff
@@ -245,6 +245,14 @@ const RenewPage = ({
 
   const nextStep = () => {
     setStep(step + 1)
+    if (step === 3) {
+      doApiRequest(`/clubs/${club.code}/?format=json`, {
```
Comment (ezwang) [inline frontend/pages/club/[club]/renew.tsx:249]: @cphalen I modified this to use an existing endpoint instead of creating a new endpoint!

backend/clubs/management/commands/deactivate.py
ezwang: @cphalen I fixed this for you, see the updated file for my changes.

I also used `.update(active=True)` instead of the for loop. The difference between the two is that the for loop runs the following SQL queries:
```
UPDATE TABLE clubs SET active = True WHERE code = 'penn-labs';
UPDATE TABLE clubs SET active = True WHERE code = 'club-one';
UPDATE TABLE clubs SET active = True WHERE code = 'club-two';
...
```
The `.update` solution runs the following query:
```
UPDATE TABLE clubs SET active = True;
```
Since we're making one query and doing it on the SQL side, the update is much faster. It doesn't really matter for this use case, since we're only running it once per year, but I thought you might want to know!

ezwang:
> A few questions:
> 
> * Do we want the `deactivate` script also to set `approved=false` for all clubs?

Yup, we want to do this! We also want the script to send out emails to all club members telling them that they have to go through the renewal process, but we can do this in a later pull request.

> * Do we want clubs to not be visible on the site when `active=false`? This is something we do when `approved=false` so if the script sets `active=false` and `approved=false` maybe this is not necessary.

Once we mark the club as `approved=False`, it won't show up on the site anyways and we won't have to worry about it. In the new system, if `active=false`, then `approved=false` (the club officers have to approve before OSA approves).

PR #155, opened 2020-08-16
Email club officers when club is approved or rejected

ezwang:
- Add a field where the admin can add a reason for approving/rejecting the club.
- Add an email when the club is approved or rejected.
- Added a test case.
- Added a email preview feature to change the value of a variable.
- Fix a bug where officers couldn't renew clubs. Oof.

image
image


ezwang:
image

PR #156, opened 2020-08-16
Membership toggle enhancement

01jongmin:
- Changing permission dropdown updates title input to be the same as the dropdown given that the title text is one of (Member, Officer, Owner) [VIDEO](https://youtu.be/NAD48n0GauA)
- Added a MembershipRole interface that was shared among two files for typescript integration

### DIFF
--- a/frontend/components/ClubEditPage/InviteCard.tsx
+++ b/frontend/components/ClubEditPage/InviteCard.tsx
@@ -85,7 +85,8 @@ export default function InviteCard({ club }: InviteCardProps): ReactElement {
 
   const updatePermissions = (opt: MembershipRole) => {
     setInviteRole(opt)
-    if (['Member', 'Officer', 'Owner'].includes(inviteTitle)) {
+
+    if (MEMBERSHIP_ROLES.map((role) => role.label).includes(inviteTitle)) {
       setInviteTitle(opt.label)
     }
   }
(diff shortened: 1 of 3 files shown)


frontend/components/ClubEditPage/InviteCard.tsx
```diff
@@ -84,6 +83,13 @@ export default function InviteCard({ club }: InviteCardProps): ReactElement {
       })
   }
 
+  const updatePermissions = (opt: MembershipRole) => {
+    setInviteRole(opt)
+    if (['Member', 'Officer', 'Owner'].includes(inviteTitle)) {
```
ezwang, line 88: @01jongmin This could be changed to `MEMBERSHIP_ROLES.map(role => role.label).includes(...)`.

What you have now is fine, but it is a good idea to use existing constants when possible so that if we ever add more roles in the future, nothing breaks.

ezwang: Feel free to merge this after this is fixed!

01jongmin: MB, I'll fix it and commit it again.
Regarding the failing codecov/patch, is it okay to ignore the error and merge?

ezwang: Thanks! Yup, you can safely ignore that.

Ideally it would be nice to have code coverage for everything, but unit tests are a pain. 😅

PR #158, opened 2020-08-18
Metadata fix branch

cphalen:
Two pages fixed

- `/renew` page (which is actually the same as `/edit` page)
- Club invite page


frontend/pages/invite/[club]/[invite]/[token]/index.js
```diff
@@ -82,6 +83,7 @@ class Invite extends Component {
             className="has-text-centered"
             style={{ margin: 30, marginTop: 60 }}
           >
+            <ClubMetadata club={club} />
```
Comment (ezwang) [inline frontend/pages/invite/[club]/[invite]/[token]/index.js:86]: This won't work as you expect, as this case only happens if the club does not exist. It would be better to use `<Metadata />` in this case.

PR #168, opened 2020-08-21
Logo w/ text on welcome page

snie11:
image


frontend/components/ClubEditPage.tsx
```diff
@@ -254,27 +255,6 @@ class ClubForm extends Component<ClubFormProps, ClubFormState> {
...
-          ),
-          disabled: !isEdit,
-        },
-        {
-          name: 'events',
-          label: 'Events',
-          content: (
-            <>
-              <EventsCard club={club} />
-            </>
```
ezwang, line 268: @snie11 This is undoing some of the changes that Davis made, might be a merge conflict issue.

frontend/constants/measurements.ts
```diff
@@ -15,7 +15,7 @@ export const DESKTOP = '1248px'
 export const TABLET = '992px'
 export const PHONE = '584px'
 
-export const NAV_HEIGHT = '3.25rem'
+export const NAV_HEIGHT = '4em'
```
ezwang, line 18: @snie11 Do we want to reset this since we're not changing the navbar?

frontend/components/ClubEditPage.tsx
ezwang: You can fix this easily by reverting the whole file to master.

PR #194, opened 2020-08-26
SAC Fair Guide

snie11:
Created guide (with screenshots) on /sacfairguide


frontend/pages/sacfairguide.tsx
```diff
@@ -0,0 +1,285 @@
...
+        control the session.
+      </li>
+      <li>
+        Have the <strong>chat window open</strong> to answer any questions. For
+        busier sessions, we recommend designating one member to be monitoring
+        the chat window
+      </li>
+      <li>
+        For busier sessions where multiple club members are present, we
+        recommend taking advantage of the <strong>breakout room feature</strong>
```
ezwang, line 149: We should warn users that they need to set up breakout rooms *beforehand* and enable the settings on their account, so they don't try to use it right before the meeting starts and have to redo their Zoom links.

frontend/pages/sacfairguide.tsx
```diff
@@ -0,0 +1,285 @@
...
+        margin: 10,
+        padding: 10,
+      }}
+      src={'static/img/sacfair_screencaps/2.png'}
+      alt={'Events Tab'}
+      title={'Events Tab'}
+    />
+
+    <p>
+      There will be an existing event with your designated date & time titled
```
ezwang, line 75: You wan to use `&amp;` here, just HTML things.

frontend/pages/sacfairguide.tsx
```diff
@@ -0,0 +1,285 @@
...
+        margin: 10,
+        padding: 10,
+      }}
+      src={'static/img/sacfair_screencaps/3.png'}
+      alt={'Edit SAC Fair Event'}
+      title={'Edit SAC Fair Event'}
+    />
+    <p>
+      Edit the club details: you <strong>MUST</strong> provide a Zoom link
+      (preferably from your school Zoom account) where prospective students can
```
ezwang, line 91: Maybe talk about why school Zoom account, and the 40 minute meeting limit for free accounts.

frontend/pages/sacfairguide.tsx
```diff
@@ -0,0 +1,285 @@
...
+      page, or email <Contact /> with your Penn email and club name to be added
+      as an officer.
+    </p>
+    <img
+      style={{
+        maxHeight: 400,
+        verticalAlign: 'middle',
+        margin: 10,
+        padding: 10,
+      }}
```
ezwang, line 56: Use a wrapper to set all of the styles for these images, will be much cleaner.

frontend/pages/sacfairguide.tsx
ezwang: Also would be great to add screenshots of exactly what settings to enable, I know this has been a pain point in the past.

frontend/pages/sacfairguide.tsx
```diff
@@ -0,0 +1,285 @@
...
+    <Subheading>REQUIRED: Editing Your SAC Fair "Booth"</Subheading>
+
+    <p>
+      You will need to add a Zoom link to your club's SAC Fair event. Once you
+      navigate to your club, click the green "Manage Club" button on the
+      top-right corner of the page. If you do not see this button, it means you
+      are not an officer of the club. Either ask an officer listed on your club
+      to add you to the site through the "Membership" tab of the "Manage Club"
+      page, or email <Contact /> with your Penn email and club name to be added
+      as an officer.
```
ezwang, line 48: We should mention that events will appear in random order, except for one important thing.

Events that fill out their information completely (description, cover photo, etc) will appear above events that do not have their descriptions fully filled out.

frontend/pages/sacfairguide.tsx
ezwang: Ah, I see that this is mentioned below.

frontend/pages/sacfairguide.tsx
```diff
@@ -0,0 +1,285 @@
+import s from 'styled-components'
```
ezwang, line 1: We may want to name this something else, since we'll also have a guide for students. Maybe like `/sacfair/student` and `/sacfair/officer` or a guides folder or something? This works, just trying to future proof it when we'll have to add another guide.

PR #197, opened 2020-08-26
Fair Event Security

davish:
- Can't create an event with the activity fair type
- Can't change an existing event from or to that type
- Form.js will show DRF validation errors now
- Test to make sure I dont break things
- only show fair events on the /events page (TODO: remove this after fair)

### DIFF
--- a/backend/clubs/permissions.py
+++ b/backend/clubs/permissions.py
@@ -81,6 +81,23 @@ class EventPermission(permissions.BasePermission):
         else:
             return True
 
+    def has_object_permission(self, request, view, obj):
+        """
+        Do not allow transitions from and to club fair event.
+        """
+        FAIR_TYPE = 4
+        old_type = obj.type
+        new_type = request.data.get("type", old_type)
+
+        if view.action in ["update", "partial_update"]:
+            if old_type == FAIR_TYPE and new_type != FAIR_TYPE:
+                return False
+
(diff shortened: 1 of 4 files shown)


backend/clubs/permissions.py
```diff
@@ -81,6 +81,23 @@ def has_permission(self, request, view):
         else:
             return True
 
+    def has_object_permission(self, request, view, obj):
+        """
+        Do not allow transitions from and to club fair event.
+        """
+        FAIR_TYPE = 4
```
ezwang, line 88: @davish Don't redefine this constant, use `Event.FAIR` instead.

backend/clubs/permissions.py
```diff
@@ -81,6 +81,23 @@ def has_permission(self, request, view):
...
+        old_type = obj.type
+        new_type = request.data.get("type", old_type)
+
+        if view.action in ["update", "partial_update"]:
+            if old_type == FAIR_TYPE and new_type != FAIR_TYPE:
+                return False
+
+            if old_type != FAIR_TYPE and new_type == FAIR_TYPE:
```
ezwang, line 96: Is it possible to use XOR (`^`) here? This solution works fine and is very readable. I wonder if having the two values be integers would screw anything up.

Also, in general it is Pythonic to use `not a == b` instead of `a != b`.

PR #208, opened 2020-08-28
Add server side dynamic options

ezwang:
- Use the django-runtime-options to implement dynamic settings.
- A bit inefficient as this is loaded on every page, so we should take care to minimize the number of settings we create and ensure they're applicable to multiple pages. The efficiency loss may be negligible, idk.

### DIFF
--- a/frontend/utils.tsx
+++ b/frontend/utils.tsx
@@ -1,5 +1,5 @@
 import fetch from 'isomorphic-unfetch'
-import { ReactElement } from 'react'
+import { createContext, ReactElement, useContext } from 'react'
 
 import { MembershipRank } from './types'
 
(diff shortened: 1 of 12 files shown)


frontend/utils.tsx
```diff
@@ -18,14 +18,18 @@ export function stripTags(val: string): string {
...
+export const OptionsContext = createContext({})
+
+export function useSetting(key: string): string | number | boolean | null {
+  const options = useContext(OptionsContext)
+  const value = options[key] ?? null
+  if (value === 'true') {
+    return true
+  }
+  if (value === 'false') {
+    return false
```
ezwang, line 30: @davish I specify that these are the boolean type on the server, is there any reason why they aren't casted in the JSON response?

I'd be happy to change this in the library if it doesn't break anything else.

ezwang:
image

PR #215, opened 2020-08-30
Analytics

cphalen:
Changes:
- New `ClubVisit` model
- Multiple serializers for the `ClubVisit` model
- A viewset for the `ClubVisit` model
- A new detail on ClubViewSet for the club visits of that specific club
- Changes to the `/subscriptions` and `/favorites` details in the ClubViewSet
- Changes to the `SubscribeSerializer` and `FavoriteSerializer` (one new calculated field)
- New card for ClubEditPage called `AnalyticsCard.tsx`
- Changes to the ClubPage `index.tsx` to post to the backend every visit

Screenshot of analytics page:
image


backend/clubs/serializers.py
```diff
@@ -849,6 +850,23 @@ class Meta:
         fields = ("club", "person")
 
 
+class FavoriteTimeSerializer(serializers.ModelSerializer):
+    """
+    Used by users to get a list of clubs that they have favorited.
+    """
```
cphalen, line 856: I think I should have added an updated description for this method -- it's a new serializer because it also includes when the clubs were favorited. Also it's not for the user its for the club.

backend/clubs/migrations/0035_clubvisit.py
```diff
@@ -0,0 +1,39 @@
+# Generated by Django 3.1 on 2020-08-25 19:13
+
+from django.conf import settings
```
ezwang, line 3: Remember to merge the migrations before merging this branch, Git won't merge them for you.

backend/clubs/serializers.py
```diff
@@ -849,6 +850,24 @@ class Meta:
...
+    Used by clubs to get a list of people who have favorited their club as
+    well as when those people have favorited (for analytics).
+    """
+
+    person = serializers.HiddenField(default=serializers.CurrentUserDefault())
+    club = serializers.SlugRelatedField(queryset=Club.objects.all(), slug_field="code")
+    hour = serializers.SerializerMethodField("get_hour")
+
+    def get_hour(self, obj):
+        return obj.created_at.hour - 4
```
ezwang, line 864: Why are you subtracting 4 here?

backend/clubs/serializers.py
```diff
@@ -849,6 +850,24 @@ class Meta:
...
 
+class FavoriteTimeSerializer(serializers.ModelSerializer):
+    """
+    Used by clubs to get a list of people who have favorited their club as
+    well as when those people have favorited (for analytics).
+    """
+
+    person = serializers.HiddenField(default=serializers.CurrentUserDefault())
+    club = serializers.SlugRelatedField(queryset=Club.objects.all(), slug_field="code")
+    hour = serializers.SerializerMethodField("get_hour")
```
ezwang, line 861: Ideally, you should do this aggregation in SQL, which should be much faster than doing it in Python.

backend/clubs/serializers.py
```diff
@@ -849,6 +850,24 @@ class Meta:
         fields = ("club", "person")
 
 
+class FavoriteTimeSerializer(serializers.ModelSerializer):
+    """
+    Used by clubs to get a list of people who have favorited their club as
+    well as when those people have favorited (for analytics).
```
ezwang, line 856: We don't actually want to tell clubs who has marked their club as favorite, favorites should be for student purposes only.

If students want to give that information to the club, they can subscribe to the club.

backend/clubs/views.py
```diff
@@ -547,7 +553,46 @@ def subscription(self, request, *args, **kwargs):
...
+            serializer = SubscribeSerializer(
+                Subscribe.objects.filter(club=club, created_at__day=date.day), many=True
+            )
+        else:
+            serializer = SubscribeSerializer(Subscribe.objects.filter(club=club), many=True)
+
+        return Response(serializer.data)
+
+    @action(detail=True, methods=["get"])
+    def favorite(self, request, *args, **kwargs):
```
ezwang, line 567: See comment above, I don't think we want this, unless it is aggregated.

backend/clubs/views.py
```diff
@@ -547,7 +553,46 @@ def subscription(self, request, *args, **kwargs):
...
+        Return a list of all visits to a club page.
+        """
+        club = self.get_object()
+        if "date" in request.query_params.keys():
+            date = datetime.datetime.strptime(request.query_params["date"], "%Y-%m-%d")
+            serializer = ClubVisitSerializer(
+                ClubVisit.objects.filter(club=club, created_at__day=date.day), many=True
+            )
+        else:
+            serializer = ClubVisitSerializer(ClubVisit.objects.filter(club=club), many=True)
```
ezwang, line 594: We don't want clubs to see who visited their page individually, I think that's a little too much. Aggregate statistics with anonymous information should be good.

backend/clubs/management/commands/send_emails.py
```diff
@@ -198,9 +208,9 @@ def handle(self, *args, **kwargs):
                                     invite.send_owner_invite()
                                 else:
                                     invite.send_mail()
-                        elif action == "fair":
+                        elif action == "physical_fair":
                             send_fair_email(club, receiver)
-                        elif action == "postfair":
+                        elif action == "physical_postfair":
```
ezwang, line 213: I'm not sure why we're seeing this in the diff, perhaps you need to rebase on master again?

backend/clubs/models.py
```diff
@@ -159,6 +159,30 @@ class Club(models.Model):
...
+
+        emails = self.get_officer_emails()
+
+        if emails:
+            send_mail_helper(
+                name="fair_info",
+                subject="[ACTION REQUIRED] SAC Fair Setup and Information",
+                emails=emails,
+                context=context,
+            )
```
ezwang, line 184: Also not sure why we're seeing this in the diff, it should all be in master already.

backend/clubs/serializers.py
```diff
@@ -998,6 +999,26 @@ class Meta:
...
+
+    class Meta:
+        model = ClubVisit
+        fields = ("club", "person")
+
+
+class UserClubVisitWriteSerializer(UserSubscribeSerializer):
+    club = serializers.SlugRelatedField(queryset=Club.objects.all(), slug_field="code")
+
+    class Meta(UserClubVisitSerializer.Meta):
```
ezwang, line 1018: I'm pretty sure you can exclude the Meta class and it'll pick up the super one. Not 100% sure though, but it works when I tried it with events.

backend/clubs/views.py
```diff
@@ -173,6 +178,20 @@ def filter_note_permission(queryset, club, user):
     return queryset
 
 
+def hour_to_string_helper(hour):
+    hour_string = ""
+    if hour == 0:
+        hour_string = "12am"
+    elif hour < 12:
+        hour_string = str(hour) + "am"
```
ezwang, line 186: It is generally good practice to do `f"{hour}am"` or `"{}am".format(hour)` instead of `str(hour) + "am"`.

backend/clubs/views.py
```diff
@@ -513,6 +532,58 @@ def subscription(self, request, *args, **kwargs):
...
+        favorites_data = []
+        subscriptions_data = []
+        max_value = 0
+
+        for hour in range(24):
+            visits_data.append({"x": hour_to_string_helper(hour), "y": 0})
+            favorites_data.append({"x": hour_to_string_helper(hour), "y": 0})
+            subscriptions_data.append({"x": hour_to_string_helper(hour), "y": 0})
+
+        visits = ClubVisit.objects.filter(club=club, created_at__day=date.day)
```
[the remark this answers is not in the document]
cphalen: Ok I'm gonna leave this until later I think because my main priority is just having this be functional for SAC fair. This is something I will optimize in the future though because I have some ideas for it.

backend/clubs/views.py
```diff
@@ -515,6 +533,110 @@ def subscription(self, request, *args, **kwargs):
...
+        analytics_dict = {
+            "visits": visits_data,
+            "favorites": favorites_data,
+            "subscriptions": subscriptions_data,
+            "max": max_value,
+        }
+
+        return HttpResponse(json.dumps(analytics_dict))
+
+    @action(detail=True, methods=["get"])
```
ezwang, line 588: @cphalen You seem to have duplicated the `analytics` method twice.

frontend/pages/club/[club]/index.tsx
```diff
@@ -91,6 +91,15 @@ const ClubPage = ({
...
   }
 
+  useEffect(() => {
+    doApiRequest('/clubvisits/?format=json', {
+      method: 'POST',
+      body: {
+        club: code,
+      },
+    })
+  })
```
Comment (ezwang) [inline frontend/pages/club/[club]/index.tsx:101]: @cphalen This will send club visit requests to the server in an infinite loop, since you did not specify the second argument of `useEffect`. Please modify this to have `[]` as the second argument.

frontend/utils.tsx
```diff
@@ -1,5 +1,6 @@
 import fetch from 'isomorphic-unfetch'
 import { createContext, ReactElement, useContext } from 'react'
+import FormData from 'form-data'
```
ezwang, line 3: @cphalen Remove this if you're not using it, seems to be accidentally added.

ezwang:
image

@cphalen Looks amazing! Fantastic work!

- When you un bookmark or un subscribe from a page, this removes it from the graph. This behavior could potentially be confusing.
- Would be nice to have the chart resize to the box width, but that can be done in a later commit.
- Would also be cool to select a date and time range, instead of a single day. It would also be nice to select the granularity of the reporting, as I suspect that clubs may be interested in the 3 hours in particular when the SAC fair happens and see more complex breakdowns there.

PR #220, opened 2020-09-02
Use uvicorn instead of daphne

ezwang:
- Switch to Uvicorn, which should be faster and supports multiple workers.

### DIFF
--- a/backend/Pipfile
+++ b/backend/Pipfile
@@ -42,6 +42,7 @@ social-auth-app-django = "*"
 django-redis = "*"
 channels-redis = "*"
 uwsgi = {version = "*", sys_platform = "== 'linux'"}
+uvicorn = "*"
 
 [requires]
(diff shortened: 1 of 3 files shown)


backend/scripts/django-run
```diff
@@ -7,4 +7,4 @@
 cd /app/backend/
 
 # Run Daphne
-exec /usr/local/bin/daphne -b 0.0.0.0 -p 80 pennclubs.asgi:application
+exec /usr/local/bin/uvicorn --host 0.0.0.0 --port 80 --workers 3 pennclubs.asgi:application
```
davish, line 10: Looking at the Uvicorn page, looks like it mentions running uvicorn through gunicorn to manage processes:

https://www.uvicorn.org/deployment/

No anchor tag, but scroll down to "using a process manager".

PR #228, opened 2020-09-22
[WIP] Rework searchbar/filtering

sigmachirality:
Searchbar and filtering files need to be restructured, there's too much stuff going on in Searchbar. Additionally, the current structure does not work well with any mobile redesigns we might want to do down the road.


ezwang:
@sigmachirality can I close this?

Unfortunate, but I think this is too out of date with the current master branch. We'll have to do a redesign of the search bar to work with mobile pages at some point, but I think we'll have to restart at that point.

sigmachirality:
Yeah, that's fair, go ahead.

PR #230, opened 2020-10-02
[Feature] Subscribe button in cards

esinx:
Added subscribe button in club cards!

There are some layout changes (the Wrapper component adds margin to its not-first children = icons) but it's quite flexible, we could add more icons later if we need to.

Screenshot below for preview

Screen Shot 2020-10-03 at 7 53 45 AM


ezwang (approved):
LGTM! Fantastic work! 🎉 

Just noticed that there's a state bug when switching between list mode and card mode. Was not introduced in this PR, so we can fix it in a later PR.

PR #235, opened 2020-10-07
Start date

alnasir7:
Calculate the time difference from a new moment object instead of the original one. previously, the startOf function was applied to the string that is being displayed.


ezwang (approved):
LGTM!

PR #241, opened 2020-10-13
Parent child perms

rohan-gupta8:
Added functionality to provide parents with similar functionality to children. Role comparison is the same as that for children. Uses membership instance with highest authority if multiple found.

[ch2699]


backend/clubs/permissions.py
```diff
@@ -78,9 +119,8 @@ def has_permission(self, request, view):
                 return False
             if not request.user.is_authenticated:
                 return False
-            membership = Membership.objects.filter(
-                person=request.user, club__code=view.kwargs["club_code"]
-            ).first()
+            obj = Club.objects.get(code=view.kwargs["club_code"])
+            membership = find_membership_helper(request.user, obj)
```
ezwang, line 123: Generally, a good goal is to minimize the number of SQL queries that you perform. Each query adds a nonnegligible amount of delay to the loading time of the request. In fact, there are some cases where performing extra work on the SQL side in order to reduce the number of queries can make your program faster overall.

With this change, you go from 1 query for membership checks to at least 3 queries for a membership check. The code fetches this object from the database just to match the existing input format, which is okay but not optimal for performance. It should be possible to reduce this down to 1 query, but that would not be too easy.

The code right now is fine since refactoring this to be slightly more efficient would be nontrivial effort.

backend/clubs/permissions.py
```diff
@@ -263,9 +297,8 @@ def has_object_permission(self, request, view, obj):
                 return True
 
             # otherwise, allow club officers to edit and delete comments
-            membership = Membership.objects.filter(
-                person=request.user, club__code=view.kwargs["club_code"]
-            ).first()
+            object = Club.objects.get(code=view.kwargs["club_code"])
```
ezwang, line 300: Instead of doing this, it might be more efficient to use the `obj` object, which is a `QuestionAnswer` model.

cphalen (approved):
Everything looks good to me! Really awesome work on this

ezwang (approved):
LGTM!

Would fix up the rest of the `obj = Club.objects.get(code=view.kwargs["club_code"])` to use the permission `obj` where possible, but other than that looks ready to merge.

backend/clubs/models.py
```diff
@@ -873,6 +873,7 @@ class Profile(models.Model):
 
     user = models.OneToOneField(get_user_model(), on_delete=models.CASCADE, primary_key=True)
     image = models.ImageField(upload_to=get_user_file_name, null=True, blank=True)
+    uuid_secret = models.UUIDField(default=uuid.uuid4, editable=False)
```
ezwang, line 876: Consider making this `editable`, you might want to provide a way for the user to regenerate their UUID secret if it is compromised.

Some tools, like Canvas, have this functionality available.

backend/clubs/serializers.py
```diff
@@ -1088,7 +1088,7 @@ class Meta(ClubListSerializer.Meta):
 
 class FavoriteSerializer(serializers.ModelSerializer):
     """
-    Used by users to get a list of clubs that they have favorited.
+        Used by users to get a list of clubs that they have favorited.
```
ezwang, line 1091: This usually isn't supposed to be indented.

backend/clubs/views.py
```diff
@@ -1526,6 +1528,33 @@ def get_queryset(self):
...
+class FavoriteCalendarAPIView(APIView):
+    """
+        get: Return a .ics file of the user's favorite clubs' events.
+    """
+
+    def get(self, request, *args, **kwargs):
+        calendar = ICSCal()
+        calendar.extra.append(ICSParse.ContentLine(name="X-WR-CALNAME", value="Penn Clubs Events"))
+        favs = Favorite.objects.filter(person__profile__uuid_secret=kwargs["user_secretuuid"])
+        clubs = {fav.club for fav in favs}  # set contains distinct elements
```
ezwang, line 1540: You can replace this with `Club.objects.filter(favorite_set__person__profile__uuid_secret=kwargs["user_secretuuid"])`. You may or may not need to add a `.distinct()` and it might be named `favorites` instead of `favorite_set`.

ezwang: Actually, even better, do `Event.objects.filter(club__favorite_set__person__....)` given the logic below.

Your current club lookup is `O(n)` in terms of number of SQL queries, and the events lookups are also `O(n)` in terms of SQL queries. You can eliminate some of this using `prefetch_related`, but properly refactoring this would be the best solution.

backend/clubs/views.py
```diff
@@ -1526,6 +1528,31 @@ def get_queryset(self):
...
+        all_events = Event.objects.filter(
+            club__favorite__person__profile__uuid_secret=kwargs["user_secretuuid"]
+        ).distinct()
+        for event in all_events:
+            e = ICSEvent()
+            e.name = event.club.name + ": " + event.name
+            e.begin = event.start_time
+            e.end = event.end_time
+            e.location = event.location
+            e.description = ("" if event.url is None else event.url) + "\n\n" + event.description
```
ezwang, line 1549: You can simplify this to `(event.url or "")`.

Also, its nice to use format strings. In this case, you can do `"{}\n\n{}".format(event.url or "", event.description)`.

backend/clubs/views.py
```diff
@@ -1526,6 +1528,31 @@ def get_queryset(self):
...
+    def get(self, request, *args, **kwargs):
+        calendar = ICSCal()
+        calendar.extra.append(ICSParse.ContentLine(name="X-WR-CALNAME", value="Penn Clubs Events"))
+
+        all_events = Event.objects.filter(
+            club__favorite__person__profile__uuid_secret=kwargs["user_secretuuid"]
+        ).distinct()
+        for event in all_events:
+            e = ICSEvent()
+            e.name = event.club.name + ": " + event.name
```
ezwang, line 1545: I kind of prefer the calendar name event format to use brackets for the club name, but that may be because Canvas is already in that format. I'll leave the final decision up to you!

image

PR #245, opened 2020-10-18
Searchbar refactor

ezwang:
- Refactor the searchbar to be more clear and extensible.
- Hide appropriate search bar items on Hub@Penn.

[ch2742]


esinx (approved):
lgtm! Not too crucial/from like two months ago, but I do have one question about SearchBar.tsx#44, 
```typescript
export const getInitialSearch = (): SearchInput => {
  return { ordering: 'featured' }
}
```
Perhaps this line is a little hardcoded? I do see that it's a failsafe option, but I think we are assuming that the objects we are filtering/sorting have the sort-by-feature logic. The default value could be changed from the client anyways, so I don't think it's a huge issue.

PR #251, opened 2020-10-19
[WIP]Recurring event

xiayufeilucy:


backend/clubs/models.py
```diff
@@ -428,6 +428,34 @@ class Testimonial(models.Model):
...
+    start_date = models.DateTimeField()
+    end_date = models.DateTimeField()
+    offset = models.IntegerField()
+
+    """
+    RECURRING_DATES = (
+        ('m', 'Monday'),
+        ('t', 'Tuesday'),
+        ('w', 'Wednesday'),
+        ('th', 'Thursday'),
```
davish, line 445: not sure if we want to stick to Penn's standard that they use for courses, where the days are each one letter (`M,T,W,H,F,S,U`)

PR #254, opened 2020-10-21
Resolved members n+1 slowdown issue

cphalen:
Literally wrote like 20 characters, but look at the number of queries decrease!

Problem was that we were using fields from the `Profile` model (image, has_been_prompted, graduation_year, etc.) in the `MembershipSerializer` without prefetching that model.

Screen Shot 2020-10-21 at 1 56 45 AM
Screen Shot 2020-10-21 at 1 57 42 AM

[ch2772]


ezwang (approved):
LGTM! Nice work!

PR #255, opened 2020-10-23
Added student type model and m2m field in Club

rohan-gupta8:
[ch2747]

Added new model to represent student type.

Model has m2m relationship with Club.

Added to populate to initialize data.

Returns student type only in Club view (not in Club List view to save bandwidth)

Frontend tasks pending


backend/clubs/migrations/0047_auto_20201023_1848.py
```diff
@@ -1,15 +1,28 @@
-# Generated by Django 3.1.2 on 2020-10-18 21:41
+# Generated by Django 3.1.2 on 2020-10-23 22:48
```
ezwang, line 1: Since the `0047_profile_uuid_secret` migration is already in production, we may not want to delete this migration.

It is much easier to delete our local development db than to modify the production db. Try to avoid modifying any migrations committed to `master`, as those are already in production.

backend/clubs/models.py
```diff
@@ -636,6 +637,14 @@ class Note(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
 
 
+class StudentType(models.Model):
+
```
ezwang, line 641: A simple docstring here would be nice, to let people know what this is supposed to represent. For example:

```
Represents a student type that the club is intended for. For example, "International Students", "Transfer Students", or "Online Students".
```

backend/clubs/serializers.py
```diff
@@ -745,6 +745,7 @@ class Meta:
             "founded",
             "image_url",
             "is_favorite",
+            "favorite_count",
```
ezwang, line 748: Why was this line added? Not sure if good or bad, just curious.

rohan-gupta8: accidentally deleted it

rohan-gupta8: oh wait turns out it is redundant. I think rebasing might have fixed my deletion. will recommit without.

ezwang (approved):
LGTM!

esinx (approved):
This pr looks like a good guideline to adding new models :) looks very clear / good to me!

PR #258, opened 2020-10-24
Branding Google Analytics tracking code switch

cphalen:
Added the two tracking codes to the `branding.tsx` file and setup `analytics.tsx` to use the imported tracking code from branding.


ezwang (approved):
LGTM! Thanks for doing this!

PR #260, opened 2020-10-26
Add student type field frontend

ezwang:
- Add student type field to the frontend (club edit page, search sidebar).
- Only load student type field if enabled for the site (using branding constants).

image

[ch2747]


esinx (approved):
LGTM! Nice touch on conditional loading 👍

PR #265, opened 2020-10-28
signature-events

alnasir7:
- Added signature events field. The field is only rendered for resources.

Screen Shot 2020-10-28 at 5 49 39 PM

[ch2780]

### DIFF
--- a/frontend/pages/club/[club]/index.tsx
+++ b/frontend/pages/club/[club]/index.tsx
@@ -247,14 +247,11 @@ const ClubPage = ({
             </StyledCard>
           )}
           <Events data={events} />
-          {isClubFieldShown('signature_events') && (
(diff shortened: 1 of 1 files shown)


ezwang (approved):
LGTM!! Love the screenshot.

Just one minor change before merging. Instead of displaying the "No signature events for this for this resource" message, just hide the signature event field altogether.

frontend/pages/club/[club]/index.tsx
```diff
@@ -247,6 +247,14 @@ const ClubPage = ({
             </StyledCard>
           )}
           <Events data={events} />
+          {isClubFieldShown('signature_events') && club.signature_events && (
```
Comment (ezwang) [inline frontend/pages/club/[club]/index.tsx:250]: This check will actually pass if `club.signature_events` is an empty string (`""`). You need to check `club.signature_events.length` in order to properly catch this edge case.

PR #266, opened 2020-10-29
Convert all forms to Formik

ezwang:
- Convert all forms to Formik.
- Make `ModelForm` class work with Formik, basic typescript conversions.
- A lot of types are `any` for now, need to fix them at some point.
- Forms look and work pretty much the same as they did before.
- Need to add in additional checks for `dirty` and `isSubmitting` for various forms, not done yet.

[ch1417]


backend/clubs/views.py
```diff
@@ -154,7 +154,12 @@ def upload_endpoint_helper(request, cls, field, save=True, **kwargs):
         return Response(
             {"detail": "No image file was uploaded!"}, status=status.HTTP_400_BAD_REQUEST,
         )
-    return Response({"detail": "{} image uploaded!".format(obj.__class__.__name__)})
+    return Response(
+        {
+            "detail": "{} image uploaded!".format(obj.__class__.__name__),
```
davish, line 159: f-string here?
```suggestion
            "detail": f"{obj.__class__.__name__} image uploaded!",
```

frontend/components/ClubEditPage/ClubEditCard.tsx
```diff
@@ -353,14 +378,88 @@ export default function ClubEditCard({
...
+
   return (
-    <Form
-      fields={fields}
-      defaults={Object.keys(club).length ? club : creationDefaults}
+    <Formik
+      initialValues={
+        Object.keys(club).length
+          ? doFormikInitialValueFixes(
+              Object.entries(club).reduce((acc, [key, value]) => {
```
davish, line 394: A comment here on what exactly this reduce is doing would be super helpful!

frontend/components/ClubEditPage/ClubEditCard.tsx
```diff
@@ -353,14 +378,88 @@ export default function ClubEditCard({
...
+            )
+          : creationDefaults
+      }
       onSubmit={submit}
-      enableSubmitWithoutEdit={true}
-    />
+      enableReinitialize
+    >
+      {({ dirty, isSubmitting }) => (
+        <FormFieldClassContext.Provider value="is-horizontal">
```
davish, line 411: Could this be factored out into a separate component? Could help with organization, not sure exactly what this block is without a name.

frontend/components/FormComponents.tsx
```diff
@@ -0,0 +1,416 @@
...
+import React, {
+  ReactElement,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react'
+import DatePicker from 'react-datepicker'
+import Select from 'react-select'
+import s from 'styled-components'
```
davish, line 16:
```suggestion
import styled from 'styled-components'
```
I know this hasn't been done consistently in the codebase, but I really think we should move towards `styled` rather than `s` -- it's recommended in the docs and also prettier will pick up CSS and lint it if it's inside a `styled` block but it doesn't know what `s` is!

frontend/components/FormComponents.tsx
```diff
@@ -0,0 +1,416 @@
...
+interface BasicFormField {
+  name: string
+  label?: string
+  helpText?: string
+  placeholder?: string
+  noLabel?: boolean
+}
+
+interface AnyHack {
+  [key: string]: any
```
davish, line 36: hehe -- maybe leave a comment as to why this hack is necessary?

frontend/components/FormComponents.tsx
```diff
@@ -0,0 +1,416 @@
...
+            ),
+          ),
+        )
+        textValue.current = props.value
+      }
+    }, [props.value])
+
+    return (
+      <div>
+        <Head>
```
davish, line 125: i really do love nextjs

frontend/components/FormComponents.tsx
```diff
@@ -0,0 +1,416 @@
...
+        <>
+          <div className="field-label">
+            <label className="label">{titleize(props.name)}</label>
+          </div>
+          <div className="field-body">{innerBody}</div>
+        </>
+      ) : (
+        innerBody
+      )}
+    </div>
```
davish, line 414: this file is awesome, in general tho some documentation on what each component does would be nice.

frontend/components/ModelForm.tsx
```diff
@@ -0,0 +1,515 @@
...
+  font-size: 1.2em;
+  margin-bottom: 0.75em;
+`
+
+type ModelFormProps = any
+
+type ModelFormState = any
+
+export const doFormikInitialValueFixes = (currentObject: {
+  [key: string]: any
```
davish, line 53: should the `AnyHack` type from the other file also be used here?

frontend/components/ModelForm.tsx
```diff
@@ -0,0 +1,515 @@
...
+    prev[key] = val
+    return prev
+  }, {})
+}
+
+/*
+ * Creates a form with CRUD (create, read, update, delete)
+ * capabilities for a Django model using a provided endpoint.
+ */
+export class ModelForm extends Component<ModelFormProps, ModelFormState> {
```
davish, line 71: Why class component?

frontend/components/ModelForm.tsx
```diff
@@ -0,0 +1,515 @@
...
+
+    this.state = {
+      objects: null,
+      currentlyEditing: null,
+      createObject:
+        props.defaultObject != null ? { ...props.defaultObject } : {},
+      newCount: 0,
+    }
+
+    this.onSubmit = this.onSubmit.bind(this)
```
davish, line 83: if you're using class components, can get around the bind issue by using lambda syntax! They take on the `this` from the parent automatically.

frontend/components/ModelForm.tsx
```diff
@@ -0,0 +1,515 @@
...
+    this.onDelete = this.onDelete.bind(this)
+    this.onCreate = this.onCreate.bind(this)
+    this.onChange = this.onChange.bind(this)
+  }
+
+  /**
+   * This is called when the create/edit form has its contents changed.
+   */
+  onChange(obj) {
```
davish, line 92:
```suggestion
  onChange = (obj) => {
```

davish: to get around bind()!

frontend/components/ModelForm.tsx
```diff
@@ -0,0 +1,515 @@
...
+                      <Icon name="trash" alt="trash" /> Delete
+                    </span>
+                    <ModelStatus status={object._status} />
+                  </FormFieldClassContext.Provider>
+                </Form>
+              )}
+            </Formik>
+          </ModelItem>
+        ))}
+        {allowCreation && (
```
davish, line 504: This definitely matches the current expectation, but I know in future dashboard designs we might be splitting out the list of existing entities from the new entity form. Would it make sense to factor that out here and then make another component that puts the two components together?

frontend/components/Settings/ProfileForm.tsx
```diff
@@ -30,94 +37,83 @@ const ProfileForm = ({
       .then(setMajors)
   }, [])
 
-  function submit(data) {
+  function submit(data, { setSubmitting, setStatus }) {
```
davish, line 40: Annotation with a return type would be useful to show it returns a promise

davish (changes requested):
Love this! some comments on documentation and whether or not this PR should include a few more refactorings. Let me know what you think!

frontend/components/FormComponents.tsx
ezwang: We currently have the entire codebase as `s` with no exceptions, so I think for this PR it would be nice to keep it that way.

However, I'm down to do a mass migration from `s` to `styled` at some point, should just be a massive find and replace.

frontend/components/ModelForm.tsx
ezwang: It used to be a class component, one of the last files that have not been converted over to a functional component yet.

No good reason except for it was that way before and we haven't gotten around to converting it yet.

frontend/components/ModelForm.tsx
ezwang: This is a neat trick! Thanks for letting me know about this.

frontend/components/FormComponents.tsx
ezwang: Created a task in [ch2813].

frontend/components/ModelForm.tsx
ezwang: Split it out a little bit, so that the list component and model form component are separate.

It would be more difficult to split out the ModelForm into a child component, since there are actually two views that the ModelForm can be in.

One is a table view where you can edit one item at a time and the other is a inline view where you can edit all of the objects simultaneously. You can switch between the views depending on whether or not you specify the table columns.

frontend/components/ModelForm.tsx
davish: That makes sense -- if you could leave a TODO to convert that would be great!

frontend/components/FormComponents.tsx
```diff
@@ -32,13 +37,49 @@ interface BasicFormField {
...
+ * to be passed to any of the form inputs.
+ *
+ * The ideal solution would be to explicitly define what
+ * props each form field accepts.
+ *
+ * This is used for now so that we can at least guarantee
+ * that a form field has the basic properties expected
+ * of all form fields.
+ */
 interface AnyHack {
```
davish, line 51: is there any benefit of `AnyHack` over `any` as a type?

davish (approved):
LGTM! One clarifying comment but let's get this merged

frontend/components/FormComponents.tsx
ezwang: I figured that it would be good to specify the purpose of this hack, so that its cleaner to cleanup later. You know which items are applicable to this particular hack and can clean them up easily when the time comes. Either way is okay, I think.

frontend/components/ModelForm.tsx
ezwang: TODO created! [ch2820]

frontend/components/FormComponents.tsx
ezwang: Created a task for cleanup. [ch2821]

PR #267, opened 2020-10-30
Display files

alnasir7:
- Uploaded files are displayed for members of any rank on the club view page.

Screen Shot 2020-10-30 at 3 37 59 AM


ezwang (approved):
LGTM! Nice work! 🎉

frontend/pages/club/[club]/index.tsx
```diff
@@ -212,6 +213,14 @@ const ClubPage = ({
             <StrongText ref={questionsScrollRef}>FAQ</StrongText>
             <QuestionList club={club} questions={questions} />
           </div>
+          {club.is_member !== false && club.files && club.files.length && (
```
Comment (ezwang) [inline frontend/pages/club/[club]/index.tsx:216]: Is there a reason you use `!== false` instead of just checking for true?

Comment (alnasir7) [reply frontend/pages/club/[club]/index.tsx:216]: Superusers have their is_member = 0. This will be evaluated as false, which isn't the behaviour I want.

PR #270, opened 2020-10-31
Add filter to events created by unapproved clubs

xiayufeilucy:


Pipfile
```diff
@@ -0,0 +1,8 @@
+[[source]]
```
ezwang, line 1: Remove this!

backend/clubs/views.py
```diff
@@ -1008,6 +1008,28 @@ def get_queryset(self):
...
+
+        now = timezone.now()
+        if view.action in ["list"]:
+            queryset = queryset.filter(end_time__gte=now)
+
+        return (
+            queryset.select_related("club", "creator", )
+                .prefetch_related("club__badges")
+                .order_by("start_time")
+        )
```
ezwang, line 1030: You should put all of the stuff that the user cannot control in the `get_queryset` method, and put all of the stuff you want the user to be able to alter in the event filter. This way, it makes it easier to use the filter for other purposes.

xiayufeilucy: So filter by club_code in get_queryset and filtering by time and approval of clubs in ValidEventFilter?

xiayufeilucy: So filter by club_code in get_queryset and filtering by time and approval of clubs in ValidEventFilter?

ezwang: This should actually all be in the `get_queryset` method, actually. You want to put the `start_time` and `end_time` filters in here that filter off of GET parameters.

Pipfile
ezwang: @xiayufeilucy The wrong Pipfile was removed. You want to remove the one in the root directory, not the `backend` folder.

PR #276, opened 2020-11-01
New H@P banner design

snie11:
- Banner + overlayed PNG.
- only on H@P, not for PennClubs. works for all pages (hopefully)

Screen Shot 2020-10-31 at 10 22 40 PM


ezwang (approved):
LGTM!

Header is extremely large now, not sure if this is what Hub@Penn actually had in mind when they specified this. Happy to merge this for now and change it up later if necessary.

PR #277, opened 2020-11-01
FYH Email templates

snie11:
- changed wording, copy from FYH box page
- Unsure if this will work with base email directory atm, wasn't able to see previews on /api/emailpreview 

[ch2822]


ezwang (approved):
LGTM! Merging this in now so we can preview these templates with the branding switch.

PR #278, opened 2020-11-01
Added Recruitment Cycle Field

rohan-gupta8:
* Added initialisation data to populate
* Added an IntegerField called “recruiting_cycle” with 5 choices to club – Unknown, Fall, Spring, Both, Open
* Added to admin panel; included as filter option for admin panel
* Added to frontend filter logic
* Added to both clublist and club serializers


backend/clubs/migrations/0054_auto_20201101_1152.py
```diff
@@ -0,0 +1,55 @@
...
+                    (1, "Open Membership"),
+                    (3, "Audition Required"),
+                    (2, "Tryout Required"),
+                    (4, "Application Required"),
+                    (5, "Application and Interview Required"),
+                ],
+                default=4,
+            ),
+        ),
+        migrations.AlterField(
```
ezwang, line 41: These entries shouldn't appear in this commit. Perhaps you need to delete this commit, rebase with master, and regenerate this commit?

ezwang: Ah, never mind, I see that this is a typo fix. Nice!

ezwang (approved):
LGTM! Great work!

backend/clubs/migrations/0054_auto_20201101_1152.py
rohan-gupta8: I changed a type in "Intereview" to "Interview".

rohan-gupta8: Is it still unintended?

PR #282, opened 2020-11-03
Add recruiting cycle field to frontend

ezwang:
- Add recruiting cycle field on Penn Clubs (not on Hub@Penn).

image
image

[ch2842]

### DIFF
--- a/frontend/components/ClubEditPage/ClubEditCard.tsx
+++ b/frontend/components/ClubEditPage/ClubEditCard.tsx
@@ -5,6 +5,7 @@ import { ReactElement } from 'react'
 import {
   Club,
   ClubApplicationRequired,
+  ClubRecruitingCycle,
   ClubSize,
   Major,
   School,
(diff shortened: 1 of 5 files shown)


rohan-gupta8:
LGTM! Quick question, if we were to modify recruiting cycles and say add something like 'Summer' we would have to modify the `const` right? Is there some for-loop implementation we could do over `Club.RECRUITING_CYCLES` or is that information not available to the frontend?

ezwang:
> LGTM! Quick question, if we were to modify recruiting cycles and say add something like 'Summer' we would have to modify the `const` right? Is there some for-loop implementation we could do over `Club.RECRUITING_CYCLES` or is that information not available to the frontend?

Yup, we would have to modify the frontend! We could try syncing the backend values to the frontend, but this might not be a trivial task. We would want to do this in a way that is systematic enough to be used with all of our other fields. We would need to use Python introspection to find out which values exist and write a code generator for Javascript that would create these fields. We would then need to remember to run the script every time a backend field is updated or write a test case to ensure that it is never forgotten.

This isn't an issue with the many to many fields, since we fetch the available options when the page is loaded.

PR #283, opened 2020-11-05
Filter events from unapproved clubs

xiayufeilucy:
- Event from unapproved clubs are filtered

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -1227,6 +1227,8 @@ class EventViewSet(viewsets.ModelViewSet):
         if self.kwargs.get("club_code") is not None:
             qs = qs.filter(club__code=self.kwargs["club_code"])
 
+        qs = qs.filter(club__approved=True)
+
         now = timezone.now()
         if self.action in ["list"]:
             qs = qs.filter(end_time__gte=now)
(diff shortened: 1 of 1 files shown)


backend/clubs/views.py
```diff
@@ -1227,6 +1227,8 @@ def get_queryset(self):
         if self.kwargs.get("club_code") is not None:
             qs = qs.filter(club__code=self.kwargs["club_code"])
 
+        qs = qs.filter(club__approved=True)
```
ezwang, line 1230: You can do `Q(club__approved=True) | Q(club__ghost=True)`!

What the `ghost` parameter does is handle the edge case where changes to a club are pending approval. The club is marked as approved, but the newest version of the club changes have not yet been reviewed.

We handle this by introducing a `ghost` boolean parameter. When a club is edited, `approved=False` and `ghost=True`. When retrieving data for clubs, if `ghost=True`, instead of retrieving the data from the club object, retrieve data from the club history object with the latest approved version. When the club is approved, set `approved=True` and `ghost=False`.

xiayufeilucy: Makes sense! Add to the filter

ezwang (approved):
LGTM! 🎉

PR #286, opened 2020-11-06
Refactor form

alnasir7:
- Refactored  ModelForm to functional component.

- Refactored s to styled everywhere in the code base.


alnasir7:
I played around with the site for a while and works the same as before. I am not very sure that everything works since I changed 70+ files. Does passing the tests means that it is working? at least with issues that could result from renaming variables since these should be easy to detect?

ezwang (approved):
LGTM! Thanks for taking the effort to do this.

The test cases can catch any glaring issues with your changes, such as if a particular page is broken. Tested the model form and page components briefly, everything looks good to me.

PR #287, opened 2020-11-06
ICS Calendar Export UI

esinx:
The modal UI is ready, just need some more detail on how ics export will work and we'll be good to go.

* looking for a clean way to implement copy; should we just stick with `navigator.clipboard api`?
* need confirmation on the directions to be written!

Screen Shot 2020-11-07 at 1 11 43 AM

[ch2741]

### DIFF
--- a/frontend/components/EventPage/SyncModal.tsx
+++ b/frontend/components/EventPage/SyncModal.tsx
@@ -6,7 +6,6 @@ import {
   CLUBS_GREY,
   CLUBS_NAVY,
   WHITE,
-  WHITE_ALPHA,
 } from '../../constants/colors'
 
 const ModalContainer = styled.div`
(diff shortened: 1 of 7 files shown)


frontend/components/EventPage/SyncModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <div
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+          marginTop: '10px',
+        }}
+      >
+        <b>URL: </b>
+        <URLCell readOnly value={'https://pennclubs.com/abc'} />
+        <URLCopyButton>Copy</URLCopyButton>
```
ezwang, line 65: Instead of doing this yourself, you can use [Bulma](https://bulma.io/documentation/form/general/#form-addons), the current CSS library we're using.

This might save you some time with formatting, and will make it look a little more consistent with the rest of the site.

We also use [Feather Icons](https://feathericons.com/) for icons. You can use an icon with `<Icon name="x" />`. If the icon svg isn't already downloaded, you'll need to download it and put it in the right place.

ezwang:
> looking for a clean way to implement copy; should we just stick with navigator.clipboard api?

It looks like support for `navigator.clipboard` and `document.execCommand` are pretty good ([source](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard) and [source](https://caniuse.com/document-execcommand)). You could choose one of them to use, or if you really want to be sure use the other as a fallback.

I'm fine if we don't support internet explorer, though the `document.execCommand("copy")` method should support this.

ezwang (changes requested):
Great work!

It would be nice to add some kind of visual indicator to indicate that the copy was successful after you have pressed the copy button. A lot of apps have a "Copied!" popover that briefly occurs in this scenario.

It would also be nice to have a few links to help pages for Google Calendar and iCalendar with walkthroughs on how to import the calendar, especially since some of our users are not tech savvy.

Perhaps the following icon would be better than the plus icon? It also might be good to rename "Sync" to "Export" to give people a better idea of what the functionality does.
image

Let me know which items you're willing to fix, I'd be happy to assist with any of the suggestions mentioned above.

frontend/pages/events.tsx
```diff
@@ -58,6 +61,11 @@ const CardList = styled.div`
   }
 `
 
+const iconStylesDark = {
+  transform: 'translateY(0px)',
+  opacity: 0.6,
+  color: `${WHITE_ALPHA(0.8)} !important`,
+}
```
ezwang, line 68: Remove this variable since it doesn't seem to be used anywhere.

frontend/pages/events.tsx
```diff
@@ -123,37 +125,55 @@ const ViewContext = createContext<
...
-        viewOption === EventsViewOption.CALENDAR ? 'is-selected is-info' : ''
-      }`}
-      aria-label="switch to calendar view"
-      onClick={() => {
-        setViewOption(EventsViewOption.CALENDAR)
-      }}
-    >
-      <Icon name="calendar" alt="calendar view" />
+      <Icon name="refresh" style={{ marginRight: '10px' }} />
+      Export
```
ezwang, line 176: Not sure if this is intentional or you just want a space between the icon and the text, you can do `{' '}` to force React to acknowledge a space here.

ezwang:
Latest look:

image

PR #288, opened 2020-11-06
Events with actual calendar view

esinx:
A lot of stuff going on in this pr!

Currently we have:
* [x] A proper calendar view 
* [x] Option to switch between day / week / month / agenda view
* [ ] Popup for event preview (not perfect though, needs polishing)
* [x] Use card instead of modal for event navigation

Upcoming, we have:
* [ ] Animation for displaying popups
* [ ] Determine popup regions (the popup component isn't really perfect just yet)
* [x] Update testing; the expected page is quite different and we should add new testing to make this work


esinx:
Change in checklist: we reverted back to modals, and this branch is now ready to be merged!

ezwang (approved):
LGTM, fantastic work! 🎉 

It would be good to use Bulma components for the header buttons instead of custom components.

image

There are properties like `is-large` that can control the size of buttons and make them bigger. Connected buttons are also a thing in Bulma. Using the built-in color schemes `is-primary`, `is-info`, `is-light`, etc can produce a result that is pretty similar to what you have now.

ezwang:
Not sure how difficult this might be, but it would be nice to add a small white border to events like you see in Google Calendar. If there are multiple events of the same type on the same time slot, it might be a bit difficult to see.

image

frontend/pages/events.tsx
```diff
@@ -39,6 +56,151 @@ const CardList = styled.div`
...
+        onClick: () => {
+          setViewOption(EventsViewOption.CALENDAR)
+        },
+      },
+    ]}
+  />
+)
+
+const CalendarHeader = ({
+  // date,
```
ezwang, line 148: @esinx If this won't be used in the near future, it'll be a good idea to remove it for now (the comment and the type annotation) to avoid confusion later on.

esinx:
Converted `SegmentedButtons` into bulma components & added borders in event view for more readability!

frontend/pages/events.tsx
esinx: Yup! Seems like it shouldn't be there.

frontend/pages/events.tsx
```diff
@@ -39,6 +52,161 @@ const CardList = styled.div`
...
+  return (
+    <StyledHeader>
+      <Title className="title" style={{ color: CLUBS_GREY, margin: 0 }}>
+        Events
+      </Title>
+      <div className="tools">
+        <div className="view-label">{label}</div>
+        <div className="buttons has-addons mt-0 mb-0">
+          <button
+            className={`button is-medium`}
```
ezwang, line 171: @esinx You can just do `className="button is-medium"` in these cases.

Minor issue, can merge without this change.

PR #289, opened 2020-11-07
Added backend endpoint to retrieve a user's UUID

rohan-gupta8:
[ch2843]

– Added backend endpoint to retrieve a user's UUID
– Used for ICS Calendar export

### DIFF
--- a/backend/clubs/serializers.py
+++ b/backend/clubs/serializers.py
@@ -1411,7 +1411,6 @@ class UserSerializer(serializers.ModelSerializer):
             )
         return value
 
-
     def get_image_url(self, obj):
         if not obj.profile.image:
             return None
--- a/backend/clubs/serializers.py
+++ b/backend/clubs/serializers.py
@@ -1143,6 +1143,21 @@ class UserMembershipSerializer(serializers.ModelSerializer):
         fields = ("club", "role", "title", "active", "public")
 
 
+class UserUUIDSerializer(serializers.ModelSerializer):
+    """
+    Used to get the uuid of a user (for ICS Calendar export)
+    """
+
+    uuid = serializers.SerializerMethodField("get_uuid")
+
+    def get_uuid(self, obj):
+        return f"{settings.DEFAULT_DOMAIN}/api/calendar/{str(obj.profile.uuid_secret)}"
+
+    class Meta:
+        model = get_user_model()
+        fields = ("uuid",)
+
+
 class UserSubscribeSerializer(serializers.ModelSerializer):
     """
     Used by users to get a list of clubs that they have subscribed to.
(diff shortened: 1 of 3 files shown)


backend/clubs/serializers.py
```diff
@@ -1143,6 +1143,21 @@ class Meta:
         fields = ("club", "role", "title", "active", "public")
 
 
+class UserUUIDSerializer(serializers.ModelSerializer):
+    """
+    Used to get the uuid of a user (for ICS Calendar export)
+    """
+
+    uuid = serializers.SerializerMethodField("get_uuid")
```
ezwang, line 1151: @rohan-gupta8 A slightly more DRF way to do this is to use code like something below:

```python
username = serializers.CharField(source="person.username", read_only=True)
```

Also, if you want to make this a bit easier for the frontend, you can return an entire URL instead of just the UUID to make it easier for frontend to implement this. Also more robust IMO.

rohan-gupta8: How do I return a URL?

backend/clubs/views.py
```diff
@@ -2121,6 +2122,21 @@ def get_object(self):
...
+    queryset = get_user_model().objects.all()
+    serializer_class = UserUUIDSerializer
+    permission_classes = [IsAuthenticated]
+    http_method_names = ["get"]
+
+    def get_object(self):
+        user = self.request.user
+        return user
+
+
```
ezwang, line 2139: An alternative to this could be making an action under `UserUpdateAPIView` instead of a brand new endpoint.

There are no problems with your code, but the proposed solution might be cleaner, since both endpoints share the same `queryset`, `permission_classes`, and `get_object` functionality.

rohan-gupta8: Oh ok I thought we wanted a separate endpoint just for this, but that makes sense.

backend/clubs/serializers.py
ezwang: https://github.com/pennlabs/penn-clubs/blob/master/backend/clubs/views.py#L616

^ Except instead of returning a QR code, just return a string instead. Ideally wrap the string in a JSON response, like `{url: "..."}`.

ezwang: Basically just return a string, where the string is a URL.

backend/clubs/urls.py
```diff
@@ -75,6 +76,7 @@
     path(r"settings/zoom/meetings/", MeetingZoomAPIView.as_view(), name="users-zoom-meeting"),
     path(r"settings/permissions/", UserPermissionAPIView.as_view(), name="users-permission"),
     path(r"clubs/<slug:club_code>/invite/", MassInviteAPIView.as_view(), name="club-invite"),
+    path(r"settings/calendar_url", UserUUIDAPIView.as_view(), name="user-uuid"),
```
ezwang, line 79: @rohan-gupta8 Add a slash to the end of this URL! I think Django will complain with a warning otherwise.

rohan-gupta8: Ok. Are we good to merge otherwise?

ezwang (approved):
LGTM! Great work! 🎉

PR #291, opened 2020-11-07
[Refactor] removed isHappening as a prop

esinx:
`isHappening` is used in `EventCard` and `EventModal` as a property, but it was quite redundant since we could also compare `new Date()` with `event.start_time` and `event.end_date` to see if an event is happening now or not.


ezwang (approved):
Looks great! Nice work! 🎉

PR #292, opened 2020-11-08
hub style fixes

snie11:
- header refactor to Anna's new image
- letter spacing in title
- reduce opacity on logo text background

image

[ch2835] [ch2863]


frontend/constants/measurements.ts
```diff
@@ -7,6 +7,7 @@ const MEASUREMENTS = {
     HEADER_SHADOW: 'default',
     TITLE_CLASS: '1.5rem',
     TITLE_MARGIN: '15px',
+    TITLE_SPACING: '0rem',
```
ezwang, line 10: Using the `inherit` keyword might be better in this case, not sure. I assume the default is something like `0rem` anyways.

ezwang (approved):
LGTM!

PR #301, opened 2020-11-14
Equality check for start and end times

rohan-gupta8:
- required for ICS module
- added 1 second delta if equal

[ch2885]


ezwang (approved):
LGTM! Good to merge in if tests pass.

PR #305, opened 2020-11-15
Populate Events

cphalen:
Updated test data for events:
- Spans two weeks forward and backward
- Alternates days with overlapping events (30 minute overlap) and zero overlap events
- 10 events for each day

[ch2853]


ezwang (approved):
LGTM!

PR #311, opened 2020-11-18
Require contact

alnasir7:
- Not Complete
- Disable continue if there are no public contacts while creating resources
- Add onUpdate prop for ModelForm to update parent component

[ch2838]

### DIFF
--- a/frontend/components/ClubEditPage/AdvisorCard.tsx
+++ b/frontend/components/ClubEditPage/AdvisorCard.tsx
@@ -3,7 +3,7 @@ import { ReactElement, useState } from 'react'
 import styled from 'styled-components'
 
 import { RED } from '../../constants/colors'
-import { Club } from '../../types'
+import { Advisor, Club } from '../../types'
 import {
   OBJECT_NAME_SINGULAR,
   SHOW_MEMBERS,
@@ -32,7 +32,9 @@ export default function AdvisorCard({
   const [advisorsCount, setAdvisorsCount] = useState<number>(
     club.advisor_set.length || 0,
   )
-  const updateAdvisors = (newAdvisors) => {
+  const updateAdvisors = (
+    newAdvisors: (Advisor & { _error_message?: string })[],
+  ): void => {
     let validCount = 0
     if (newAdvisors.length) {
       validCount = newAdvisors.filter(
@@ -87,7 +89,7 @@ export default function AdvisorCard({
           )}
           fields={fields}
         />
-        {SITE_ID === 'Hub@Penn' && (
+        {SITE_ID === 'fyh' && advisorsCount <= 0 && (
           <RequireText>At least one public contact is required*</RequireText>
         )}
       </BaseCard>
--- a/frontend/components/ClubEditPage/AdvisorCard.tsx
+++ b/frontend/components/ClubEditPage/AdvisorCard.tsx
@@ -11,7 +11,11 @@ import {
   SITE_ID
 =======
   SITE_ID,
+<<<<<<< HEAD
 >>>>>>> cleanup
+=======
+  SITE_NAME,
+>>>>>>> fix name
 } from '../../utils/branding'
 import { Text } from '../common'
 import { CheckboxField, TextField } from '../FormComponents'
@@ -107,7 +111,7 @@ export default function AdvisorCard({
 
       <BaseCard title="Internal Points of Contact">
         <Text>
-          These private points of contact will be shown to only {SITE_ID}{' '}
+          These private points of contact will be shown to only {SITE_NAME}{' '}
           administrators.
         </Text>
         <ModelForm
--- a/frontend/components/ClubEditPage/AdvisorCard.tsx
+++ b/frontend/components/ClubEditPage/AdvisorCard.tsx
@@ -1,11 +1,13 @@
 import { Field } from 'formik'
-import { ReactElement } from 'react'
-
-import { Club } from '../../types'
+import { ReactElement, useState } from 'react'
+import styled from 'styled-components'
+import { RED } from '../../constants/colors'
+import { Club} from '../../types'
 import {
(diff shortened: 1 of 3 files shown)


frontend/components/ModelForm.tsx
```diff
@@ -402,6 +415,7 @@ export const ModelForm = (props: ModelFormProps) => {
           onDelete={onDelete}
           deleteVerb={deleteVerb}
           noun={noun}
+          requireOne={requireOne}
```
ezwang, line 418: It might make more sense to change this to `minimumRequired` or similar and make it an integer field, which makes it more flexible for future purposes.

frontend/components/ClubEditPage/AdvisorCard.tsx
```diff
@@ -15,9 +17,36 @@ import BaseCard from './BaseCard'
 type Props = {
   club: Club
   schools: School[]
+  validateAdvisors?: (any) => void
```
ezwang, line 20: You can annotate this with `(valid: boolean) => void` or a similar type to be more accurate.

frontend/components/ClubEditPage/AdvisorCard.tsx
```diff
@@ -15,9 +17,36 @@ import BaseCard from './BaseCard'
...
+  margin-top: 1rem;
+`
+
+export default function AdvisorCard({
+  club,
+  schools,
+  validateAdvisors,
+}: Props): ReactElement {
+  const [advisorsCount, setAdvisorsCount] = useState<number>(
+    club.advisor_set.length || 1,
```
ezwang, line 34: Why do we assume that this is 1 if it is not set?

frontend/components/ClubEditPage/AdvisorCard.tsx
```diff
@@ -64,13 +94,19 @@ export default function AdvisorCard({ club, schools }: Props): ReactElement {
           )}
           fields={fields}
         />
+        {SITE_NAME === 'Hub@Penn' && (
```
ezwang, line 97: Ideally, use `SITE_ID` instead of `SITE_NAME`, just for consistency purposes. If we ever change up this check in the future, we can find all instances in the codebase by grepping for "SITE_ID".

frontend/components/ModelForm.tsx
alnasir7: I don't want to just pass the number because this way I will need to implement counting the number of valid objects on the form, which is not ideal since different uses might have different validations, or I will have to pass the number of valid objects also from above. So I think it is easier to just pass whether it meets the required minimum or not. I can rename it to some more expressive.

ezwang: Ah, I see you've removed this altogether. That's even better!

frontend/utils/branding.tsx
```diff
@@ -1,4 +1,4 @@
-const site = process.env.NEXT_PUBLIC_SITE_NAME || 'clubs'
+const site = process.env.NEXT_PUBLIC_SITE_NAME || 'fyh'
```
ezwang, line 1: @alnasir7 It looks like this was accidentally committed!

frontend/components/ModelForm.tsx
```diff
@@ -93,6 +94,7 @@ type ModelTableProps = {
   objects: ModelObject[]
   allowEditing?: boolean
   allowDeletion?: boolean
+  minimumRequired?: number
```
ezwang, line 97: @alnasir7 Remove this variable if you're not using it anymore!

ezwang:
@alnasir7 Changes look good to me! If you fix up the remaining comments, you'll be good to merge.

You'll definitely run into merge conflicts when rebasing, since I moved the advisor form to the first page. This merge conflict shouldn't be too bad to fix.

frontend/components/ClubEditPage/AdvisorCard.tsx
```diff
@@ -64,13 +94,18 @@ export default function AdvisorCard({ club, schools }: Props): ReactElement {
...
         />
+        {SITE_ID === 'fyh' && (
+          <RequireText>At least one public contact is required*</RequireText>
+        )}
       </BaseCard>
+
       <BaseCard title="Internal Points of Contact">
         <Text>
-          These private points of contact will be shown to only {SITE_NAME}{' '}
+          These private points of contact will be shown to only {SITE_ID}{' '}
```
ezwang, line 104: @alnasir7 You actually want to use `SITE_NAME` here, `SITE_NAME` is equal to "Hub@Penn", while `SITE_ID` is equal to "fyh".

ezwang: Use SITE_NAME for user facing names, use SITE_ID when doing comparison checks.

frontend/components/ClubEditPage/AdvisorCard.tsx
```diff
@@ -14,9 +17,35 @@ import BaseCard from './BaseCard'
...
+const RequireText = styled.p`
+  color: ${RED};
+  margin-top: 1rem;
+`
+
+export default function AdvisorCard({
+  club,
+  validateAdvisors,
+}: Props): ReactElement {
+  const [advisorsCount, setAdvisorsCount] = useState<number>(
```
ezwang, line 32: @alnasir7 You should remove this state variable if you're not using it anywhere else.

ezwang: It could be nice to only display the "At least one public contact is required*" text only if there is an insufficient number of advisors, to make it more apparent when the condition is met.

frontend/components/ModelForm.tsx
ezwang: @alnasir7 Reopening this because it seems like this still exists without being used.

PR #315, opened 2020-11-21
Added end-to-end test case for ICS endpoint

rohan-gupta8:
- End-to-end test case to test database inconsistencies between production and local (resolution: uuid function behavior differed in local and in production)
- tests filter query's user retrieval and time filter
- retrieves ICS file, parses and checks if contents are appropriate

[ch2896]


ezwang (approved):
LGTM!

PR #333, opened 2020-11-26
Expanded club visit model

cphalen:
Expanded the club visit model to have the five following types:
- Club Page Visit
- Event Modal Visit
- Event Link Clicked
- Manage Page Visit
- Fair Page Visit

And the frontend functionality to log these events. Happy to make additional changes, just lmk. Happy Thanksgiving!

[ch2861]


ezwang (approved):
LGTM! Great work! 🎉 

Note that this will distort the current analytics shown on the club management page. You could add a filter for those endpoints to only show club page visits for now.

cphalen:
> LGTM! Great work! 🎉
> 
> Note that this will distort the current analytics shown on the club management page. You could add a filter for those endpoints to only show club page visits for now.

This is a great point! I added a filter in the two views that we access the `ClubVisit` model, so hopefully that will solve the problem.

PR #340, opened 2020-12-03
Remove special case in club list queryset

davish:
Removes pending clubs that a user is a member of from the club list.

Not sure if this will fix the issue, but if it's a SQLite vs postgres issue this might be the one way to check easily for it.

This PR means that club owners won't be able to see their pending clubs in the main list anymore, but they'll still see it on their user profile page.


ezwang (approved):
What a weird issue. Thanks for the PR!

PR #342, opened 2020-12-04
Add model for activities fair

ezwang:
- Add a model for the activities fair so we don't need to worry about manually turning stuff on and off.
- There's a lot of room for expansion for the model. For instance, we could build a UI for clubs to register for activities fairs, allow fair organizers to view and control various information about the fair, and so on.
- Cram a bunch of changes into this PR.

image
image

[ch2942] [ch2940]


backend/clubs/models.py
```diff
@@ -490,6 +490,30 @@ def __str__(self):
...
+    organization = models.TextField()
+    contact = models.TextField()
+    time = models.TextField()
+
+    # these fields are rendered as raw html
+    information = models.TextField()
+    registration_information = models.TextField()
+
+    start_time = models.DateTimeField()
+    end_time = models.DateTimeField()
```
rohan-gupta8, line 508: Do we need any validation on this? Additionally, is this entirely gonna be rendered as HTML or do we want to be able to put these fairs as events into a calendar or something?

backend/clubs/views.py
```diff
@@ -451,6 +456,53 @@ def filter_queryset(self, request, queryset, view):
...
+        # get register/unregister action status
+        status = request.data.get("status")
+        if isinstance(status, str):
+            status = status.strip().lower() == "true"
+        elif not isinstance(status, bool):
+            status = True
+
+        # check if user can actually register club
+        mship = find_membership_helper(request.user, club)
+        if mship is not None and mship.role <= Membership.ROLE_OFFICER or request.user.is_superuser:
```
rohan-gupta8, line 492: Small stylistic comment:` if mship and ... `? not sure what it's like in the rest of the file

backend/clubs/views.py
```diff
@@ -2352,6 +2404,45 @@ def get_used_variables(self):
         return sorted(self._called)
 
 
+class OptionListView(APIView):
```
rohan-gupta8, line 2407: What does this do?

backend/clubs/views.py
ezwang: If the officer is an owner, then the value will be `0`, which will make this if statement fail instead of passing. This is why I explicitly check for None instead of all falsey values.

backend/clubs/models.py
ezwang: Currently, there is no way to create a `ClubFair` unless you use the Django admin dashboard. For the time being, I'm planning for us to just manually enter the information for the club fairs we will be doing.

If we allow users to create the fair in the future, the validation would be found in the serializer.

backend/clubs/views.py
ezwang: Penn Labs has created a package called [django-runtime-options](https://pypi.org/project/django-runtime-options/), which allows you to set options in the Django admin. This is a bit cleaner than having to make a code change and PR every time we want to tweak a setting, like the status of fair registration.

Since we're now loading this information from a model and appropriately calculating some of the fields, I override the options endpoint to return the correct value for the fair options. An alternative solution would be to have another endpoint to return fair specific information, but since this endpoint would be called every time a page is loaded on Penn Clubs, I figure that it would be easier to consolidate all of the site wide variables into one endpoint.

PR #343, opened 2020-12-04
Create recurring events

xiayufeilucy:
1.  Frontend should be able to create recurring events by posting data to /api/events or /api/clubs/<club-name>/events endpoints
2. Frontend needs to specify is_recurring, offset, and end_date

[ch2601]


Pipfile
```diff
@@ -0,0 +1,8 @@
+[[source]]
```
ezwang, line 1: Please remove this file! This is the Pipefile in the root folder, not in `backend/Pipfile`. This file is created when you accidentally run `pipenv` in the root folder instead of the backend folder.

backend/clubs/migrations/0059_recurringevent.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.1.2 on 2020-12-04 21:59
```
ezwang, line 1: You can merge these two migrations by deleting them and running `./manage.py makemigrations` again.

backend/clubs/models.py
```diff
@@ -490,6 +490,17 @@ def __str__(self):
         return self.text
 
 
+class RecurringEvent(models.Model):
+    """
+    Represents a recurring event hosted by a club.
+    """
+
+    recurring_event_id = models.AutoField(primary_key=True)
```
ezwang, line 498: You don't need to do this, by default Django models have an `id` field that is a primary key and auto incrementing integer.

backend/clubs/views.py
```diff
@@ -1198,9 +1199,35 @@ def create(self, request, *args, **kwargs):
             raise DRFValidationError(
                 detail="Approved activities fair events have already been created. "
                 "See above for events to edit, and "
-                f"please email {settings.FROM_EMAIL} if this is en error."
+                "please email contact@pennclubs.com if this is en error."
```
ezwang, line 1202: Change this back to `settings.FROM_EMAIL`.

The reason is that Hub@Penn has a different email address and we use the same backend. The error messages should have the right email to contact.

backend/clubs/views.py
```diff
@@ -1198,9 +1199,35 @@ def create(self, request, *args, **kwargs):
...
+            event_data = request.data.copy()
+            start_time = parse(event_data.pop("start_time"))
+            end_time = parse(event_data.pop("end_time"))
+            offset = event_data.pop("offset")
+            end_date = parse(event_data.pop("end_date"))
+            event_data.pop("is_recurring")
+            club_code = event_data.pop("club")
+
+            club = Club.objects.get(code=club_code)
+            print(event_data)
```
ezwang, line 1216: Remove this `print` statement.

backend/clubs/views.py
```diff
@@ -1198,9 +1199,35 @@ def create(self, request, *args, **kwargs):
...
 
+        if request.data.get("is_recurring", None) is not None:
+            parent_recurring_event = RecurringEvent.objects.create()
+            event_data = request.data.copy()
+            start_time = parse(event_data.pop("start_time"))
+            end_time = parse(event_data.pop("end_time"))
+            offset = event_data.pop("offset")
+            end_date = parse(event_data.pop("end_date"))
+            event_data.pop("is_recurring")
+            club_code = event_data.pop("club")
```
ezwang, line 1213: This will allow anyone to create an event for any club, which is a security issue. You need to validate that the user can actually do this. To match other endpoints, you may want to read this from the URL, if applicable.

backend/clubs/views.py
```diff
@@ -1198,9 +1199,35 @@ def create(self, request, *args, **kwargs):
...
                 detail="Approved activities fair events have already been created. "
                 "See above for events to edit, and "
-                f"please email {settings.FROM_EMAIL} if this is en error."
+                "please email contact@pennclubs.com if this is en error."
             )
 
+        if request.data.get("is_recurring", None) is not None:
+            parent_recurring_event = RecurringEvent.objects.create()
+            event_data = request.data.copy()
+            start_time = parse(event_data.pop("start_time"))
```
ezwang, line 1208: Ideally, you would want to perform input validation on these items. This way, when the user enters something that is invalid, they get a nice error message instead of the endpoint returning a 500 and them having no idea what is going on.

This is a bit of effort, so you don't have to add it in this PR. You could redesign your API call to be a detail instead of through the main edit endpoint, which could be cleaner. In that case, you could have the detail endpoint use a custom `Serializer` (not a `ModelSerializer` since the creation is kind of weird) and use some of Django's built in parsing to make it easier for you.

ezwang (changes requested):
Great work! Thanks for adding the test case, that's always a great addition.

I left some comments, be sure to clean those up before merging!

xiayufeilucy:
> Great work! Thanks for adding the test case, that's always a great addition.
> 
> I left some comments, be sure to clean those up before merging!

Thanks Eric! Will fix the comments

backend/clubs/views.py
```diff
@@ -1201,6 +1202,31 @@ def create(self, request, *args, **kwargs):
...
+                    creator=request.user,
+                    club=club,
+                    start_time=start_time,
+                    end_time=end_time,
+                    parent_recurring_event=parent_recurring_event,
+                )
+                start_time = start_time + datetime.timedelta(days=offset)
+                end_time = end_time + datetime.timedelta(days=offset)
+
+            return Response({"success": True})
```
ezwang, line 1228: A potentially better response for this would be to return the serialized list of created clubs. This can be added in a future PR.

backend/clubs/views.py
```diff
@@ -1201,6 +1202,31 @@ def create(self, request, *args, **kwargs):
...
+            end_time = parse(event_data.pop("end_time"))
+            offset = event_data.pop("offset")
+            end_date = parse(event_data.pop("end_date"))
+            event_data.pop("is_recurring")
+            club_code = kwargs["club_code"]
+
+            club = Club.objects.get(code=club_code)
+            while start_time < end_date:
+                Event.objects.create(
+                    **event_data,
```
ezwang, line 1218: This will actually allow you to bypass the permissions checks that we have set up for activity fair types, which occur in the serializer.

You can replicate all of the checks here, but that will result in some code duplication which is generally not a good idea. A clean solution would be to let the serializer parse the data for you, and use the parsed data without using the serializer to save. See [this post](https://stackoverflow.com/a/46881848) for an example on how to do so.

ezwang: This will also fix the error issue mentioned in the comment above, as long as you return a good message for when the serializer input ins invalid.

backend/clubs/views.py
ezwang: This is also ~3-5 lines of code, so it wouldn't be too hard to add to this PR.

backend/clubs/admin.py
```diff
@@ -370,6 +371,7 @@ class QuestionAnswerAdmin(admin.ModelAdmin):
 admin.site.register(MembershipInvite, MembershipInviteAdmin)
 admin.site.register(Profile, ProfileAdmin)
 admin.site.register(QuestionAnswer, QuestionAnswerAdmin)
+admin.site.register(RecurringEvent)
```
ezwang, line 374: You might want to override the `__str__` function in the `RecurringEvent` model, otherwise the admin won't be too useful. The alternative would be to add more columns to the admin using a custom `ModelAdmin` object, which would be cleaner but more work.

I would recommend the string override returning a format like `{event name}: {start date} - {end date} ({occurences} times)`, where the event name is the name for the earliest instance, the start date is the date for the earliest instance, and the end date is the date for the latest instance (ideally, these would be separate columns in the admin).

PR #344, opened 2020-12-04
ICS import 

rohan-gupta8:
- Added class method to add events retrieved from an ICS file
- Added an ICS URL field to store a constant URL for each club
- Added a UUID field for events
- Added a is_ics_field for events
- Added a command script to be run at some regular frequency'

- Defined the following fallback priority for addition of new ICS events: 
-- an event exists in the database with the same uuid (updated)
-- an event exists in the database with the same club, start time, and end time (updated)
-- a new event (created)

[ch2915]


ezwang (changes requested):
Looks great! Could you add a test case for the management script? You can see existing script tests [here](https://github.com/pennlabs/penn-clubs/blob/master/backend/tests/clubs/test_commands.py).

Just create a new class and add a test case method there. This should be very straightforward, except you'll need to mock the HTTP request.

You'll want to `mock` the request object to not actually make the HTTP request but instead return a dummy HTTP response. The `mock` library is built into Python under the unit test library. You can see various posts like [this one](https://stackoverflow.com/a/28507806) for an idea on how to do this.

backend/clubs/management/commands/import_calendar_events.py
```diff
@@ -0,0 +1,9 @@
+from django.core.management.base import BaseCommand
+
+from clubs.models import Club
+
+
+class Command(BaseCommand):
```
ezwang, line 6: @rohan-gupta8 Add a `help` attribute to this command, see any of the other commands for an example. Should be a description of what this script does, will show up when you do `./manage.py import_calendar_events --help`.

backend/clubs/models.py
```diff
@@ -244,6 +246,20 @@ def __str__(self):
     def create_thumbnail(self, request=None):
         return create_thumbnail_helper(self, request, 200)
 
+    def add_ics_events(self):
+        url = self.ics_import_url
+        if url is not None:
```
ezwang, line 251: To be more robust, you can do `if url:` so that an empty string also does nothing instead of throwing an error.

backend/tests/clubs/test_commands.py
```diff
@@ -2,18 +2,71 @@
...
+
+        def text(self):
+            return self.text
+
+    cal = Calendar()
+    event = Ev()
+    event.name = "A test event"
+    event.description = "A test description"
+    event.begin = datetime.datetime(2020, 1, 2, 3, 4, 5)
+    event.end = datetime.datetime(2020, 1, 2, 3, 4, 5) + datetime.timedelta(minutes=60)
```
ezwang, line 35: It would be a good idea to make this test use the current time plus a few days. In the future, we may not want to import events that are too far into the past.

In addition, if you use `timezone.now()`, it will add the timezone information so you do not need to use `tzutc`.

backend/clubs/models.py
```diff
@@ -244,6 +246,20 @@ def __str__(self):
...
+            calendar = Calendar(requests.get(url).text)
+            for event in calendar.events:
+                ev = Event.objects.create(
```
ezwang, line 254: You should handle the case where an event already exists. Every time you import, you will create a new event. Since this script will be run daily, this will lead to duplicate events.

There are several ways to go about this. I think that ICS events have a unique identifier, which you could potentially save and use to avoid duplicate events.

An alternative method would be to treat events with the same start and end time as the same, since there's not many cases where a club would want to hold two duplicate meetings at the same time. You could also add in the `name` field, but this would break in cases where the event is renamed, so I would recommend against this.

backend/tests/clubs/test_commands.py
```diff
@@ -2,18 +2,71 @@
...
 
+from dateutil.tz import tzutc
 from django.contrib.auth import get_user_model
 from django.core import mail
 from django.core.management import call_command
 from django.core.management.base import CommandError
 from django.test import TestCase
 from django.utils import timezone
+from ics import Calendar
+from ics import Event as Ev
```
ezwang, line 15: Ideally make this more descriptive, `CalendarEvent` or `ICSEvent` instead of `Ev`.

backend/clubs/migrations/0059_auto_20201204_1751.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 3.1.3 on 2020-12-04 22:51
```
ezwang, line 1: Remember to rebase and delete/regenerate this migration before merging!

PR #361, opened 2020-12-27
Use a runtime config

ArmaanT:
This PR configures clubs to use a [next runtime config](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) that will allow us to dynamically provide a DOMAIN/SITE_ORIGIN to use.

The idea is that nothing will change in production, but we will be able to run integration tests in CI using docker-compose (integration tests must be run on localhost because of CORS/CSRF issues)

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -18,4 +18,9 @@ module.exports = {
       },
     ]
   },
+  publicRuntimeConfig: {
+    SITE_ORIGIN: process.env.DOMAIN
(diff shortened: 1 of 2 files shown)


ezwang (approved):
LGTM!

davish (approved):
This looks good! My question is, should we make heavier use of this for branding? It's definitely not a priority for us but if we could use the runtime config to allow an alternate `custom-branding.json` then any other school could spin up a clubs instance with their branding amazingly easily

PR #362, opened 2020-12-28
Even more runtime config

ArmaanT:
This PR adds an additional runtime config option that allows us to define the frontend styling. This means we can run both clubs and hub from the same docker images.

### DIFF
--- a/frontend/utils/branding.tsx
+++ b/frontend/utils/branding.tsx
@@ -1,8 +1,10 @@
+import getConfig from 'next/config'
 import { ReactNode } from 'react'
 
 import { ClubEventType } from '../types'
 
-const site = process.env.NEXT_PUBLIC_SITE_NAME || 'clubs'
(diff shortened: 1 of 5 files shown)


frontend/utils/branding.tsx
```diff
@@ -1,8 +1,10 @@
+import getConfig from 'next/config'
 import { ReactNode } from 'react'
 
 import { ClubEventType } from '../types'
 
-const site = process.env.NEXT_PUBLIC_SITE_NAME || 'clubs'
+const { publicRuntimeConfig } = getConfig()
+const site = publicRuntimeConfig.NEXT_PUBLIC_SITE_NAME || 'clubs'
```
davish, line 7: I feel like this should be unconditionally set to NEXT_PUBLIC_SITE_NAME, and we have the || clubs inside of the runtime config itself.

ezwang (approved):
LGTM!

PR #369, opened 2021-01-02
Barebone Club Application

xiayufeilucy:
Add club application with the following fields:

- Application start time (when does this application open?)

- Application end time (when does this application close?)

- Results release time (when are results released for this application?)

- External url (what is the URL for this application?)

[ch2969]


backend/clubs/models.py
```diff
@@ -1194,6 +1194,25 @@ def __str__(self):
...
+    Represents an application to a club.
+    """
+
+    club = models.ForeignKey(Club, on_delete=models.CASCADE)
+    application_start_time = models.DateTimeField()
+    application_end_time = models.DateTimeField()
+    result_release_time = models.DateTimeField()
+    application_url = models.URLField()
+
+    created_at = models.DateTimeField(auto_now_add=True)
```
ezwang, line 1208: It would be good to rename `application_url` to `external_url`, but that's just personal preference (since the object is already called application).

We might also want a name text field, for when there are multiple applications for the same semester (different member and officer applications, etc). This should be allowed to be blank, in which case we can generate an appropriate name from the club and dates.

You might want to add `updated_at`, just in case.

backend/clubs/views.py
```diff
@@ -2979,6 +2981,15 @@ def get_serializer_class(self):
         return UserProfileSerializer
 
 
+class ClubApplicationViewSet(viewsets.ModelViewSet):
+    permission_classes = [ProfilePermission | IsSuperuser]
```
ezwang, line 2985: You probably don't want to use `ProfilePermission` in this case. Only officers should be able to create, edit, and delete club applications. A more appropriate permission class would be the one that the `Event` or `Testimonial` view sets use.

backend/clubs/views.py
```diff
@@ -2979,6 +2981,15 @@ def get_serializer_class(self):
...
+    serializer_class = ClubApplicationSerializer
+    http_method_names = ["get", "post", "delete"]
```
ezwang, line 2987: You should add `put` and `patch` here so that club officers can edit the event.

ezwang (changes requested):
LGTM! Nice work! 🎉 

I've left a few comments on the PR. After you resolve these comments, this should be good to merge.

backend/clubs/models.py
```diff
@@ -1194,6 +1194,27 @@ def __str__(self):
         return self.user.username
 
 
+class ClubApplication(models.Model):
+    """
+    Represents an application to a club.
+    """
+
+    club = models.ForeignKey(Club, on_delete=models.CASCADE)
+    name = models.CharField(max_length=255, blank=True)
```
ezwang, line 1203: I would recommend using a `TextField` instead of a `CharField` for unlimited length! The performance impact is negligible and who knows how long people would want to make their titles.

backend/clubs/views.py
```diff
@@ -2979,6 +2981,22 @@ def get_serializer_class(self):
...
 
 
+class ClubApplicationViewSet(viewsets.ModelViewSet):
+    """
+    list: Retrieve a list of applications of the club.
+
+    get: Retrieve the details for a given application.
+    """
+
+    permission_classes = [IsSuperuser]
```
ezwang, line 2991: You don't want to just do `IsSuperuser`, since normal users need to view the applications and club officers need to be able to create, edit, and delete the applications.

Look at `ClubItemPermission` for more details, and how the `TestimonialViewSet` and the `EventViewSet` use it.

backend/clubs/views.py
```diff
@@ -2979,6 +2981,22 @@ def get_serializer_class(self):
...
+    serializer_class = ClubApplicationSerializer
+    http_method_names = ["get", "post", "put", "patch", "delete"]
+    lookup_field = "name"
```
ezwang, line 2994: Having the `lookup_field` be the name won't work well, since it is possible for two applications to have the same name and no unique constraint is being enforced.

I would recommend just removing this field and using the model IDs (built-in) for the individual endpoints. Be sure to return the IDs in the serializer so the client knows what to use.

xiayufeilucy: If the url is {club_code}/clubapplication/name, would it make sense to look up by name?

backend/clubs/serializers.py
```diff
@@ -1819,6 +1820,51 @@ class Meta:
...
 
 
+class ClubApplicationSerializer(serializers.ModelSerializer):
+    club = serializers.SlugRelatedField(queryset=Club.objects.all(), slug_field="code")
+    name = serializers.SerializerMethodField("get_name")
+
+    def get_name(self, obj):
+        if obj.name:
+            return obj.name
+        return "{}-{}".format(obj.club.name, obj.application_start_time.strftime("%b %d, %Y"))
```
ezwang, line 1830: It would be good to format this a little nicer! It'll be displayed directly to the user without any modification from the frontend.

I would recommend something like `f"{obj.club.name} Application"` and leave out the date, as that is a pretty generic default that should work for most cases. We'll render the application start and end times in the frontend anyways and make it look nicer, so no need to duplicate that information.

Also check out `f"{...}"` strings, they're a shortcut for `"{}".format(...)`! They're relatively new (Python 3.6+), but we've upgraded everything enough to use them now.

ezwang (approved):
LGTM! Great work! 😄

PR #370, opened 2021-01-03
Added ip field to ClubVisit model

cphalen:
Added IP address to the ClubVisit model and updated serializer to save IP address. All existing ClubVisit entires are defaulted to have `null` IP.

[ch2980]


backend/clubs/models.py
```diff
@@ -762,6 +762,7 @@ class ClubVisit(models.Model):
 
     person = models.ForeignKey(get_user_model(), on_delete=models.CASCADE)
     club = models.ForeignKey(Club, on_delete=models.CASCADE)
+    ip = models.CharField(max_length=255, blank=True, null=True)
```
ezwang, line 765: Instead of a `CharField`, there is a [GenericIPAddressField](https://docs.djangoproject.com/en/3.1/ref/models/fields/#genericipaddressfield) that you can use.

backend/clubs/serializers.py
```diff
@@ -1409,10 +1409,14 @@ class UserClubVisitSerializer(serializers.ModelSerializer):
 
     person = serializers.HiddenField(default=serializers.CurrentUserDefault())
     club = ClubListSerializer(read_only=True)
+    ip = serializers.SerializerMethodField("get_ip")
+
+    def get_ip(self, obj):
+        return self.context["request"].META["REMOTE_ADDR"]
```
ezwang, line 1415: This won't actually do anything to save the IP into the model, it'll just return the user their own IP in the response. You'll need to save the IP when you save the `ClubVisit` object, not when you fetch it.

ezwang (changes requested):
Left a few comments on the PR. Should be good to merge after you fix them!

backend/clubs/serializers.py
cphalen: Ok great point not sure how I didn't notice this earlier

backend/clubs/models.py
cphalen: Thanks!

backend/clubs/serializers.py
```diff
@@ -1409,10 +1409,25 @@ class UserClubVisitSerializer(serializers.ModelSerializer):
 
     person = serializers.HiddenField(default=serializers.CurrentUserDefault())
     club = ClubListSerializer(read_only=True)
+    ip = serializers.HiddenField(default=None)
+
+    def save(self):
+        if not self.instance:
+            if not self.validated_data.get("ip"):
```
ezwang, line 1416: You can remove this if statement, you don't want to save the IP if the user specifies it (always trust the server, no reason for user to specify it).

cphalen: Ok will do!

backend/clubs/migrations/0068_clubvisit_ip.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.1.4 on 2021-01-02 23:42
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("clubs", "0067_profile_show_profile"),
```
ezwang, line 9: You'll need to delete and regenerate this migration after Lucy merged in her PR, since you've both made database edits.

backend/clubs/serializers.py
```diff
@@ -1409,10 +1409,23 @@ class UserClubVisitSerializer(serializers.ModelSerializer):
...
+        if x_forwarded_for:
+            ip = x_forwarded_for.split(",")[0]
+        else:
+            ip = self.context["request"].META.get("REMOTE_ADDR", None)
+        return ip
 
     class Meta:
         model = ClubVisit
-        fields = ("club", "visit_type", "person")
+        fields = ("club", "visit_type", "person", "ip")
```
ezwang, line 1428: I'm pretty sure you can leave out the `"ip"` and the HiddenField altogether, and it will do the same thing. You'll need to test this, though.

backend/clubs/migrations/0068_clubvisit_ip.py
ezwang: Also, delete the `backend/.swp` file in this PR.

backend/clubs/serializers.py
cphalen: I double checked this and yes it did the same thing! Thanks for pointing that out

ezwang (approved):
LGTM!

PR #378, opened 2021-01-09
Kraken Migration

ArmaanT:
Everything but integration tests are currently working. I'm running into a problem with cypress.


ezwang:
Rebasing this to make sure it doesn't fall too out of date.

PR #383, opened 2021-01-12
Hub changes

ezwang:
- Add links between Penn Clubs and Hub@Penn when there are no search results.
- Update social media `og:image` metadata tags.
- Replace the feedback button with a scroll up button on Hub@Penn.
- Fix animation for feedback button on Penn Clubs.
- Hide the information box if there is no information.
- Fix bug with email template rendering with no arguments.

### DIFF
--- a/backend/clubs/management/commands/send_emails.py
+++ b/backend/clubs/management/commands/send_emails.py
@@ -3,6 +3,8 @@ import csv
 import datetime
 import os
 import re
+import requests
+import tempfile
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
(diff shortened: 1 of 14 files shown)


backend/clubs/management/commands/send_emails.py
```diff
@@ -185,10 +198,7 @@ def handle(self, *args, **kwargs):
                     contacts.append("Staff member")
 
                 # Format names in comma separated form
-                recipient_string = ""
-                for contact in contacts:
-                    recipient_string += contact + ", "
-                recipient_string = recipient_string[:-2]
+                recipient_string = ", ".join(contacts)
```
ezwang, line 201: @cphalen FYI

backend/clubs/views.py
```diff
@@ -3599,23 +3648,22 @@ def post(self, request):
...
 
 
 def get_initial_context_from_types(types):
     """
     Generate a sample context given the specified types.
     """
+    # handle edge case where user explicitly says there are no types
+    if isinstance(types, str) and types == "None":
+        return {}
+
```
ezwang, line 3666: @cphalen Fixed the edge case with the email previews here.

PR #388, opened 2021-01-16
Zoom meeting visit model and updated webhook

cphalen:
[ch3028]


backend/clubs/views.py
```diff
@@ -2878,33 +2878,64 @@ def get(self, request):
...
-            return Response({"count": 0})
-        key = f"zoom:meeting:live:{id}"
-        ans = conn.get(key)
-        if ans is None:
-            ans = 0
-        else:
-            ans = int(ans)
-        return Response({"count": ans})
+        meeting_id = request.query_params.get("event")
+        return Response({"count": ZoomMeetingVisit.objects.filter(meeting_id=meeting_id).count()})
```
cphalen, line 2882: Right now this is the same functionality as before, but can be pretty easily expanded upon so that we can list users within the meeting if that's something we decide will be helpful to prospective applicants

backend/clubs/views.py
```diff
@@ -2878,33 +2878,64 @@ def get(self, request):
...
-                conn.decr(key)
-            conn.expire(key, datetime.timedelta(hours=8))
+        if action == "meeting.participant_joined":
+            email = (
+                request.data.get("payload", {})
+                .get("object", {})
+                .get("participant", {})
+                .get("email", None)
+            )
+            person = get_user_model().objects.filter(email=email)
```
cphalen, line 2893: I'm not positive that this is the best way to get the user (filtering the user model) but I couldn't think of any other way to get the user from our DB using the information provided by the webhook API

backend/clubs/views.py
```diff
@@ -2878,33 +2878,64 @@ def get(self, request):
...
+
+            event_name = request.data.get("payload", {}).get("object", {}).get("topic", None)
+            event = Event.objects.filter(name=event_name).order_by("-start_time")[0]
```
cphalen, line 2896: My logic here is that we can get the event by name and then the most recent one will be the correct event (because the participant is currently in the event) but this might not work with recurring events that are planned into the future (because order by DESC means those will come first). There has got to be a better way here so lmk if anyone has ideas

backend/clubs/views.py
ezwang: This won't actually work, as the Penn emails are `pennkey@seas.upenn.edu`, etc while the Zoom ones are `pennkey@upenn.edu`. You instead want to parse the username from the email and do matching on the username.

backend/clubs/views.py
ezwang: This may be too fragile, the topic isn't guaranteed to be the event name and the topic can always be changed by the user.

You should instead match on Zoom meeting IDs, which can't be changed as easily. See if you can match based on the ID in the Zoom URL.

backend/clubs/views.py
```diff
@@ -2878,33 +2878,64 @@ def get(self, request):
...
+            )
+            leave_time = (
+                request.data.get("payload", {})
+                .get("object", {})
+                .get("participant", {})
+                .get("leave_time", None)
+            )
+            ZoomMeetingVisit.objects.filter(meeting_id=meeting_id).filter(
+                participant_id=participant_id
+            ).update(leave_time=leave_time)
```
ezwang, line 2938: This won't work if the user joins and leaves the meeting multiple times, you'll set all of the leave times to be the latest value.

There is also another edge case where you have the user join the meeting twice from two different devices. Although this is a rare case, it would be nice to handle it.

backend/clubs/models.py
```diff
@@ -846,6 +846,25 @@ def __str__(self):
...
 
 
+class ZoomMeetingVisit(models.Model):
+    """
+    Stores information on a user's attendance to a Zoom meeting
+    """
+
+    person = models.ForeignKey(get_user_model(), on_delete=models.CASCADE)
+    club = models.ForeignKey(Club, on_delete=models.CASCADE)
+    event = models.ManyToManyField("Event", null=True)
```
ezwang, line 856: This should not be a ManyToMany. We want to record the meeting times for each meeting visit. We therefore need one `ZoomMeetingVisit` object for each user visit. An event can have multiple visits, but each visit should only be associated with one event, so you want a foreign key instead.

Also, club is redundant if you have a ForeignKey since you can look it up through event.

backend/clubs/views.py
```diff
@@ -2878,33 +2878,64 @@ def get(self, request):
...
+                .get("participant", {})
+                .get("join_time", None)
+            )
+
+            if event:
+                club = event.club
+                ZoomMeetingVisit.objects.create(
+                    person=person,
+                    club=club,
+                    event=event,
```
ezwang, line 2917: You originally defined event as a ManyToMany field, but you use it like a foreign key here, which would throw an error.

I would recommend writing a test case to ensure that the behavior is what you want. You have the sample input format in the clubhouse post (replace any sensitive user information with dummy values).

backend/clubs/views.py
ezwang: Also handle the edge case where there is no username match. This could happen if someone gets the link from their friend, never logged into Penn Clubs before, and joins the meeting.

In this case, you could set this field to null and make the person foreign key nullable.

backend/clubs/views.py
ezwang: You can also combine a `.filter(a).filter(b)` into a `.filter(a, b)`. When you do comma separated, it is `and` by default. If you want to do `or`, it would be `.filter(Q(a) | Q(b))`.

backend/clubs/models.py
ezwang: As a general note, there will be no effect of `null=True` on a ManyToMany field. They can be empty anyways, and in the database representation a ManyToMany is a new SQL table, and there is no concept of a null row (the equivalent is an empty table when filtering on one of the sides).

backend/clubs/views.py
```diff
@@ -2878,33 +2878,74 @@ def get(self, request):
...
+                .get("participant", {})
+                .get("email", None)
+            )
+
+            person = None
+            if email:
+                username = email.split("@")[0]
+                res = get_user_model().objects.filter(username=username)
+                if res:
+                    person = res.first()
```
ezwang, line 2899: This is redundant, `.first()` will return `None` if there are no matching objects.

backend/clubs/views.py
```diff
@@ -2878,33 +2878,74 @@ def get(self, request):
...
+
+            meeting_id = request.data.get("payload", {}).get("object", {}).get("id", None)
+            event = Event.objects.filter(url__contains=meeting_id).first()
```
ezwang, line 2902: I'd recommend using some regex to make this more precise (see `__regex` for details)!

backend/clubs/views.py
cphalen: Awesome thanks for pointing this out!

backend/clubs/views.py
cphalen: Ok great tip I will check that out!

PR #389, opened 2021-01-16
Verify the Authorization header in the Zoom Webhook

xiayufeilucy:
Verify whether incoming post request from Zoom is valid

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -2895,7 +2895,7 @@ class MeetingZoomWebhookAPIView(APIView):
         return Response({"count": ans})
 
     def post(self, request):
-        if "HTTP_AUTHORIZATION" in request.META:
+        if settings.ZOOM_VERIFICATION_TOKEN is not None:
             authorization = request.META["HTTP_AUTHORIZATION"]
             if authorization != settings.ZOOM_VERIFICATION_TOKEN:
                 return Response(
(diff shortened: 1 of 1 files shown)


backend/clubs/views.py
```diff
@@ -2895,6 +2895,14 @@ def get(self, request):
         return Response({"count": ans})
 
     def post(self, request):
+        if "HTTP_AUTHORIZATION" in request.META:
```
ezwang, line 2898: This is a security issue. You can bypass the security check by just not including the authorization header.

ezwang (changes requested):
LGTM! Fix the one comment and we should be good to go.

backend/clubs/views.py
```diff
@@ -2895,6 +2895,14 @@ def get(self, request):
         return Response({"count": ans})
 
     def post(self, request):
+        if settings.ZOOM_VERIFICATION_TOKEN is not None:
```
ezwang, line 2898: You can make this check a bit more relaxed by removing the `is not None`, which also fails on empty string.

backend/clubs/views.py
```diff
@@ -2895,6 +2895,14 @@ def get(self, request):
...
+            authorization = request.META["HTTP_AUTHORIZATION"]
```
ezwang, line 2899: You can use a `request.META.get()` to return 403 instead of error if they do not specify the header. (`[]` errors, `.get()` returns None if not exists)

ezwang (approved):
LGTM!

ezwang (approved):
LGTM!

PR #390, opened 2021-01-18
FAQ updates for hub/clubs

snie11:
- simple FAQ updates with links to Hub/Clubs redirecting users to the other resource

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,4 +52,3 @@ frontend/coverage
 
 # Logs
 yarn-error.log
-frontend/yarn.lock
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ frontend/coverage
 
 # Logs
 yarn-error.log
+frontend/yarn.lock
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -52,3 +52,4 @@ frontend/coverage
 
 # Logs
 yarn-error.log
+frontend/yarn.lock
```
ezwang, line 55: You don't want to ignore this file. This ensures that the exact versions of the packages you intend are installed in production.

frontend/constants/routes.ts
```diff
@@ -41,3 +41,6 @@ export const REPORT_EDIT_ROUTE = (slug?: number): string =>
 
 export const PROFILE_ROUTE = (slug?: string): string =>
   slug ? `/user/${slug}` : '/user/[user]'
+
+export const CLUBS_HOME = 'http://pennclubs.com/'
+export const HUB_HOME = 'http://hub.provost.upenn.edu/'
```
ezwang, line 46: You should use the `https` versions of the sites instead of the `http` version.

Also grep the codebase for existing constants that don't use this new constants, and update them.

ezwang (changes requested):
Should be good after the comments are fixed!

PR #391, opened 2021-01-18
Retrieve User Information In Club Fairs

rohan-gupta8:
[ch3031]

Given a fair this endpoint returns, grouped by event id,

- How many participants there are in the meeting.
- The names/usernames of participants in the meeting who are officers or owners.
- How many participants already attended this meeting or are currently attending.


backend/clubs/views.py
```diff
@@ -506,6 +517,37 @@ def get_serializer_class(self):
...
+                event.id: {
+                    "num_participants": len(
+                        ZoomMeetingVisit.objects.filter(event=event, leave_time=None)
+                    ),
+                    "officers": get_officers(event),
+                    "already_attended": len(ZoomMeetingVisit.objects.filter(event=event)),
+                }
+            }
+            for event in events
+        ]
```
ezwang, line 547: You want to return a dictionary, not a list of dictionaries. Your current output generates something like this:

```
[{3: ...}, {4: ...}, {5: ...}]
```

The desired result is this:

```
{3: ..., 4: ..., 5: ...}
```

You can do list comprehensions with dictionaries, for example you can try running `{a: a for a in range(3)}` which will produce `{0: 0, 1: 1, 2: 2}`.

rohan-gupta8: woops sorry, I had it as something like "event" : event.id  : {num_part, officers, already..} before but I removed the "event" key because it was redundant

backend/clubs/views.py
```diff
@@ -481,8 +481,19 @@ def filter_queryset(self, request, queryset, view):
         return new_queryset
 
 
+def get_officers(event):
+    officers = []
+    for visit in ZoomMeetingVisit.objects.filter(event=event, leave_time=None):
+        m = Membership.objects.get(club=event.club, person=visit.person)
+        if m and m.role <= Membership.ROLE_OFFICER:
+            officers.append(visit.person.username)
```
ezwang, line 489: This is extremely inefficient and will result in `(# of events) * (# of visits) * 2` SQL queries in the worst case. Given ~600 events and ~10 visits per event, this is 1,200 SQL queries per request.

You can get this down to 1 SQL query with a carefully crafted query.

backend/clubs/views.py
```diff
@@ -506,6 +517,37 @@ def get_serializer_class(self):
             return WritableClubFairSerializer
         return ClubFairSerializer
 
+    @action(detail=True, methods=["get"])
+    def get_events_user_info(self):
```
ezwang, line 521: This will result in the endpoint being called `/clubfairs/{clubid}/get_events_user_info/?format=json`, which is kind of verbose. I would name it something like `live`.

backend/clubs/views.py
ezwang: Ah, typoed here, it would actually be 12,000 SQL queries.

backend/clubs/views.py
```diff
@@ -506,6 +517,37 @@ def get_serializer_class(self):
...
+        )
+
+        response = [
+            {
+                event.id: {
+                    "num_participants": len(
+                        ZoomMeetingVisit.objects.filter(event=event, leave_time=None)
+                    ),
+                    "officers": get_officers(event),
+                    "already_attended": len(ZoomMeetingVisit.objects.filter(event=event)),
```
ezwang, line 543: These two queries are too inefficient, as this will result in 1,200 SQL queries per request (given a ~600 event fair).

ezwang: Also this double counts users, as a user could join three times and that would count as 3 separate entries.

backend/clubs/views.py
rohan-gupta8: Ok let me probe

PR #400, opened 2021-01-29
Table

alnasir7:
- New Table component with search, filtering and sorting
Screen Shot 2021-01-29 at 5 20 37 PM

### DIFF
--- a/frontend/components/ClubEditPage/MembersCard.tsx
+++ b/frontend/components/ClubEditPage/MembersCard.tsx
@@ -85,6 +85,24 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
             name: 'email',
           },
         ]}
+        filterOptions={[
+          {
+            label: 'role',
+            options: ['Owner', 'Member'],
+            func: (selection, object) => {
+              switch (selection) {
+                case 'Owner':
+                  if (object.role === 0) return true
+                  else return false
+                  break
+                case 'Member':
+                  if (object.role > 0) return true
+                  else return false
+                  break
+              }
+            },
+          },
+        ]}
         currentTitle={(obj) =>
           obj != null ? `${obj.name} (${obj.email})` : 'Kicked Member'
         }
--- a/frontend/components/ClubEditPage/MembersCard.tsx
+++ b/frontend/components/ClubEditPage/MembersCard.tsx
@@ -88,7 +88,11 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
         filterOptions={[
           {
             label: 'role',
-            options: [{key: 0, label:"Owner"}, {key:20, label:'Member'}, {key:10, label :'Officer'}],
+            options: [
+              { key: 0, label: 'Owner' },
+              { key: 20, label: 'Member' },
+              { key: 10, label: 'Officer' },
+            ],
             filterFunction: (selection, object) => object.role === selection,
           },
         ]}
--- a/frontend/components/ClubEditPage/MembersCard.tsx
+++ b/frontend/components/ClubEditPage/MembersCard.tsx
@@ -99,6 +99,7 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
                   if (object.role > 0) return true
                   else return false
                   break
+                default: return false
               }
             },
           },
--- a/frontend/components/ClubEditPage/MembersCard.tsx
+++ b/frontend/components/ClubEditPage/MembersCard.tsx
@@ -88,15 +88,19 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
         filterOptions={[
           {
             label: 'role',
-            options: ['Owner', 'Member'],
+            options: ['Owner', 'Member', 'Officer'],
             func: (selection, object) => {
               switch (selection) {
                 case 'Owner':
                   if (object.role === 0) return true
                   else return false
                   break
+                case 'Officer':
+                  if (object.role === 10) return true
+                  else return false
+                  break
                 case 'Member':
-                  if (object.role > 0) return true
+                  if (object.role > 10) return true
                   else return false
                   break
                 default:
--- a/frontend/components/ClubEditPage/MembersCard.tsx
+++ b/frontend/components/ClubEditPage/MembersCard.tsx
@@ -99,7 +99,8 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
                   if (object.role > 0) return true
                   else return false
                   break
-                default: return false
+                default:
+                  return false
               }
             },
           },
--- a/frontend/components/ClubEditPage/MembersCard.tsx
+++ b/frontend/components/ClubEditPage/MembersCard.tsx
@@ -89,7 +89,7 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
           {
             label: 'role',
             options: [{key: 0, label:"Owner"}, {key:20, label:'Member'}, {key:10, label :'Officer'}],
-            func: (selection, object) => object.role === selection,
+            filterFunction: (selection, object) => object.role === selection,
           },
         ]}
         currentTitle={(obj) =>
--- a/frontend/components/ClubEditPage/MembersCard.tsx
+++ b/frontend/components/ClubEditPage/MembersCard.tsx
@@ -88,25 +88,8 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
         filterOptions={[
           {
             label: 'role',
-            options: ['Owner', 'Member', 'Officer'],
-            func: (selection, object) => {
-              switch (selection) {
-                case 'Owner':
-                  if (object.role === 0) return true
-                  else return false
-                  break
-                case 'Officer':
-                  if (object.role === 10) return true
-                  else return false
-                  break
-                case 'Member':
-                  if (object.role > 10) return true
-                  else return false
-                  break
-                default:
-                  return false
-              }
-            },
+            options: [{key: 0, label:"Owner"}, {key:20, label:'Member'}, {key:10, label :'Officer'}],
+            func: (selection, object) => object.role === selection,
           },
         ]}
         currentTitle={(obj) =>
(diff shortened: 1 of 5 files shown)


frontend/components/ModelForm.tsx
```diff
@@ -145,92 +146,30 @@ export const ModelTable = ({
...
+      
+    },
+    {
+      label: 'name',
+      options: ['Thomas Jefferson', "James Madison"],
+      
+    },
+
+    
+  ]
```
ezwang, line 162: This seems to be hardcoded, are there still changes we need to make before merging?

frontend/components/ModelForm.tsx
```diff
@@ -145,92 +146,30 @@ export const ModelTable = ({
...
-            {objects.length} total entries, {pageSize} entries per page
-          </span>
-        </div>
-      )}
+      <Table
+        data={objects}
+        columns={tableFields}
+        searchableColumns={['name']}
+        filterOptions = {filterOptions}
+        actions = {[{name: "edit", classes:"button is-primary is-small", Icon:"edit", clickFunction:(target) => onEdit(target) } ,
```
ezwang, line 170: It might actually be better to pass in a function that returns react elements than this JSON object, because you get more functionality. It isn't too much more additional work, and you get that pure flexibility.

You would want to keep your casing consistent if you used this approach (`Icon` -> `icon`).

ezwang: In fact, it would be better to make this a column and then add the ability to custom render columns. That seems like the cleanest solution.

This way you could have multiple action columns if you wanted, and you would also have the ability to do custom formatting like we do now:
image

We're aiming for a table solution that is flexible enough to replace every table on Penn Clubs! This will make it easier to make/maintain tables going forward.

frontend/components/ModelForm.tsx
alnasir7: Yeah, I will change this. I used this to test how filtering works.

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,407 @@
...
+ 
+type tableProps = {
+ columns: Row[],
+ data: Row[],
+ searchableColumns: string[]
+ filterOptions: filterOption [],
+ actions?: Action[]
+}
+ 
+const Styles = styled.div`
```
ezwang, line 71: I love how you styled this table by hand, but it might be ideal to use the existing [Bulma table styles](https://bulma.io/documentation/elements/table/) to match the look of the rest of the site.

Reusing existing library components instead of making your own can save a lot of time going forward, as you have less code to maintain and easier to modify styling.

ezwang: In this case, you wouldn't want to use CSS rules to mimic the styles of existing buttons, but instead use the appropriate `className` to take advantage of Bulma styles.

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,407 @@
...
+ searchableColumns,
+ filterOptions,
+ actions,
+}: tableProps): ReactElement => {
+ const [searchQuery, setSearchQuery] = useState<string>('')
+ const [tableData, setTableData] = useState<Row[]>([])
+ const [selectedFilter, setSelectedFilter] = useState<selectFilter[]>([])
+ useEffect(() => {
+   const searchedData = data.filter((item) => {
+     if (!searchQuery || searchQuery.length < 0) return true
```
ezwang, line 156: `searchQuery.length` will never be less than 0. Taking out useless statements can help towards keeping code maintainable!

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,407 @@
...
+     const searschString = searchQuery
+     if (!searschString || searschString.length < 3) {
```
ezwang, line 158: Fix any typos you find! If you use an intelligent editor like VSCode, you can easily find and replace all instances of a variable name.

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,407 @@
...
+           .startsWith(searschString.toLowerCase())
+       )
+         valid = true
+     }
+     return valid
+   })
+   const filteredData = searchedData.filter((item) => {
+     if (selectedFilter && selectedFilter.length > 0) {
+       let valid = true;
+       selectedFilter.forEach(filter => {
```
ezwang, line 181: You can replace this `forEach` loop with an [every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) statement.

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,407 @@
...
+     filterTypes,
+   },
+   useFilters,
+   useGlobalFilter,
+   useSortBy,
+   usePagination,
+ )
+ 
+ const handleSearchChange = (e) => {
+   if (e.target.value.length >= 0) {
```
ezwang, line 243: Is there a case where this could be less than 0?

frontend/components/ModelForm.tsx
alnasir7: Yes, I will fix these before merging, just to test.

ezwang:
Rebase to the latest master branch to fix the continuous integration!

frontend/components/ClubEditPage/MembersCard.tsx
```diff
@@ -85,6 +85,26 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
...
+            label: 'role',
+            options: ['Owner', 'Member'],
+            func: (selection, object) => {
+              switch (selection) {
+                case 'Owner':
+                  if (object.role === 0) return true
+                  else return false
+                  break
+                case 'Member':
+                  if (object.role > 0) return true
```
ezwang, line 99: There is an officer case here that you're missing.

frontend/components/ModelForm.tsx
```diff
@@ -60,6 +67,7 @@ type ModelFormProps = {
   empty?: ReactElement | string
   fields: any
   tableFields?: TableField[]
+  filterOptions?: filterOption[]
```
ezwang, line 70: This should be title case: `FilterOption` instead of `filterOption`, just a code style guideline.

ezwang:
@alnasir7 It looks like there's still a few bugs with the title/permissions:

image

Also capitalize "Search" in "search 4 entries".

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,362 @@
...
+                      </td>
+                    )
+                  })}
+                </tr>
+              )
+            })}
+          </tbody>
+        </table>
+      ) : (
+        <h1>No matches where found. Please change your filters.</h1>
```
ezwang, line 309: This is a typo.

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,362 @@
...
+                    styles={styles}
+                    components={components}
+                    onChange={(value) =>
+                      handleFilterChange({
+                        value: value ? value.label : null,
+                        label: filterOption.label,
+                      })
+                    }
+                    isClearable={true}
+                    placeholder={`Filter By ${filterOption.label}`}
```
ezwang, line 254: If you call `titleize` on this, it might look a bit better.

ezwang:
Fix the formatting for the column names (capitalization). You should allow the user to pass in a custom label, and if they don't then use the `titleize` version of the attribute name.

image

frontend/components/ClubEditPage/MembersCard.tsx
```diff
@@ -85,6 +85,30 @@ export default function MembersCard({ club }: MembersCardProps): ReactElement {
             name: 'email',
           },
         ]}
+        filterOptions={[
+          {
+            label: 'role',
+            options: ['Owner', 'Member', 'Officer'],
```
ezwang, line 91: It would be better if you made `options` a list of dictionaries with the keys `id` and `label`. This would make `func` cleaner. Also rename `func` to be more descriptive.

Think of a case where you would render react for the option label. Would you be able to easily compare which option is the correct one?

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,362 @@
...
+  const [tableData, setTableData] = useState<Row[]>([])
+  const [selectedFilter, setSelectedFilter] = useState<any>({})
+  useEffect(() => {
+    const searchedData = data.filter((item) => {
+      if (!searchQuery) return true
+      if (!searchQuery || searchQuery.length < 3) {
+        return true
+      }
+      let valid = false
+      for (const searchId in searchableColumns) {
```
ezwang, line 109: This could use the [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) method.

frontend/components/common/Table.tsx
```diff
@@ -0,0 +1,362 @@
...
+}: tableProps): ReactElement => {
+  const [searchQuery, setSearchQuery] = useState<string>('')
+  const [tableData, setTableData] = useState<Row[]>([])
+  const [selectedFilter, setSelectedFilter] = useState<any>({})
+  useEffect(() => {
+    const searchedData = data.filter((item) => {
+      if (!searchQuery) return true
+      if (!searchQuery || searchQuery.length < 3) {
+        return true
+      }
```
ezwang, line 107: These two statements are duplicates.

esinx:
Just noticed that one of the superuser test fails, added a null/undefined-safe protection in `ModelForm.tsx`
```js
      return {
        ...column,
        render: (id) => {
          const obj = objects?.[id]
          const value = obj?.[column.name]
          return obj && value ? renderFunction(value, obj) : 'N/A'
        },
      }
```

frontend/components/ModelForm.tsx
```diff
@@ -46,6 +46,19 @@ type TableField = {
...
+
+type Option = {
+  label: string
+  key: any
+}
+
+type FilterOption = {
+  options: Option[]
+  label: string
+  filterFunction: (a, b) => boolean
```
esinx, line 61: Might be a good idea to add type annotation/more concrete argument description here, but this is fine too!

esinx (approved):
LGTM! Might consider adding a few type annotations for future uses, but that's for later.

PR #401, opened 2021-01-31
Search Logging

cphalen:


backend/clubs/models.py
```diff
@@ -965,6 +965,15 @@ def __str__(self):
         )
 
 
+class Search(models.Model):
+    person = models.ForeignKey(get_user_model(), on_delete=models.CASCADE)
```
ezwang, line 969: This has to be nullable, what if the user is not logged in?

backend/clubs/models.py
```diff
@@ -965,6 +965,15 @@ def __str__(self):
...
+    query = models.CharField(max_length=255)
```
ezwang, line 970: You will receive an error if the user passes in something greater than 255 characters. Either switch to a `TextField`, or truncate the input if it is over 255 characters.

backend/clubs/models.py
```diff
@@ -965,6 +965,15 @@ def __str__(self):
         )
 
 
+class Search(models.Model):
```
ezwang, line 968: You might want to rename this to something more descriptive, like `SearchQuery`.

backend/clubs/serializers.py
```diff
@@ -1536,6 +1537,19 @@ class Meta(UserClubVisitSerializer.Meta):
         pass
 
 
+class SearchSerializer(serializers.ModelSerializer):
+    """
+    Used by users to enter search result into backend.
+    """
+
+    person = serializers.HiddenField(default=serializers.CurrentUserDefault())
+    query = serializers.CharField(allow_null=True)
```
ezwang, line 1546: Why would we want to allow null here? Blank is not the same as null.

backend/clubs/views.py
```diff
@@ -2691,6 +2692,20 @@ def get_serializer_class(self):
...
+    list: Return a list of search queries
+
+    create: Add a new search query
+    """
+
+    permission_classes = [IsAuthenticated]
+    http_method_names = ["post"]
+
+    def get_serializer_class(self):
+        return SearchSerializer
```
ezwang, line 2706: We don't want to give a list of every search query to anyone who requests it. People should be able to create search queries, but only superusers should be able to list them.

ezwang: In this case, you document a `list` endpoint, but the user can't actually list stuff. Just delete this comment and you should be fine.

frontend/components/SearchBar.tsx
```diff
@@ -340,6 +341,10 @@ export const SearchBarTextItem = ({
   const inputRef = useRef() as React.MutableRefObject<HTMLInputElement>
 
   useEffect(() => {
+    doApiRequest('/searches/', {
```
ezwang, line 344: You want to put this inside the timeout logic. This will greatly reduce the number of incomplete searches that get sent to the server.

ezwang (changes requested):
See the comments before merging!

backend/clubs/models.py
cphalen: Good point!

backend/clubs/models.py
cphalen: Yeah I was thinking that no one would enter a search that long, but pays to be safe!

backend/clubs/models.py
cphalen: Good call

frontend/components/SearchBar.tsx
cphalen: Ok that's a great point

backend/clubs/urls.py
```diff
@@ -53,6 +54,7 @@
 router.register(r"favorites", FavoriteViewSet, basename="favorites")
 router.register(r"subscriptions", SubscribeViewSet, basename="subscribes")
 router.register(r"clubvisits", ClubVisitViewSet, basename="clubvisits")
+router.register(r"searches", SearchViewSet, basename="searches")
```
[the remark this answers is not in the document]
cphalen: One step ahead of you! It's already in the PR

xiayufeilucy (approved):

backend/clubs/serializers.py
```diff
@@ -1531,6 +1532,19 @@ class Meta(UserClubVisitSerializer.Meta):
         pass
 
 
+class SearchQuerySerializer(serializers.ModelSerializer):
+    """
+    Used by users to enter search result into backend.
+    """
+
+    person = serializers.HiddenField(default=serializers.CurrentUserDefault())
+    query = serializers.CharField()
```
ezwang, line 1541: There is no need to define this if you're not going to customize the default behavior.

backend/clubs/views.py
```diff
@@ -2705,6 +2707,26 @@ def get_serializer_class(self):
...
+    http_method_names = ["get", "post"]
+
+    def get_queryset(self):
+        if self.request.user.is_superuser:
+            return SearchQuery.objects.all()
+        else:
+            return SearchQuery.objects.filter(person=self.request.user)
+
+    def get_serializer_class(self):
+        return SearchQuerySerializer
```
ezwang, line 2727: You can do `serializer_class = ` instead of this, which is recommended if you're just returning the serializer object.

PR #402, opened 2021-02-02
Place reports under admin

esinx:
This PR places `/reports` under `/admin` (now all tests are directed to `/admin/reports`). I also split some of the tab pages into separate components for better readability.

Somewhat important changes include:
* The admin page now shows a "reports" tab
* The reports page(tab) now checks for permission using `apiCheckPermission`
* The admin page now uses slash(/) routing instead of hash(#) routing

As this is a breaking change from end-user perspective, it might be nice to setup a temporary redirection from `/reports` to `/admin/reports`


ezwang (approved):
LGTM! Made some changes to use the `BrowserTabView` I added to the club edit page.

PR #403, opened 2021-02-02
Form fixes

ezwang:
- Fix model form issue with not updated from server provided response values.
- Allow users to show/hide Q&A requests and responses.
- Make the answer field for Q&A rich text.
- Refactor edit page to use query parameters instead of hashes for tabs.
- Modify the embed object functionality to autocorrect iframe embeds.
- Rename snake case variables to camel case.

[ch3065]


esinx (approved):
lgtm! Seems like #402 would need to update to use BrowserTabView after this gets merged.

PR #408, opened 2021-02-05
Cross origin member list

rohan-gupta8:
[ch3063]

Add the following fields on the membership object (between a person and a club):

Member description - A description of this member's role within the club. 

Member image - An image of this member for a particular club, overrides the image set for the user profile if it exists.

Additionally, enable cross-origin information retrieval to allow third party websites to retrieve member data (no editing)

This is enabled at `api/external/members/<code>`, where `code` corresponds to the club code.


backend/clubs/serializers.py
```diff
@@ -600,9 +600,10 @@ def get_email(self, obj):
     def get_image(self, obj):
         if not obj.public:
             return None
-        if not obj.person.profile.image:
-            return None
-        image_url = obj.person.profile.image.url
+        if not obj.image:
+            if not obj.person.profile.image:
+                return None
+        image_url = obj.image.url or obj.person.profile.image.url
```
xiayufeilucy, line 606: You might want to check if object.image exists before calling obj.image.url. Otherwise calling obj.image.url would return a value error :)

rohan-gupta8: Good catch. Done

backend/clubs/models.py
```diff
@@ -126,6 +126,12 @@ def get_event_small_file_name(instance, fname):
     )
 
 
+def get_membership_image_file_name(instance, fname):
+    return os.path.join(
+        "membership", "{}.{}".format(instance.code, fname.rsplit(".", 1)[-1])
```
cphalen, line 131: The instance here is an instance of `Membership`, right? Because `Membership` has no field `code` field I think that this will default to `None` and you could get conflicts if two users upload files with the same filename.

Maybe you can do something like
```
"{}.{}.{}".format(instance.club.code, instance.person.username, fname.rsplit(".", 1)[-1])
```
which will be distinct because there is only one membership for each (person, club) pairing

backend/pennclubs/settings/base.py
```diff
@@ -108,6 +108,8 @@
     {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator"},
 ]
 
+CORS_ALLOW_ALL_ORIGINS = True
+CORS_URLS_REGEX = r"^/api/.*$"
```
cphalen, line 112: If you haven't already, maybe double check with Armaan about allowing CORS to the whole API. He might prefer it to be restricted just to certain endpoints (like the member list).

cphalen (commented):
A few things I noticed but looks about done!

backend/clubs/serializers.py
```diff
@@ -600,9 +600,11 @@ def get_email(self, obj):
     def get_image(self, obj):
         if not obj.public:
             return None
-        if not obj.person.profile.image:
-            return None
-        image_url = obj.person.profile.image.url
+        if not obj.image:
+            if not obj.person.profile.image:
+                return None
```
ezwang, line 605: This could be reduced to a single if statement.

cphalen: Lucy mentioned this earlier I think!

backend/clubs/models.py
rohan-gupta8: Yup totally. I missed that when I copied over from the other similar methods.

backend/clubs/serializers.py
rohan-gupta8: Changed!

backend/pennclubs/settings/base.py
rohan-gupta8: Messaged him

cphalen:
LGTM! Thanks for revising all of the comments we mentioned earlier. Shouldn't have any CORS issues now that only the external viewset is exposed. I'll merge it in tomorrow

backend/clubs/views.py
```diff
@@ -4013,6 +4014,22 @@ def get_queryset(self):
         )
 
 
+class ExternalMemberListViewSet(viewsets.ModelViewSet):
```
ArmaanT, line 4017: This should probably be a [ReadOnlyViewSet](https://www.django-rest-framework.org/api-guide/viewsets/#readonlymodelviewset), right?

PR #411, opened 2021-02-09
Hub fixes

ezwang:
- Remove Graduate and Undergraduate tags, add resources who chose one or the other to a specific degree type.
- Ensure that the user doesn't select every possible value for the school, major, year, and student type fields.


cphalen:
LGTM!

PR #414, opened 2021-02-15
Route Progress Bar

esinx:
Some pages take a bit to load and I thought it might be a good idea to entertain the user with a progress bar while the page is loading new resources.

This uses next.js's router events to simulate "progress" (which really isn't, it's just a pseudo-progress based on start-complete state events) and its real purpose is to just tell the user that nothing is really frozen, it's just stuff happening in the background.

Inspired by YouTube I guess?

penn-clubs-route-loading

If you want a higher resolution preview, here's the original screen recording: https://user-images.githubusercontent.com/62971511/107964328-8177a900-6fec-11eb-85f2-d70fb937c4be.mov

### DIFF
--- a/frontend/components/RouteProgressBar.tsx
+++ b/frontend/components/RouteProgressBar.tsx
@@ -56,10 +56,14 @@ const keyframeForState = (state: RouteProgressState): Keyframes | null => {
   }
 }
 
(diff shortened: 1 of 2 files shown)


ezwang (approved):
This looks amazing!!

Looks like there's a minor typescript issue that should be fixed before merge. `yarn tsc` to check.

PR #415, opened 2021-02-15
H@P weekly fix

xiayufeilucy:
- [x] Remove school filter
Eunsoo's fixes:
-  [x] Fix bug with student types not showing up on the resource creation page
-  [x] Remove target schools


frontend/pages/index.tsx
```diff
@@ -440,11 +440,11 @@ const Splash = (props: SplashProps): ReactElement => {
               ]}
             />
           )}
-          <SearchBarCheckboxItem
+          {SITE_ID !== 'fyh'&&<SearchBarCheckboxItem
```
ezwang, line 443: Use the `isClubFieldShown('target_schools')` function! It was designed for this purpose.

frontend/pages/club/[club]/renew.tsx
```diff
@@ -222,7 +222,11 @@ const RenewPage = (props: RenewPageProps): ReactElement => {
   }
 
   if (SITE_ID === 'fyh') {
-    return <ResourceCreationPage {...props} />
+    return (
+      <ResourceCreationPage
+        {...{ ...props, student_types: props.studentTypes }}
```
Comment (ezwang) [inline frontend/pages/club/[club]/renew.tsx:227]: The outer later of `{...  }` is unnecessary.

Also, I would refactor the resource creation page to accept studentTypes instead of changing this page to pass in `student_types`. Camel case is more inline with Javascript code style.

frontend/components/ClubEditPage/ClubEditCard.tsx
```diff
@@ -493,7 +493,7 @@ export default function ClubEditCard({
           type: 'multiselect',
           placeholder: `Select schools relevant to your ${OBJECT_NAME_SINGULAR}!`,
           choices: schools,
-          hidden: !showTargetFields,
+          hidden: schools.length === 0 || !showTargetFields,
```
ezwang, line 496: Instead of doing this, you can take advantage of how all the other fields are shown or hidden in `branding.tsx`. See how some fields are hidden or shown based on the variables in that file.

This allows all of our differences in behavior to be standardized.

frontend/pages/club/[club]/renew.tsx
Comment (esinx) [reply frontend/pages/club/[club]/renew.tsx]: Just noticed that `ClubForm` from ClubEditPage.tsx also does this. Seems like it takes parameters from `doBulkLookup` which kind of enforces snake case naming. Should we fix this for both components?

Pipfile
```diff
@@ -0,0 +1,8 @@
+[[source]]
```
ezwang, line 1: Remove this file.

frontend/components/ResourceCreationPage.tsx
```diff
@@ -46,7 +46,7 @@ type ResourceCreationPageProps = {
   years: Year[]
   majors: Major[]
   tags: Tag[]
-  studentTypes: StudentType[]
+  student_types: StudentType[]
```
ezwang, line 49: Do we still want to do this with @esinx's refactor? Should rebase on master and fix.

ezwang (changes requested):
Looks mostly good, just 2 possible changes before merging this in.

frontend/components/ResourceCreationPage.tsx
esinx: we can merge this pr first then fix this in the refactor. I think #418 still needs some testing so I would probably rebase from that branch instead.

backend/clubs/serializers.py
```diff
@@ -1136,6 +1136,20 @@ def validate_target_years(self, value):
...
+                    "In order to best optimize our sorting algorithm,\
+                     you need to select only the few \
+                    tags that apply to your resource."
+                    "If you feel that all the tags apply, that's great!\
+                     In that case you would select 'Yes' to the \
+                    question asking if your resource applies to\
+                     all undergraduate, graduate, and professional Penn students."
+                    "Thanks for doing your part to ensure that Hub@Penn\
+                     quickly and efficiently gets resources to our Penn community. "
+                )
```
ezwang, line 1152: It would be better to use the:

```
abc = (
    "This is a sentence! "
    "This is another sentence!"
)
```

This multiline string formation doesn't result in a bunch of whitespace between the text. Although the browser will collapse the text while rendering it, it'll be nice to fix this at the server level.

```python
>>> abc = (
...     "This is a sentence! "
...     "This is another sentence!"
... )
>>> abc
'This is a sentence! This is another sentence!'
>>> abc = (
...     "This is a sentence!\
...      This is another sentence!"
... )
>>> abc
'This is a sentence!     This is another sentence!'

```

backend/clubs/serializers.py
```diff
@@ -1169,6 +1211,20 @@ def validate_student_types(self, value):
...
+                     for one or more of the previous tags."
+                    "In order to best optimize our sorting algorithm,\
+                     you need to select only the few \
+                    tags that apply to your resource."
+                    "If you feel that all the tags apply, that's great!\
+                     In that case you would select 'Yes' to the \
+                    question asking if your resource applies to\
+                     all undergraduate, graduate, and professional Penn students."
+                    "Thanks for doing your part to ensure that Hub@Penn\
+                     quickly and efficiently gets resources to our Penn community. "
```
ezwang, line 1226: This is a ton of code duplication.

Consider saving this string to a variable and using the variable throughout the program?

ezwang: Imagine if they wanted to change this text, it would be easy to miss an entry if you're not aware that there are multiple.

frontend/components/ClubEditPage/ClubEditCard.tsx
```diff
@@ -458,24 +458,62 @@ export default function ClubEditCard({
...
+                      color={BLACK}
+                    />{' '}
+                    <span className="ml-1">
+                      No, my {OBJECT_NAME_SINGULAR} is restricted to certain
+                      student groups.
+                    </span>
+                  </CheckboxLabel>
+                </div>
+                <Text>
+                  If no, Hub@Penn has provided a way for certain student
```
ezwang, line 487: This is used by both Penn Clubs and Hub@Penn.

This text should not appear in Penn Clubs. Test with both sites to be sure they look appropriate in both.

frontend/pages/club/[club]/renew.tsx
```diff
@@ -222,7 +222,9 @@ const RenewPage = (props: RenewPageProps): ReactElement => {
   }
 
   if (SITE_ID === 'fyh') {
-    return <ResourceCreationPage {...props} />
+    return (
+      <ResourceCreationPage {...props} student_types={props.studentTypes} />
```
Comment (ezwang) [inline frontend/pages/club/[club]/renew.tsx:226]: It would be better to modify `ResourceCreationPage` to accept `studentTypes` instead of `student_types` than doing this.

ezwang (changes requested):
Left a few comments to fix before merging!

frontend/components/ClubEditPage/ClubEditCard.tsx
xiayufeilucy: I think this text only show up in H@P because of the conditional statement at the front

frontend/pages/club/[club]/renew.tsx
Comment (esinx) [reply frontend/pages/club/[club]/renew.tsx:226]: So `ResourceCreationPage` now accepts `studentTypes` in camel case, but seems like `create.tsx`'s `doBulkLookup` method was feeding it with `student_types`. So I tried using `[string, string][]` form argument to rename the request data, but seems like it was broken(`doBulkLookup` won't work with `[string, string][]`; specifically, it would fail to parse the json). This branch now contains the fix, so we should probably check if there are any other parts that depend on this functionality.

backend/clubs/serializers.py
```diff
@@ -45,6 +45,28 @@
 from clubs.utils import clean
 
 
+ALL_TAGS_SELECTED_ERROR_MESSAGE = (
+    (
+        "We noticed you selected all of the options"
```
ezwang, line 50: You will be missing spaces in this text. You will need to manually add a space at the end (`of the options"` => `of the options "`).

frontend/components/ClubEditPage/ClubEditCard.tsx
ezwang: It looks good now! I think that `If no` should be `If not`, but refer to their original document for wording.

frontend/utils.tsx
```diff
@@ -176,6 +176,15 @@ export function apiCheckPermission(
   return false
 }
 
+const chooseEndpoint = (input: [string, string] | string) => {
+  if (typeof input === 'string') {
+    return `/${input}/?format=json`
+  } else {
+    if (input[1].startsWith('/')) return input[1]
+    return chooseEndpoint(input[1])
+  }
```
ezwang, line 185: This seems like you can use an `else if` to make the code a bit clearer, and have the return statement at the end. Its good practice to usually put `{ ... }` even on one line statements, just to make the code easier to read and to reduce bugs if you expand the if statement.

PR #416, opened 2021-02-20
Club Archiving

cphalen:
Moved to archiving clubs instead of deleting them. Let me know if I missed any querysets which also need to be filtering out archived clubs!

[ch3103]


ezwang (changes requested):
There are a few edge cases missing. Think events, fairs, etc.

ezwang:
@xiayufeilucy This will be useful for the OSA refactor, all of the points that Campbell finds are points that you will need to update when you have added the new states.

cphalen:
> There are a few edge cases missing. Think events, fairs, etc.

backend/clubs/serializers.py
```diff
@@ -1736,7 +1736,10 @@ def get_clubs(self, obj):
         # hide non public memberships if not superuser
```
ezwang, line 1736: @cphalen I would also hide this from superusers and only allow it on the Django admin console, to prevent confusing OSA.

cphalen: That's an excellent point thanks

backend/clubs/views.py
```diff
@@ -1834,11 +1839,14 @@ def list(self, request, *args, **kwargs):
...
     def perform_destroy(self, instance):
         """
-        Manually delete uploaded asset instances because of PostgreSQL integrity rules.
+        Set archived boolean to be True so that the club appears to have been deleted
         """
-        for asset in instance.asset_set.all():
-            asset.delete()
-        instance.delete()
+
+        club = self.get_object()
```
ezwang, line 1845: @cphalen Have you tested this code? I have a feeling that `get_object` wouldn't exist/be appropriate and you would want to use `instance` instead.

ezwang: Given the test cases passes I think you'd be fine, but I would use `instance` just in case.

cphalen: Ok awesome good catch — thanks Eric!

cphalen:
I checked for queries to models that have `Club` as a `ForeignKey` and found a decent number of other filters that I was missing. All of these models should be good now though:
[x] Favorite
[x] Subscribe
[x] ClubVisit
[x] MembershipRequest
[x] Advisor
[x] Membership
[x] MembershipInvite
[x] Badge
[x] Asset
[x] ClubApplication
[x] QuestionAnswer
[x] ClubFairRegistration

backend/clubs/views.py
```diff
@@ -2539,7 +2538,9 @@ class QuestionAnswerViewSet(viewsets.ModelViewSet):
 
     def get_queryset(self):
         club_code = self.kwargs["club_code"]
-        questions = QuestionAnswer.objects.filter(club__code=club_code)
+        questions = QuestionAnswer.objects.filter(
+            club__code=club_code, club__archived=False
+        )
```
cphalen, line 2543: @ezwang would
```QuestionAnswer.objects.filter(club__code=club_code).select_related("archived").filter(club__archived=False)```
be better here? I tested with DjDT and it looks like this isn't an N+1 issue, but the Django ORM has tricked me before so figured I should double check

ezwang: You are correct, this would not be better here. Django can automatically optimize filter conditions, and adding a filter condition shouldn't ever cause a N+1 issue.

PR #417, opened 2021-02-21
small changes

alnasir7:
- Switched function of arrows
- Default value of "none" for undefined fields
- bug fixes
- added sort icons


frontend/components/common/Table.tsx
```diff
@@ -303,17 +332,17 @@ const Table = ({
...
       {pageOptions.length > 1 && (
-        <div className="is-clearfix" style={{ display: 'flex' }}>
+        <div className="is-clearfix">
           <button
             style={{ marginRight: '0.5rem' }}
-            onClick={() => gotoPage(0)}
+            onClick={() => previousPage()}
             disabled={!canPreviousPage}
           >
             <Icon name="chevrons-left" />
```
ezwang, line 341: This is the correct behavior, only the right buttons have the behavior swapped.

The single arrow buttons should go forward/back one page, while the double arrow buttons should go all the way to the beginning/end.

frontend/components/ModelForm.tsx
```diff
@@ -151,10 +151,14 @@ export const ModelTable = ({
...
         ...column,
-        render: (id, _) => {
-          const obj = objects?.[id]
+        render: (id) => {
+          const obj = objects?.filter((item) => item.id === id)[0]
+            ? objects?.filter((item) => item.id === id)[0]
+            : objects?.[id]
           const value = obj?.[column.name]
-          return obj && value ? renderFunction(value, obj) : 'N/A'
+          return obj && !(value === null) && !(value === undefined)
```
ezwang, line 159: You can simplify this with `value != null`, which checks to make sure that value is neither `null` nor `undefined`.

frontend/components/ModelForm.tsx
```diff
@@ -151,10 +151,14 @@ export const ModelTable = ({
...
+            ? renderFunction(value, obj)
+            : 'N/A'
```
ezwang, line 161: Instead of `N/A`, `None` might be more appropriate for the general case. For example, if we do not have a user's name, this doesn't mean the user doesn't have a name, we just couldn't retrieve it.

ezwang (changes requested):
1. When the table is empty, you render nothing. You should still render the table column headers in this case, and have a td with full colspan that informs the user that the table is empty. This will avoid confusion on the user's part.

image

2. You are not able to search by all of the columns, just one, but you should be able to. For example, in the members table, you should be able to search by both the full name and the email address. You should show all entries that match either field.

3. The icons for the events table for ICS used to be colored. When improving a component, keep as many of the old (useful) features as you can.

image

4. Using grid is suboptimal styling here, as the display becomes weird for certain screen sizes:
image

In the example above, you can see that there is more than enough space for both controls to be on the same line, but since you use grid the controls are forced to wrap. You should use floats instead.

5. Fix all other comments mentioned in this review.

6. Avoid using your own buttons when possible, and use Bulma buttons for style consistency. In this case, you can use `is-small` and something like `is-light` to replicate the button styles you have now.

image

ezwang (changes requested):
You might not want to delete the icon files. The fact that they were there before you made this PR gives a strong indication that some icon is still using those files.

Otherwise looks good to me! Haven't executed the code though, will let you know if anything comes up. Merge after Armaan has completed the AWS migration.

alnasir7:
> You might not want to delete the icon files. The fact that they were there before you made this PR gives a strong indication that some icon is still using those files.
> 
> Otherwise looks good to me! Haven't executed the code though, will let you know if anything comes up. Merge after Armaan has completed the AWS migration.



> You might not want to delete the icon files. The fact that they were there before you made this PR gives a strong indication that some icon is still using those files.
> 
> Otherwise looks good to me! Haven't executed the code though, will let you know if anything comes up. Merge after Armaan has completed the AWS migration.

I added the icons on the previous PR. They were not used before. In this PR I changed to something else so I removed them

frontend/components/ClubEditPage/EventsCard.tsx
```diff
@@ -292,7 +292,22 @@ const eventTableFields = [
   {
     name: 'is_ics_event',
     label: 'ICS',
-    converter: (a: boolean): ReactElement => <Icon name={a ? 'check' : 'x'} />,
+    converter: (a: boolean): ReactElement => (
+      <Icon
+        className={`has-text-${a ? 'has-text-success' : 'has-text-danger'}`}
```
esinx, line 297: Not sure if a class name with `has-text-has-text-success` or `has-text-has-text-danger` exists. Could you check if this is a typo?

frontend/components/ModelForm.tsx
```diff
@@ -145,29 +148,30 @@ export const ModelTable = ({
...
     if (column.converter) {
       const renderFunction = column.converter
       return {
         ...column,
-        render: (id, _) => {
-          const obj = objects?.[id]
+        render: (id) => {
+          const obj = objects?.filter((item) => item.id === id)[0]
+            ? objects?.filter((item) => item.id === id)[0]
+            : objects?.[id]
```
esinx, line 159: This expression ​can be simplified:
```javascript
const obj = objects?.filter(item =? item.id === id)[0] || objects?.[id]
```

frontend/components/ModelForm.tsx
```diff
@@ -145,29 +148,30 @@ export const ModelTable = ({
...
           const value = obj?.[column.name]
-          return obj && value ? renderFunction(value, obj) : 'N/A'
+          return obj && !(value === null) ? renderFunction(value, obj) : 'None'
```
esinx, line 161: This expression can be simplified:
```javascript
return (obj && value !== null) ? renderFunction(value, obj) : 'None'
```

frontend/components/common/Table.tsx
```diff
@@ -105,16 +109,19 @@ const Table = ({
...
-        const strings = item[searchId].split(' ')
-        return strings.some((string) =>
-          string.toLowerCase().startsWith(searchQuery.toLowerCase()),
-        )
+        if (typeof item[searchId] === 'string') {
+          const strings = item[searchId].split(' ')
+          return strings.some((string) =>
+            string.toLowerCase().startsWith(searchQuery.toLowerCase()),
+          )
+        } else return false
```
esinx, line 124: Suggestion: place this without `else` or add brackets for better readability

frontend/components/common/Table.tsx
```diff
@@ -217,106 +233,121 @@ const Table = ({
...
                 {headerGroup.headers.map((column) => (
                   <th {...column.getHeaderProps(column.getSortByToggleProps())}>
-                    {titleize(column.render('Header'))}
-                    <span style={{ marginLeft: '1rem' }}>
-                      {column.isSorted ? (
-                        column.isSortedDesc ? (
-                          <Icon name="chevron-down" />
+                    <div onClick={() => handleColumnsSort(column.Header)}>
+                      {titleize(column.render('Header'))}
+                      <span style={{ marginLeft: '1rem' }}>
```
esinx, line 279: It's hard to see which icons are being rendered in which conditions, so consider extracting this into a function that returns the icon name!
```javascript
const iconNameForSortedState = (column): string => {
    if(column.isSorted){
        if(column.isSortedDesc) {
            return "triangle-down"
        }else{
            return "triangle-up"
        }
    }
    return "group-arrows"
}
```

frontend/components/common/Table.tsx
```diff
@@ -217,106 +233,121 @@ const Table = ({
...
+                            column.render(row.original.id) ? (
+                              column.render(row.original.id)
+                            ) : (
+                              <GreyText>None</GreyText>
+                            )
+                          ) : row.original[column.name] ? (
+                            row.original[column.name]
+                          ) : (
+                            <GreyText>None</GreyText>
+                          )}
```
esinx, line 317: Can be simplified:
```javascript
{
    column.render ?
    (column.render(row.original.id) || (<GreyText>None</GreyText>))
    : (row.original[column.name] || (<GreyText>None</GreyText>))
}
```

frontend/components/common/Table.tsx
```diff
@@ -332,21 +363,25 @@ const Table = ({
...
             disabled={!canNextPage}
           >
-            <Icon name="chevron-right" />
+            {'>'}
           </button>
           <button
             style={{ marginRight: '0.5rem' }}
-            onClick={() => nextPage()}
+            className="is-light is-small"
+            onClick={() => gotoPage(pageCount - 1)}
```
esinx, line 375: Can `pageCount` ever be 0?

esinx (commented):
Some readability fixes can be really useful, and check for edge cases where some states are not within the expected range!

frontend/components/common/Table.tsx
alnasir7: This might still not be easy to do, in one of the cases I will be returning an empty string and not an icon.

frontend/components/common/Table.tsx
alnasir7: I don't think. CanPrevious is automatically set to disabled when we are at page 1, so I don't see how that might happen.

rohangpta:
@alnasir7 Do you mind walking me through this PR at some point? There seems to be a lot! "Table bugs" sounds pretty ominous too

PR #419, opened 2021-02-26
SearchQuery Optimisation

rohan-gupta8:
[ch3197]

Get rid of an extra POST query being made to `/api/searches` every time a user searches for something on the homepage by creating an object of SearchQuery in ClubViewSet itself.


backend/clubs/views.py
```diff
@@ -1031,8 +1031,19 @@ def get_queryset(self):
...
                 )
 
+        # if there is a search query made by a signed-in user, save it to the database
+        if self.request.query_params.get("search") and isinstance(
+            self.request.user, get_user_model()
+        ):
+            SearchQuery(
+                person=self.request.user,
+                query=self.request.query_params.get("search"),
+                created_at=datetime.datetime.now(),
```
cphalen, line 1041: Pretty sure you can leave out `created_at=datetime.datetime.now()` because the `SearchQuery` is configured to auto set the `created_at` field to the time when created.

rohan-gupta8: Done!

PR #421, opened 2021-03-14
Add field to profile for user to specify type

ezwang:
The user can specify either undergraduate, graduate, or staff member. Doesn't do much currently, but can be useful for later features if we want to provide relevant members with information tailored to their needs.

Will be good to collect this information early in case we need it later on. There is a detection method, but it is inaccurate and unreliable.

image

@ArmaanT Not sure how this might interfere with centralized user account information on Platform, so tagging you.


cphalen:
LGTM!

ArmaanT:
We have a degree type on our Major model ([link](https://github.com/pennlabs/platform/blob/7abbff365b4b2dbde09aca577f34eb44558a3d94/accounts/models.py#L21)) which can differentiate between undergrad and grad. The user should already be in a group called `platform_staff` if they're a staff memeber

backend/clubs/models.py
```diff
@@ -1428,6 +1441,36 @@ class Profile(models.Model):
...
+        Try to detect appropriate values for profile fields based on what platform has
+        returned. Currently only supports detecting the affiliation.
+        This method is not very accurate and should only be used to provide the user
+        an initial guess.
+
+        Overwrites existing information.
+        """
+
+        # detect the affilation
+        if self.user.groups.filter(name="platform_student"):
```
ArmaanT, line 1455: Oof, does this check work because an empty queryset is Falsey? Why is python like this

ezwang: Actually I wouldn't bet on this, nice catch! I'm pretty sure this actually doesn't work haha.

frontend/components/Settings/ProfileForm.tsx
```diff
@@ -80,6 +80,14 @@ const ProfileForm = ({
     }
   }
 
+  const affiliations = [
+    { value: 0, label: 'Undergraduate Student' },
+    { value: 1, label: 'Masters Student' },
+    { value: 2, label: 'Professional Student' },
+    { value: 3, label: 'Doctoral Student' },
+    { value: 4, label: 'Faculty or Staff Member' },
```
ezwang, line 88: @ArmaanT Rerequesting review so I can get your opinion on these labels. Are there better labels that we can use? What is the difference between a masters and a professional student? They generally seem very intertwined when Penn refers to them, would everyone be able to tell which one they should select?

ArmaanT: On platform we're using the labels that Penn used to label degrees ([link](https://catalog.upenn.edu/programs/)). I'm sure Penn isn't very consistent in their naming, but I think this should be good enough for us. There are definitely conflicts (for example an Accounting MBA is both a Master's and a Professional degree) so we just choose the first label that applies. Bachelor's, then Master's, then PhD, then Professional

PR #426, opened 2021-03-26
Import Paideia Events

rohan-gupta8:
[ch3269]

Create a script to import events from https://snfpaideia.upenn.edu/wp-json/wp/v2/event into the Events model.

Modify cron job to do this at a set frequency.

### DIFF
--- a/backend/clubs/management/commands/import_paideia_events.py
+++ b/backend/clubs/management/commands/import_paideia_events.py
@@ -6,9 +6,8 @@ from html import unescape
 
 from django.core.management.base import BaseCommand
 from django.utils import timezone
-from django.utils.html import strip_tags
 
-from clubs.models import Event
+from clubs.models import Club, Event
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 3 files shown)


backend/clubs/management/commands/import_paideia_events.py
```diff
@@ -0,0 +1,69 @@
...
+                    for x in string_time[-8:].split(":")
+                ]
+
+                ev.start_time = timezone.make_aware(
+                    datetime(yyyy, mm, dd, start_hrs + start_24h, start_mins)
+                )
+                ev.end_time = timezone.make_aware(
+                    datetime(yyyy, mm, dd, end_hrs + end_24h, end_mins)
+                )
+
```
cphalen, line 68: It also looks like there is a `event["featured_images"]` which is in HTML (to make life difficult) but we could extract the `src` attribute out of that (probably with regex?) and set `ev.image` accordingly

cphalen, line 68: One last thing, we should set `ev.type = Event.OTHER` — it defaults to `Event.RECRUITMENT` which probably won't be right in their case.

backend/clubs/management/commands/import_paideia_events.py
```diff
@@ -0,0 +1,69 @@
...
+                # prevent duplicates
+                existing = Event.objects.filter(code=event["slug"]).first()
+                if existing:
+                    existing.delete()
+
+                ev = Event()
+                # ev.club = Is there a club for this? What even is SNF Paideia?
+                ev.code = event["slug"]
+                ev.name = unescape(event["title"]["rendered"])
+                ev.description = (
```
cphalen, line 36: I think `event["excerpt"]` might be a better description/introduction. What do you think?

cphalen (commented):
Looks really good! Thanks so much for doing this. I left a few changes but shouldn't be too much.

PR #428, opened 2021-04-02
Batch Remove Tags

cphalen:


backend/clubs/management/commands/once_batch_remove_tags.py
```diff
@@ -0,0 +1,37 @@
...
+        "Remove target tags from orgs which selected all target tags: "
+        "target_year (Degree Type), target_school (Target School), "
+        "student_types (Students Type). This script should only be run "
+        "once on the Hub@Penn database. This script should be deleted "
+        "after it has been run once."
+    )
+    web_execute = True
+
+    def handle(self, *args, **kwargs):
+        target_years_count = Year.objects.distinct().count()
```
rohan-gupta8, line 18: Distinctness is not actually relevant here right? Otherwise could mess with the way you filter by `target_years_count` and such.

Should we delete this script or just mention that it's shelved somewhere as a comment/wherever? Could be useful sometime in the future

LGTM otherwise!

ezwang (commented):
You should be able to do this all from the Django admin console.

image

The admin console will also take care of any foreign key and many to many relationships when deleting models.

cphalen:
Ok sounds goods thanks Eric and Rohan! I'll close this branch and do it through admin console.

ezwang:
My bad, this isn't possible through the admin console.

LGTM!

backend/clubs/management/commands/once_batch_remove_tags.py
cphalen: Ok removed the `distinct()` from the query. Thanks king!

PR #433, opened 2021-04-11
Add ability for umbrella organizations to edit badges

ezwang:
image

### DIFF
--- a/frontend/components/ClubEditPage.tsx
+++ b/frontend/components/ClubEditPage.tsx
@@ -37,6 +37,7 @@ import {
   OBJECT_TAB_RECRUITMENT_LABEL,
   SHOW_APPLICATIONS,
   SHOW_MEMBERSHIP_REQUEST,
+  SHOW_ORG_MANAGEMENT,
   SITE_NAME,
 } from '../utils/branding'
 import AdvisorCard from './ClubEditPage/AdvisorCard'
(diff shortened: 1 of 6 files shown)


frontend/components/ClubEditPage.tsx
```diff
@@ -350,6 +351,15 @@ const ClubForm = ({
...
       },
+      {
+        name: 'organization',
+        label: 'Club Management',
+        content: (
+          <>
+            <ClubManagementCard club={club} />
+          </>
+        ),
+      },
```
cphalen, line 362: One question I have is that Umbrella Orgs don't really exist on Hub@Penn as far as I can tell, do you think this would be worth only displaying on Penn Clubs? Lmk what your thoughts are, I can go either way on this one.

cphalen (approved):
LGTM! Just one question but otherwise good to merge!

frontend/components/ClubEditPage.tsx
ezwang: Good point, I didn't bother disabling this on Hub@Penn but ideally this would be hidden on Hub@Penn.

ezwang: Disabled!

cphalen: Awesome thanks Eric!

PR #437, opened 2021-04-23
Run cypress tests in parallel

ArmaanT:
This needs a few fixes but maybe it can work @cphalen

### DIFF
--- a/frontend/test.sh
+++ b/frontend/test.sh
@@ -13,4 +13,4 @@ apt-get update && apt-get install -y libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify
 
 
 # Run tests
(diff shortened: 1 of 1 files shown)


rohangpta:
@joyliu-q @alxkp @cphalen  I know we were talking about faster CI at some point. Is anyone able to take a look at this?

joyliu-q:
Just looked at docs and this seems reasonable to me. Any reason push integration tests are failing though?

rohangpta:
image
This seems to be reason for failure.

joyliu-q:
@rohangpta I also don't think we're passing in the record key via `--key`, which is also required. Can we generate and add the record key as a ENV variable?

PR #444, opened 2021-05-18
[WIP] Centralized application

cphalen:


backend/clubs/serializers.py
```diff
@@ -2222,8 +2228,120 @@ class Meta:
...
+        # manually create committee choices as Django does not
+        # support nested serializers out of the box
+        if "committees" in request:
+            committees = request["committees"]
+            question_obj.committees.clear()
+            for committee in committees:
+                committee_obj = ApplicationCommittee.objects.filter(
+                    application=question_obj.application, name=committee["name"]
+                ).first()
+                question_obj.committees.add(committee_obj)
```
ArmaanT, line 2297: Does Eric's ManyToMany Save Mixin not help here?

backend/clubs/serializers.py
```diff
@@ -2222,8 +2228,120 @@ class Meta:
...
+class ApplicationSubmissionSerializer(serializers.ModelSerializer):
+    committee = ApplicationCommitteeSerializer(required=False, read_only=True)
+    responses = ApplicationQuestionResponseSerializer(
+        many=True, required=False, read_only=True
+    )
+    user_hash = serializers.SerializerMethodField("get_user_hash")
+
+    def get_user_hash(self, obj):
+        username = obj.user.username
+        return int(hashlib.sha1(username.encode("utf-8")).hexdigest(), 16) % (10 ** 8)
```
ArmaanT, line 2325: I assume this isn't supposed to be sensitive, but do you think it makes sense to hash the concatenation of PennKey and date joined or something like that just so that way it's much harder to reverse engineer values?

backend/clubs/serializers.py
```diff
@@ -2247,6 +2365,21 @@ def validate(self, data):
...
+        # manually create committee objects as Django does
+        # not support nested serializers out of the box
+        request = self.context["request"].data
+        if "committees" in request:
+            committees = request["committees"]
+            ApplicationCommittee.objects.filter(application=application_obj).delete()
+            for committee in committees:
+                ApplicationCommittee.objects.create(
+                    name=committee["value"], application=application_obj,
+                )
```
ArmaanT, line 2379: Ditto on the M2M Mixin comment

backend/clubs/views.py
```diff
@@ -4083,7 +4090,7 @@ class UserViewSet(viewsets.ModelViewSet):
     queryset = get_user_model().objects.all().select_related("profile")
     permission_classes = [ProfilePermission | IsSuperuser]
     filter_backends = [filters.SearchFilter]
-    http_method_names = ["get"]
+    http_method_names = ["get", "post"]
```
ArmaanT, line 4093: Just want to confirm we aren't currently using the fact that this was a GET only route as some kind of security feature or check and the serializer has proper checks on which fields are writable

ArmaanT (commented):
I haven't looked too closely at the backend code, but here are a couple of things I noticed

PR #455, opened 2021-07-24
first commit

rohan-gupta8:
Thinking that we will have a frontend platform (in the internal club management page?) to create multiple instances of this class which clubs can fill in and save (and preview, in one of our to-be-created boilerplate templates). They can then customize sending options like 'send on submission of app, send to all applicants, etc', which we'll then create an endpoint to handle on the backend.

Important question - how many clubs will want to use our service as opposed to just doing their own thing, once
they receive apps and corresponding applicants' emails?


cphalen (approved):
LGTM! Great work

rohangpta:
Funny how we basically came full circle with #551. This is effectively done. Penn Clubs now supports arbitrary Jinja2 templates as with custom text and variables.

PR #460, opened 2021-08-25
Favourite Events Endpoint

rohan-gupta8:
As per @01jongmin 's request, an API endpoint to return all events that are associated with user favourited clubs.


cphalen (approved):
Looks good to me!

PR #467, opened 2021-09-11
app due date and pointer on hover over wharton club application

aydangoon:
* Application due date included under the title.
* Cursor changes to a pointer when hovering over a Wharton club application on the `/apply` page. imo more intuitive but either works.


cphalen (approved):
LGTM!

PR #469, opened 2021-09-17
Fix Page Navigation

alnasir7:
- fix bug where single forward takes to the last page.


cphalen:
I think this bug has actually already been fixed! Thanks for making this PR though.

PR #471, opened 2021-09-19
Download to CSV

cphalen:
Added download to CSV functionality as well as some other bug fixes, primarily in the club officer reading view


aydangoon:
Looks good!

PR #473, opened 2021-09-24
List bug fix

aydangoon:
Story: 2594
* Fixed inconsistent bookmark and subscriber states when clicking between list and card view
* Added context provider for displayed clubs (to update state without reloads), would be good to get some other frontend opinions on the design choice!


cphalen (approved):
LGTM!

PR #475, opened 2021-10-03
Likes

Dfeng6789:
created like and unlike functionality, updated question answer model to contain a many to many field tracking the users who liked it, also updated the view with two endpoints for like and unlike and the serializer to display like count


backend/clubs/views.py
```diff
@@ -2638,6 +2639,33 @@ def get_queryset(self):
...
+        question.users_liked.add(self.request.user)
+        question.save()
+        return Response({})
+        
+
+    @action(detail=True, methods=["post"])
+    def unlike(self, request, *args, **kwargs):
+        """
+        Endpoint used to unlike a question answer.
+        ---
```
cphalen, line 2661: Replace
```
"""
Endpoint used to unlike a question answer.
---
"""
```
with
```
"""
Endpoint used to unlike a question answer.
---
requestBody: {}
responses:
    "200":
        content: {}
---
"""
```

backend/clubs/views.py
```diff
@@ -2638,6 +2639,33 @@ def get_queryset(self):
 
         return questions.filter(Q(approved=True) | Q(author=self.request.user))
 
+    @action(detail=True, methods=["post"])
+    def like(self, request, *args, **kwargs):
+        """
+        Endpoint used to like a question answer.
+        ---
+        """
```
cphalen, line 2647: Replace
```
"""
Endpoint used to like a question answer.
---
"""
```
with
```
"""
Endpoint used to like a question answer.
---
requestBody: {}
responses:
    "200":
        content: {}
---
"""
```

backend/tests/clubs/test_views.py
```diff
@@ -1796,6 +1796,15 @@ def test_club_question_answer(self):
         # ensure email was sent out notifying officer of question
         self.assertEqual(len(mail.outbox), 1, mail.outbox)
 
+        resp = self.client.post(reverse("club-questions-like"), args=("test-club", "2",))
```
cphalen, line 1799: I think this needs to be
```
resp = self.client.post(reverse("club-questions-like", args=("test-club", "2",)))
```
so just swapping up the placement of parentheses

cphalen (changes requested):
Looks awesome! I think there are just a few things and then we should be ready to merge:

1. Can you delete the `Pipfile` and `Pipfile.lock` from the base directoy (not from `/backend`)? I'm guessing you ran `pipenv shell` in the base directory instead of in `/backend` and it generates those automatically. If you just delete them locally and then `git add .` it should remove them from the PR.
2. Run the linters to make sure your code is properly styled (line limits, blank lines, etc.). After you run `pipenv shell` in `/backend` you can run `black .` and `flake8 .` — these commands will tell you what you need to change to update the style.
3. Update the `test_check_detail_level_permissions` function in `/backend/tests/clubs/test_security.py` to include the like and unlike action in the `whitelist`. Normally subverting the security tests like this is bad, but in this case want this endpoint to be publicly available (because they don't return any information, let alone personal information) so it's ok.
4. Just a few inline suggestions I also made in the PR.

Really nice work! No rush on this just wanted to give you feedback.

backend/tests/clubs/test_views.py
rohan-gupta8: @Dfeng6789 Looks like this was the issue we were running into! Always good to have a fresh set of eyes looking over it to catch silly things like this. 

Anyways, ping me if you need any further help with our testing setup to wrap this up, or if you want a refresher on the commands to run to execute test cases.

backend/clubs/views.py
```diff
@@ -2617,6 +2617,7 @@ class QuestionAnswerViewSet(viewsets.ModelViewSet):
 
     serializer_class = QuestionAnswerSerializer
     permission_classes = [QuestionAnswerPermission | IsSuperuser]
+    lookup_field = "id"
```
rohan-gupta8, line 2620: In general,`lookup_field` defaults to `id`, so this is not strictly necessary. Feel free to remove it.

rohan-gupta8 (commented):
Left some feedback inline, but generally echoing what Campbell said. Looks great overall -- hopefully this is in production soon!

backend/Pipfile
```diff
@@ -50,6 +50,7 @@ urlextract = "*"
 jsonref = "*"
 unittest-xml-reporting = "*"
 tblib = "*"
+colorama = "*"
```
cphalen, line 53: Do we need this dependency? I just delete this from the `Pipfile` unless there is something we need it for (don't want to download more dependencies than necessary).

backend/tests/clubs/test_security.py
```diff
@@ -138,7 +138,7 @@ def test_check_detail_level_permissions(self):
         """
 
         # Don't put your function here unless it never returns any private information.
-        whitelist = set()
+        whitelist = {("QuestionAnswerViewSet", "like"), ("QuestionAnswerViewSet", "unlike")}
```
cphalen, line 141: Looks like flake8 wants this line to be under 88 characters
```
./tests/clubs/test_security.py:141:89: E501 line too long (92 > 88 characters)
```
can you reformat it? Maybe
```
whitelist = {
    ("QuestionAnswerViewSet", "like"),
    ("QuestionAnswerViewSet", "unlike"),
}
```
would work?

cphalen (approved):
Looks good! After you make those two changes I mentioned we should be good to merge this into production. Awesome work!!!

backend/Pipfile
Dfeng6789: This was just because `black .` wasn't running before, but I've removed it. Other than that I've implemented the other changes, although every time I run pipenv install it'll change one line in my Pipfile.lock, I'm not sure if that's a problem?

backend/tests/clubs/test_security.py
```diff
@@ -138,7 +138,10 @@ def test_check_detail_level_permissions(self):
         """
 
         # Don't put your function here unless it never returns any private information.
-        whitelist = set()
+        whitelist = {
+            ("QuestionAnswerViewSet", "like"),
```
cphalen: Looks like there's a trailing whitespace on this line now lol sorry the formatter is kinda strict
```
./tests/clubs/test_security.py:142:47: W291 trailing whitespace
```
generally it's a good idea to just run `black .` and `flake8 .` before you commit just so that the formatter will tell you want to change before you commit.

Dfeng6789: Yeah I think I might have to push again also sorry

cphalen: Ahhhh never mind you caught it well done. I'll merge it in after these tests pass!

PR #481, opened 2021-10-08
Likes Frontend

Kelvin-Hui:
- Created like button for each FAQ questions.
- Allow user to select most liked question / most recent question.
- Changed Smooth scroll for better UX.


aydangoon:
I agree with Campbell's suggestion of not waiting on `apiSetLikeStatus` and had one small comment, but overall looks great man!

aydangoon (approved):

PR #484, opened 2021-10-10
migrate over Event and Item models from CFA

Dfeng6789:
migrated over the Event and Item models from CFA, created oneToOne between Events in Penn Clubs and the new FundedEvents from CFA


rohangpta:
Closing this because we've decided to not go ahead with the CFA <> Penn Clubs migration. For posterity, the current plan is to include a link to the CFA (#545) on the Penn Clubs homepage and separately write improvements/maintenance to the CFA codebase.

Thank you David for all your work on this!

PR #488, opened 2021-10-24
hide filters in clubs

alnasir7:
- Hid the school and year filters for Penn Clubs, these should only be viewable in hub@penn.


aydangoon (approved):
LGTM!

PR #490, opened 2021-11-01
Disable Sentry PII

cphalen:


cphalen:
Is there anything else to this? The frontend also has Sentry config (`frontend/utils/sentry.tsx`) but I couldn't find anything about PII there. Thanks Armaan!

ArmaanT (approved):
LGTM! You could also just remove the flag entirely since it defaults to false

PR #493, opened 2021-11-07
temp button to select all applications

aydangoon:
Cringe but it works. I think our table component in `common/Table` should support selectable rows as a feature (just feels like there are so many use cases), I'll play around with it.


cphalen:
> I think our table component in `common/Table` should support selectable rows as a feature (just feels like there are so many use cases), I'll play around with it.
Definitely would be a good idea to add this as a feature to the Table class! Maybe we can put that on Clubhouse and it can be a task for Kelvin? He's been looking for something to work on I believe.

cphalen (approved):

PR #495, opened 2021-11-07
Add ID and creator field to admin notes

xiayufeilucy:
Return id and creator of notes in admin notes endpoint.


rohan-gupta8 (approved):
LGTM!

PR #506, opened 2022-02-11
applicants page style improvment

alnasir7:
- Made the applicants page scrollable.
- Put scroll bar on top


rohangpta:
@alnasir7 Probably good to fix lint, rebase and merge this, right? I don't see anything that could remotely be breaking.

PR #510, opened 2022-02-20
changed default view in events page, changed toggle view button

DavidZhng:


rohangpta:
LGTM! Great work David.

PR #542, opened 2022-09-19
Deleting application submissions + version bump

cphalen:
Now deleting an application submission on a Wharton application will delete that submission and all other submissions by the same user on the same application under the same committee. This should resolve some of the complications of deleting submissions.

Also did some version bumps, largest one is to Django 4 but was pretty seamless just needed a few config changes.


rohangpta:
@cphalen Are we thinking this version bump is a good idea? It seems like a lot and might break stuff, but Django 4.1 does seem pretty cool especially with all its async support. Let me know what you think!

PR #544, opened 2022-10-02
added cycle model

tobinryan:
added the application cycle model to the models


rohangpta:
Addressed by #565. Thanks for laying the groundwork on this Ryan!

PR #565, opened 2023-01-05
Application Improvements

rohangpta:
- Application closing / opening strict enforcement
- Add table filters to submission views, fix N+1s (factor of ~100 query reduction -- should help with DB throttling).
- Add application cycles and tweak the Wharton council app script to group new applications by cycle
- Add per-page caching to the application list, club list, and submission list views (some of our heaviest views). Important to cache entire page responses since it turns out the bottleneck in our performance is the serializer (about 70-80% of the submission view is spent in the serializer).


aviupadhyayula:
LGTM, Rohan! Thanks for the great work.

PR #572, opened 2023-02-11
Migrate drf-renderer-xlsx to drf-excel

cphalen:
Summary of why this is needed:

1. Openpyxl 3.1.0 deprecated a function that `drf-renderer-xlsx` used called `save_virtual_workbook`
2. `drf-excel` patched this issue, but in order to get the patch we need to move to `drf-excel` (the new version of `drf-renderer-xlsx`)
3. I axed `drf-renderer-xlsx`, installed `drf-excel`, and updated `base.py` to use `drf-excel` instead.

Hopefully this doesn't result in a bunch of problems because of the version bump. If it does, we can just fix `openpyxl==3.0.10`.


rohangpta:
Great work, thanks. Have you tested to see if excel rendering breaks on an API endpoint?

cphalen:
> Great work, thanks. Have you tested to see if excel rendering breaks on an API endpoint?

Just tested this and I think we should be good! Thanks for pointing this out breaking the CSVs or reports would've been ungood lol

rohangpta:
Thank you Campbell! Merging

rohangpta (approved):

PR #578, opened 2023-04-13
Adding escape character to zoom link regex

joel8019:
The regex for matching zoom links allowed for non-domain formats.
For example the following could match:   https://fredazoomaus/j/123/pwd=

See the link below for fix invalidation
https://regex101.com/r/xPe7AD/1


rohangpta:
Thank you for the submission!

PR #579, opened 2023-08-14
fix: Deprecate membership_count

esinx:
`InfoBox` uses `membership_count` which does not exist anymore. `membership_count` should be removed from `Club` type definition. Seems like `InfoBox` is the only component making use of this property, so we might as well delete it, but I marked it as deprecated for safety. See [types.ts](https://github.com/pennlabs/penn-clubs/blob/fix/registered-count/frontend/types.ts#L176).

Screenshot 2023-08-14 at 2 59 01 PM


rohangpta:
Good catch, just in time for the new school year. Thanks Eunsoo! Good to have you back

rohangpta (approved):

rohangpta:
Feel free to squash and merge whenever.

Issue #580, opened 2023-09-01
Create an interface for Wharton Council admins for fine grained control over applications


rohangpta:
Create an interface for WC admins to add/delete club applications (likely a view for them to run the `wharton_council_application` script?)

- An interface visible to admins of Wharton Council only (using Django permissions)
- Control centralised application `start_time`, `end_time`, `application_cycle` and `result_release_time`, checkbox of clubs to include, and control over whether these applications are editable (currently, Wharton apps are not editable when submissions are open)
- Changes to applications need to be approved by Wharton Council (includes edits and deletions)

Frontend - Julian
Backend - Rohan Moniz / Avi

rohangpta:
@rm03 @aviupadhyayula @julianweng let me know what you all think!

rohangpta:
Done!

PR #590, opened 2023-09-25
WC Admin application cycle edit/create table

julianweng:
Implemented ModelForm on Wharton Council Admin page, connecting to WhartonCyclesView (enforces cycle start/end times on all applications with given cycle upon update).


.pre-commit-config.yaml
```diff
@@ -27,7 +27,7 @@ repos:
         args: [--config, backend/setup.cfg]
       - id: frontend
         name: Yarn Linter
-        entry: yarn --cwd frontend lint
+        entry: bash -c "cd frontend && yarn lint"
```
rohangpta, line 30: What was the issue with `--cwd`?

backend/clubs/serializers.py
```diff
@@ -96,6 +97,24 @@ def save(self):
         return super().save()
 
 
+class ApplicationCycleSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = ApplicationCycle
+        fields = ["id", "name", "start_date", "end_date"]
```
rohangpta, line 103: Do we need id?

backend/clubs/views.py
```diff
@@ -4861,6 +4863,39 @@ def get_queryset(self):
...
+    """
+
+    permission_classes = [WhartonApplicationPermission | IsSuperuser]
+    http_method_names = ["get", "post", "put", "patch", "delete"]
+    serializer_class = ApplicationCycleSerializer
+
+    def get_queryset(self):
+        return ApplicationCycle.objects.all().order_by("end_date")
+
+    def list(self, request, *args, **kwargs):
```
rohangpta, line 4879: Not 100% but don't think you need this, list should exist already and will have exactly the behaviour of calling the super class' method

backend/clubs/views.py
```diff
@@ -4861,6 +4863,39 @@ def get_queryset(self):
...
 
 
+class WhartonCyclesView(viewsets.ModelViewSet):
+    """
+    update: Update application cycle
+    list: Get list of all application cycles
+    """
+
+    permission_classes = [WhartonApplicationPermission | IsSuperuser]
+    http_method_names = ["get", "post", "put", "patch", "delete"]
```
rohangpta, line 4873: Do we need `put`, `patch`, and `delete`?

Docstring only specifies update and list?

backend/clubs/views.py
```diff
@@ -4861,6 +4863,39 @@ def get_queryset(self):
...
+        """
+        applications = ClubApplication.objects.filter(
+            application_cycle=self.get_object()
+        )
+        for app in applications:
+            app.application_start_time = self.get_object().start_date
+            app.application_end_time = self.get_object().end_date
+            if app.result_release_time < app.application_end_time:
+                app.result_release_time = app.application_end_time
+                +datetime.timedelta(days=10)
```
rohangpta, line 4894: This seems to pass lint but looks a bit awkward?

rohangpta, line 4894: Why +10 days always? Should we allow the user to pass in a field controlling the cycle's result release time?

backend/clubs/views.py
```diff
@@ -4861,6 +4863,39 @@ def get_queryset(self):
...
+            app.save()
```
rohangpta, line 4895: This won't affect prod performance but it would be good practice to use `bulk_update`!

rohangpta:
Great work! Left a few comments but otherwise looks like this is a good start towards making administration of applications easier! Thanks Julian

backend/clubs/views.py
julianweng: Was a placeholder since release_time isn't currently a field in application_cycle, in order to avoid validation errors in the [email serializer](https://github.com/pennlabs/penn-clubs/blob/8dae4d6527fee32731d33e9f08ca45a65c32914b/backend/clubs/serializers.py#L2672) , there's probably a better UI solution to maintain the different invariants plus adding release_time ofc

.pre-commit-config.yaml
julianweng: Behavior was different for some reason in my Windows environment vs a manual cd, --cwd raised a "no penn-clubs/Frontend/Frontend" error, not entirely sure why

rohangpta: Oh interesting, well both ways work so no worries with me!

backend/clubs/views.py
julianweng: Forgot to update the docstring, thanks for the catch! Default behavior for `post` and `delete` are used in `ModelForm` to let the user create delete cycles I believe, and `patch` is used for update.

backend/clubs/views.py
julianweng: Tested and it works without, thanks!

rohangpta:
Refs issue #580

backend/clubs/serializers.py
julianweng: Tested it without and it seemed to prevent editing / deletion of fields after page refresh, I assume the numerical id is used somewhere in those operations, but could be wrong.

backend/clubs/views.py
rohangpta: I meant more of:
- do we want to allow deleting?
- we should consider picking one of put/patch, they are generally considered to have similar semantics ("put" is idempotent, whereas "patch" is not always)

.pre-commit-config.yaml
aviupadhyayula: Looks like it's a `[known issue](https://github.com/yarnpkg/berry/issues/4908)`?

frontend/components/Settings/WhartonApplicationCycles.tsx
```diff
@@ -0,0 +1,300 @@
...
+}
+
+const ScrollWrapper = styled.div`
+  overflow-y: auto;
+  margin-top: 1rem;
+  height: 40vh;
+`
+
+const WhartonApplicationCycles = (): ReactElement => {
+  const [editMembership, setEditMembership] = React.useState(false)
```
esinx, line 42: Why not simply import `useState` from React?

frontend/components/Settings/WhartonApplicationCycles.tsx
```diff
@@ -0,0 +1,300 @@
...
+  const [membershipCycle, setMembershipCycle] = React.useState({
+    name: '',
+    id: null,
+  })
+
+  const [editExtensions, setEditExtensions] = React.useState(false)
+  const [extensionsCycle, setExtensionsCycle] = React.useState({
```
esinx, line 49: Can we type `extensionCycle` and `membershipCycle`?

frontend/components/Settings/WhartonApplicationCycles.tsx
```diff
@@ -0,0 +1,300 @@
...
+              exception: club.application_end_time_exception,
+              changed: false,
+            }
+          })
+          setClubsExtensions(initialOptions)
+        })
+    }
+  }, [extensionsCycle])
+
+  if (clubOptionsMembership == null) {
```
esinx, line 167: Suggestion: We should try to use something like react-query or SWR to streamline loading process

esinx (approved):
Still fine, but minor type changes may be appreciated!

rohangpta:
Looks like we're pretty close to ready to merge this? @julianweng do you have any time to get this merged in before we start school?

PR #592, opened 2023-09-26
wip-feature: Allow resend of acceptance/rejection emails

### LINKED ISSUE #585: Allow admins to resend acceptance/rejection emails to applicants
Admins can only send acceptance/rejection emails to club applicants once. If an acceptance/rejection email template is initially incorrect and emails are sent, applicants cannot be re-notified. 

For reference, the logic for this is implemented in `Backend/clubs/views.py` under `ClubApplicationViewSet`'s `send_emails()`.

Proposed fix: add an option for admins to force renotification for all applicants. Ideally, this would show up as a checkbox when sending emails, with a warning upon toggle that confirms whether admins mean for applicants to be re-notified. 

Screenshot 2023-09-23 at 9 07 53 PM

aviupadhyayula:
Creates an option for club admins to resend all acceptance/rejection emails, regardless of whether applicants have been previously notified or not. Resolves #585. 

Still a WIP. To-do: 
- [ ] Add tests
- [x] Create stronger checks against people accidentally pressing this. Maybe we can add a pop-up or change the helpText to deter admins resending emails without good reason to. (CC'ing @julianweng)

What it looks like:
Screenshot 2023-09-25 at 11 38 18 PM

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
         args: [--config, backend/setup.cfg]
       - id: frontend
         name: Yarn Linter
-        entry: yarn --cwd frontend lint
+        entry: bash -c "cd frontend && yarn lint"
         language: system
         files: ^frontend/
         require_serial: false
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -4665,6 +4665,8 @@ class ClubApplicationViewSet(viewsets.ModelViewSet):
         Send out acceptance/rejection emails for a particular application
 
         Dry run will validate that all emails have nonempty variables
+
+        Allow resend will renotify submissions that have already been emailed
         ---
         requestBody:
             content:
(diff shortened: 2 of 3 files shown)


rohangpta:
Can we get a review @julianweng @esinx @rm03?

Would be lovely to get you all some practice with reviewing code :)

esinx (approved):
Looks good to me!
Might be an unrelated question, but are we somehow tracking which mails have been sent at what time? Might be useful to display a log of what mails have been sent through Penn Clubs.

esinx:
@aviupadhyayula please feel free to merge whenever you feel comfortable about this PR!

aviupadhyayula:
> Might be an unrelated question, but are we somehow tracking which mails have been sent at what time? Might be useful to display a log of what mails have been sent through Penn Clubs.

Not that I know of, but this seems easy enough to implement. We could create a wrapper class around each email and append it to some global array. An alternative is using some package like [`django-mailer`](https://github.com/pinax/django-mailer), which would store outgoing emails in a database. 

Is there any reason we'd need this functionality? Can look into it with future PRs.

PR #598, opened 2023-10-14
temp:(banner)

esinx:
Screenshot 2023-10-14 at 5 57 02 PM

### DIFF
--- a/frontend/components/Announcement.tsx
+++ b/frontend/components/Announcement.tsx
@@ -1,0 +1,211 @@
+import { useEffect, useRef, useState } from 'react'
+
+const issueIcon =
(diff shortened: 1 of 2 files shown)


joyliu-q (commented):
Hey! I saw that we have the `Announcement.tsx` component added to PRs in the other repos as well. Any chance we could factor this out into one of platform's components (saw you making a platform import). Also, you can publish platform libraries using the `@pennlabs` account on `NPM` (login included on bitwarden).

esinx:
> Hey! I saw that we have the `Announcement.tsx` component added to PRs in the other repos as well. Any chance we could factor this out into one of platform's components (saw you making a platform import). Also, you can publish platform libraries using the `@pennlabs` account on `NPM` (login included on bitwarden).

It would be great if we can achieve this in the given time, but through some trial and error I figured that we cannot have 1 version of the shared component library in the current configuration since some of our web products are using outdated version of react/nextjs which messes with server side rendering logic & breaks component import. I will have to investigate more on this (making the @pennlabs/ui package more universally available) and fix this as the hard-coded approach is merely temporary.

PR #602, opened 2023-11-24
Individual Application Extensions

rm03:
This feature allows club admins to grant extensions to individual applicants.


backend/clubs/serializers.py
```diff
@@ -2424,6 +2425,51 @@ class Meta:
...
+        if not user:
+            raise serializers.ValidationError("Please provide a valid username!")
+        return value
+
+    def create(self, validated_data):
+        username = validated_data.get("user").pop("username")
+        validated_data["user"] = get_user_model().objects.get(username=username)
+
+        application_pk = self.context["view"].kwargs.get("application_pk")
+        validated_data["application"] = ClubApplication.objects.filter(
```
rohangpta, line 2458: Thoughts on validating `application_pk`?

backend/clubs/serializers.py
```diff
@@ -2424,6 +2425,51 @@ class Meta:
...
+            pk=application_pk
+        ).first()
+
+        return super().create(validated_data)
+
+    def update(self, instance, validated_data):
+        user_field = validated_data.pop("user", None)
+        if user_field:
```
rohangpta, line 2466: Consider using the walrus `:=` operator

backend/clubs/views.py
```diff
@@ -4456,10 +4458,17 @@ def question_response(self, *args, **kwargs):
 
         # prevent submissions outside of the open duration
         now = timezone.now()
-        if (
-            now > application.application_end_time
-            or now < application.application_start_time
-        ):
+        extension = (
```
rohangpta, line 4461: Consider `select_related` to avoid an N+1

backend/clubs/models.py
```diff
@@ -1592,6 +1592,18 @@ def validate_template(cls, template):
         return all(t in cls.VALID_TEMPLATE_TOKENS for t in tokens)
 
 
+class ApplicationExtension(models.Model):
```
rohangpta, line 1595: Consider a [unique_together](https://docs.djangoproject.com/en/4.2/ref/models/options/#unique-together) constraint on `user` and `application`

backend/clubs/views.py
```diff
@@ -4456,10 +4458,17 @@ def question_response(self, *args, **kwargs):
...
+            application.extensions.filter(user=self.request.user)
+            .order_by("-end_time")
```
rohangpta, line 4463: Unique together constraint should allow you to get rid of this `order_by`!

backend/clubs/views.py
```diff
@@ -4806,11 +4815,13 @@ def current(self, *args, **kwargs):
...
-        return Response(
-            ClubApplicationSerializer(
-                qs.filter(application_end_time__gte=timezone.now()), many=True
-            ).data
-        )
+        now = timezone.now()
+        user = self.request.user
+        q = Q(application_end_time__gte=now)
+        if user.is_authenticated:
+            q |= Q(extensions__end_time__gte=now, extensions__user=user)
```
rohangpta, line 4822: Is `select_related` applied on extensions? Otherwise we have an N+1

backend/clubs/views.py
```diff
@@ -4950,6 +4961,16 @@ def get_queryset(self):
         )
 
 
+class ApplicationExtensionViewSet(viewsets.ModelViewSet):
+    permission_classes = [ClubSensitiveItemPermission | IsSuperuser]
+    serializer_class = ApplicationExtensionSerializer
+
+    def get_queryset(self):
```
rohangpta, line 4968: Consider using `lookup_field` instead

`get_queryset` defaults to `ApplicationExtension.objects.all()` I believe that is what we want?

frontend/pages/club/[club]/application/[application]/index.tsx
```diff
@@ -194,24 +194,6 @@ const ApplicationPage = ({
               }
             }
 
-            // submissions open & close error check
```
Comment (rohangpta) [inline frontend/pages/club/[club]/application/[application]/index.tsx:197]: Why are we getting rid of this?

backend/clubs/views.py
```diff
@@ -4950,6 +4961,16 @@ def get_queryset(self):
         )
 
 
+class ApplicationExtensionViewSet(viewsets.ModelViewSet):
+    permission_classes = [ClubSensitiveItemPermission | IsSuperuser]
```
rohangpta, line 4965: Nice job with this!

backend/clubs/serializers.py
```diff
@@ -2424,6 +2425,51 @@ class Meta:
...
+            username = user_field.pop("username")
+            user = get_user_model().objects.get(username=username)
+            instance.user = user
```
rohangpta, line 2469: Why is user not read only? Is there a case where we change the user we give the extension to?

backend/clubs/serializers.py
rohangpta: And also the `end_time` should be greater than the existing end time...then you can get rid of the `max()` stuff below

rm03: Yep sorry, meant to open this as a draft PR as I was gonna update validation logic. Also, I added `max()` in case the club updates their overall application deadline after the individual extension is issued

frontend/pages/club/[club]/application/[application]/index.tsx
Comment (rm03) [reply frontend/pages/club/[club]/application/[application]/index.tsx:197]: Wasn't sure why this was necessary since we are performing validation on the backend, and adding the extension logic requires changing this check on the frontend

backend/clubs/serializers.py
rm03: If the club admin makes a typo in the user's pennkey, I figured that we should allow them to update the user. However they could also just delete the extension and create a new one, so open to making this read-only. Lmk what you think

backend/clubs/views.py
rohangpta: Ignore -- not an N+1. my bad!

backend/clubs/serializers.py
rohangpta: I was envisioning a UI where the club admin selects users from a dropdown of applicants...so no typo potential. What do you think?

frontend/pages/club/[club]/application/[application]/index.tsx
Comment (rohangpta) [reply frontend/pages/club/[club]/application/[application]/index.tsx:197]: I think I initially added on the frontend because it would save us an unnecessary call to the backend. But if the frontend is no longer doing bookkeeping on extensions, then I'm ok with this

backend/clubs/serializers.py
rm03: Hmm, wouldn't this require the applicant to submit to appear in the system before they can be granted an extension? It's definitely a minor detail but I think it might be better to leave the responsibility solely with the club admin

This is what the UI currently looks like:
image

rohangpta: I think that makes sense, that's a fair point.

Can we consider emailing applicants that have received extensions letting them know they've been granted one?

rm03: Sure! I can add that

backend/clubs/serializers.py
```diff
@@ -2462,13 +2456,50 @@ def create(self, validated_data):
...
+            if not user:
+                raise serializers.ValidationError("Please provide a valid username!")
+
+        application_pk = self.context["view"].kwargs.get("application_pk")
+        application = ClubApplication.objects.filter(pk=application_pk).first()
+
+        if not application:
+            raise serializers.ValidationError("Invalid application id!")
+
+        if (
```
rohangpta, line 2479: Nit: autoformatter separates this across a lot of lines...can you potentially separate out the logic for readability?

backend/clubs/views.py
```diff
@@ -4814,7 +4810,7 @@ def current(self, *args, **kwargs):
                                 - $ref: "#/components/schemas/ClubApplication"
         ---
         """
-        qs = self.get_queryset()
+        qs = self.get_queryset().prefetch_related("extensions")
```
rohangpta, line 4813: Should this be `prefetch_related` or `select_related`?

rm03: I believe `prefetch_related` since `Extension` is a reverse foreign key on `ClubApplication`

rohangpta:
@rm03 ready to merge?

PR #603, opened 2023-11-28
Upsert Application Submissions

rm03:
Previously, we would create new `ApplicationSubmission` and `ApplicationQuestionResponse` objects every time a user submitted their application; this change ensures that each user will have at most one submission per application per committee.

Going forward, we should be able to remove raw SQL queries like [this](https://github.com/pennlabs/penn-clubs/blob/2814025e7853a02bfe2ac2431a2013f910f00ec7/backend/clubs/views.py#L5290C21-L5299C78) since users won't have multiple submissions associated with an application, but this would be potentially problematic for past applications unless we run a script that drops all submissions that aren't the most recent.

### DIFF
--- a/backend/clubs/models.py
+++ b/backend/clubs/models.py
@@ -1695,6 +1695,9 @@ class ApplicationSubmission(models.Model):
     def __str__(self):
         return f"{self.user.first_name}: {self.application.name}"
 
+    class Meta:
+        unique_together = (("user", "application", "committee"),)
+
 
 class ApplicationQuestionResponse(models.Model):
     """
@@ -1722,6 +1725,9 @@ class ApplicationQuestionResponse(models.Model):
 
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        unique_together = (("question", "submission"),)
 
 
 class QuestionResponse(models.Model):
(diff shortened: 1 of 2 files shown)


rohangpta (approved):
Somehow I missed this! Looks great, I'm going to try and run some benchmarks on throughput with this code before merging in. 

Curious if moving from `create` to `update_or_create` will give us the performance benefits that we hypothesised

rohangpta:
RE: Raw SQL queries -- it's slightly tricky to figure out how exactly to migrate this logic. An approach we can consider:

- Compute the inverse of the queryset returned under RawSQL (everything **but** the most recent for each app) and delete them. Do this by hand via SSH (do you have access to this?).
- Migrate logic to omit the `RawSQL` cruft.
- For sanity, add a `unique_together` check on "application", "committee" and "user" for the `ApplicationSubmission` model (and maybe something of the sort to `ApplicationQuestionResponse`).

If we're doing this, let's also get stats on how much extra data we were storing. I'd imagine we package all these changes + other optimisations into a nice blog post by end of this month.

rm03:
As discussed:
- We had a total of 40,457 `ApplicationSubmission` objects stored in the database and 24,112 of them are not most recent submissions.
- Additionally, there were 165,269 `ApplicationQuestionResponse` objects associated with these submissions (out of a total of 274,445), which means that we've been storing quite a bit of extra data. 

I've removed these objects from the database so the last commit shouldn't be problematic.

PR #605, opened 2024-01-22
Backend version updates

rm03:
This PR upgrades Django from version 3.2 to 5.0 (should be able to try out `GeneratedField` in #601!) and sets the Python version to 3.11 from 3 (tried upgrading to 3.12 but unfortunately some packages still need to add support for it, so it's better to wait) along with some other related upgrades.

Additionally, it replaces Black, Flake8, and Isort with Ruff, which is a very fast linter and formatter. The translation from previous linting rules to Ruff's are close to but not exactly one-to-one, so we'll likely need to make some tweaks to get the configuration that we want.


backend/Pipfile
```diff
@@ -4,20 +4,21 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-black = "==19.10b0"
+black = "*"
```
rohangpta, line 7: Why is this still here? And why is it = "*"? (Same for flake8)

backend/Pipfile
```diff
@@ -44,7 +45,7 @@ uvloop = {version = "*", sys_platform = "== 'linux'"}
 uvicorn = {extras = ["standard"], version = "*"}
 gunicorn = "*"
 httptools = "*"
-ics = "==0.7"
+ics = "*"
```
rohangpta, line 48: Can you just confirm we're not depending on a version pin here?

backend/Pipfile
```diff
@@ -34,7 +35,7 @@ qrcode = "*"
 python-dateutil = "*"
 psycopg2 = "*"
 django-simple-history = "*"
-channels = "<3"
+channels = "*"
```
rohangpta, line 38: Can you just confirm we're not depending on a version pin here?

backend/Pipfile
```diff
@@ -57,6 +58,8 @@ pandas = "*"
 drf-excel = "*"
 numpy = "*"
 coverage = "*"
+daphne = "*"
```
rohangpta, line 61: Can you explain why we need `daphne` and `tatsu`, and why `tatsu` needs a pin?

.github/workflows/build-and-deploy.yaml
```diff
@@ -12,6 +12,7 @@ jobs:
     with:
       projectName: pennclubs
       path: backend
+      pythonVersion: 3.11-bookworm
```
rohangpta, line 15: What's this `bookworm` stuff?

rohangpta:
Overall this looks good to me! Can you just clarify some of my comments, and tell me about all the local tests you've run with this?

We should also temporarily add in integration tests to just test out a CI run with them enabled. Very possible it could catch some unwanted behaviour or runtime exceptions due to deprecated functions that the tests don't otherwise catch. 

A couple other things:
- You should browse this page to make sure we aren't missing anything major: https://docs.djangoproject.com/en/5.0/internals/deprecation/
- Looks like [Postgres 11 is no longer supported](https://docs.djangoproject.com/en/5.0/releases/4.2/#dropped-support-for-postgresql-11)

Here is what ChatGPT tells me to look out for (some of it is cruft but thought I'd paste here for you to read):

Django has undergone several changes from version 3.2 to 5.0, including new features and some backwards incompatible changes. Here's a summary of the key backwards incompatible changes in each major version from Django 3.2 to 5.0:

### Django 3.2
- Enhanced support for JSONB aggregation in PostgreSQL.
- New database operations for PostgreSQL, like `CreateCollation` and `RemoveCollation`.
- Additional attributes for `Sitemap` in `django.contrib.sitemaps`.
- Improvements in error reporting, file uploads, and forms.
- Updates to generic views and management commands.
- Extensions to the migrations system and model functionalities.

### Django 4.0
- Introduction of Malay language support and changes in the `DeleteView` to use `FormMixin`.
- Modifications in logging, management commands, and models, including the introduction of `QuerySet.contains()`.
- Changes in CSRF protection settings (`CSRF_TRUSTED_ORIGINS`).
- Removal of support for older versions of PostGIS, GDAL, GEOS, PostgreSQL 9.6, and Oracle 12.2 and 18c.

### Django 4.2
- Improved support for locales in the `makemessages` command.
- Enhancements in migrations, models, and requests/responses.
- Updated utilities and validators.
- Dropping support for MariaDB 10.3, MySQL 5.7, and PostgreSQL 11.
- Changes in the database backend API, particularly for MySQL.

### Django 5.0
- Numerous feature additions across various components like asynchronous views, decorators, file storage, and internationalization.
- Enhancements in error reporting, forms, migrations, models, pagination, signals, templates, tests, and validators.
- Removal of support for older versions of GDAL and GEOS.
- Changes in the database backend API, especially concerning the use of database functions as defaults and support for the `DEFAULT` keyword in `INSERT` queries.

Each of these versions introduced specific changes that might require adjustments in your Django projects, especially if you are directly upgrading from Django 3.2 to Django 5.0. It's important to review the detailed release notes and migration guides for each version to ensure a smooth upgrade process. 

For detailed information, you can refer to the release notes for [Django 5.0](https://docs.djangoproject.com/en/5.0/releases/5.0/), [Django 4.2](https://django.readthedocs.io/en/5.0.1/releases/4.2.html), [Django 4.0](https://docs.djangoproject.com/en/5.0/releases/4.0/), and [Django 3.2](https://docs.djangoproject.com/en/5.0/releases/3.2/).

backend/Pipfile
rm03: I might be mistaken, but I think the CI expects black and flake8 to be installed as part of the Django check, so until we migrate shared-actions to use ruff we should keep these here. I removed the version pins as I wasn't sure why they were necessary

backend/Pipfile
rm03: I believe [this](https://github.com/pennlabs/penn-clubs/blob/eec217228c589d91987661811c9c0357ffa502db/backend/clubs/routing.py#L7) was the only part of our code that needed to be updated, but I'll do further checks/tests locally to make sure unpinning the version doesnt cause any issues

backend/Pipfile
rm03: `daphne` is needed because `channels.testing` now requires it but doesn't install it as a subdependency (see [this](https://github.com/django/channels/issues/1942))

I think we should be able to get rid of the `tatsu` pin. I don't remember why I added it (could've been because they dropped support for Python 3.10, but this isn't relevant as we're using 3.11 anyway).

.github/workflows/build-and-deploy.yaml
rm03: `bookworm` is the latest stable Debian release and it seems that Docker has [removed support for `buster`](https://github.com/docker-library/python/pull/822) (which is two versions behind `bookworm`) with later versions of Python. Also if we don't specify the image it defaults to `python3.8-buster` which doesn't make sense now that we're using 3.11

rm03:
For local testing I've only run the unit tests and backend and verified that there were no errors, but I agree that integration tests will definitely help us catch errors that the unit tests won't. I also looked at the Django deprecation page when making the upgrades and didn't notice any glaring incompatible changes, but will take a closer look to make sure I didn't miss anything

backend/Pipfile
rm03: Just updated shared-actions to use ruff, and got rid of black, flake8, and isort

backend/Pipfile
rm03: I remember why I added `tatsu` now, it's because the latest Django base image uses Python 3.10. Will try and use an image with Python 3.11

rm03:
Looks like integration tests are [passing](https://github.com/pennlabs/penn-clubs/actions/runs/7967241979/job/21749866450)!

rohangpta:
LGTM. A few guidelines around the merge
- Please monitor sentry, DB util for 24+h
- Squash and merge, so that revert is easy if needed
- Things are likely to break, so be prepared.

rohangpta (approved):

PR #606, opened 2024-01-24
Stress Testing Throughput for Club Submission Applications

shiva-menta:
Conduct customizable stress tests using `python manage.py stress_test`. Change parameters including number of users, applications per user, number of clubs, etc.

HTTP requests are mocked using `APIRequestFactory`. Tested minimally before and after #603, and saw some performance gains in SQLite, but SQLite's concurrency features should be considered ([thread](https://stackoverflow.com/questions/4060772/sqlite-concurrent-access) explaining how SQLite handles concurrent writes!). Have not tested on Docker Postgres yet.


backend/clubs/management/commands/stress_test.py
```diff
@@ -0,0 +1,173 @@
...
+        self.num_users = 10
+        self.subset_size = 1
+        self.num_questions_per_club = 2
+        self.total_submissions = 1
+        self.club_prefix = "test_club_"
+        self.user_prefix = "test_user_"
+
+        self.uri = "/users/question_response/"
+        self.factory = APIRequestFactory()
+        self.view = UserViewSet.as_view({"post": "question_response"})
```
rohangpta, line 35: This will create objects for real right? So we want to tear down?

backend/clubs/management/commands/stress_test.py
```diff
@@ -0,0 +1,173 @@
...
+                    email=str(i) + "@upenn.edu",
+                    password="test",
+                )
+                for i in range(self.num_users)
+            ]
+        )
+        self.users = list(User.objects.filter(username__startswith=self.user_prefix))
+        print("Finished setting up users.")
+
+    @sync_to_async
```
rohangpta, line 109: This decorator is misleading. All this is doing is taking your function and convering it to an `awaitable`. It unfortunately doesn't make your code "async for free". In fact, with `thread_sensitive = True` (which is the default behaviour), these functions will all run in the main thread, and probably even slower then usual since it goes through the overhead of "asyncifying". Some more information [here](https://docs.djangoproject.com/en/5.0/topics/async/#asgiref.sync.sync_to_async)

It isn't worth changing it back since this is just a test but I thought I should let you know about this behaviour. If you wanted better async, you might want to use httpx or something like that (I think newer versions of Django support Async more natively).

backend/clubs/management/commands/stress_test.py
```diff
@@ -0,0 +1,173 @@
...
+
+        request = self.factory.post(self.uri, data, format="json")
+        request.user = user
+
+        self.view(request)
+        end_time = time.time()
+        return end_time - start_time
+
+    def tearDown(self):
+        Club.objects.filter(code__startswith=self.club_prefix).delete()
```
rohangpta, line 126: Might have stray `ClubApplication`, `ApplicationSubmission`, etc objects which you'll need to tear down

rohangpta:
Can you post your test results here?

rohangpta:
Great start Shiva! Some notes:

- Test overall looks good to me, I like your choice of mocking HTTP request objects and overall the code is cleanly structured.

I think we're ready to take some steps towards productionising this test, some points from me:

- How long does this test take to run locally? Could you add some sleeps between submissions to make it more realistic?
- Can you check whether your asyncifying is doing anything? (you can do this by noting the current `end_time` - `start_time` as a baseline, and comparing that against N * (individual `end_time` - `start_time`) values. I'm curious if `sync_to_async` is actually allowing parallel processing here, although my understanding is not.
- Let's talk through the points [here](https://docs.djangoproject.com/en/5.0/ref/databases/) and [here](https://docs.djangoproject.com/en/5.0/topics/db/optimization/) on Friday: some points on optimisation and DB configuration!
- Before we productionise this, we should enable as many AWS insight metrics as possible so we can see where exactly the load is landing.

backend/clubs/management/commands/stress_test.py
shiva-menta: Yeah the `tearDown()` function handles that, but if it needs to be more explicit I can do that.

backend/clubs/management/commands/stress_test.py
shiva-menta: Yeah I was starting to notice that in the end. I originally was designing it as sending HTTP requests and was going to use `httpx` but then the mocking seemed like a better option. Thanks for sharing!

backend/clubs/management/commands/stress_test.py
shiva-menta: Checking from the models, this should all be handled with `on_delete` behavior? It should form separate "trees" with roots of Clubs and Users.

rohangpta: Sure, this makes sense. Please do double check empirically though; we should definitely avoid excess footprint on the DB, especially if we test in prod multiple times.

shiva-menta:
> Can you post your test results here?

Yeah I'm going to conduct more testing tonight, but I wanted to address the async behaviors first and potentially consider doing HTTP requests instead, but if our goal is to isolate database behavior, mocking makes more sense. Will show benchmarks for SQLite and PostgreSQL!

backend/clubs/management/commands/stress_test.py
shiva-menta: Yeah just double checked, no rows are left over. The only concern I guess could be the autoincrementing index and offsetting that by the number of test instances. Should not be an issue at all, but if we want full isolation, we could create a separate database on the same instance.

rohangpta: I think this should be fine! Thank you for double checking

shiva-menta:
> Great start Shiva! I think we're ready to take some steps towards productionising this test, some notes:
> 
> * Test overall looks good to me, I like your choice of mocking HTTP request objects and overall the code is cleanly structured.
> * How long does this test take to run locally? Could you add some sleeps between submissions to make it more realistic?
> * Can you check whether your asyncifying is doing anything? (you can do this by noting the current `end_time` - `start_time` as a baseline, and comparing that against N * (individual `end_time` - `start_time`) values. I'm curious if `sync_to_async` is actually allowing parallel processing here, although my understanding is not.
> * Let's talk through the points [here](https://docs.djangoproject.com/en/5.0/ref/databases/) and [here](https://docs.djangoproject.com/en/5.0/topics/db/optimization/) on Friday: some points on optimisation and DB configuration!
> * Before we productionise this, we should enable as many AWS insight metrics as possible so we can see where exactly the load is landing.

Took a look, and yeah you're completely correct about `sync_to_async`, I made the incorrect assumption of seeing a difference in thread completion times implied concurrent connections; I believe it just indicated the overhead. When I set `thread_sensitive=False`, I get a bunch of database is locked errors due to the concurrent writes. I can't seem to configure SQLite's timeout setting properly so the database writes aren't blocking until the database is unlocked from the main thread. 

I believe the implementation will be highly dependent on the actual database for now. I'm going to try seeing if PostgreSQL functions as is.

rohangpta:
Yeah sounds good. I don't think `thread_sensitive=False` will end up working here since database adapters in Django require that they are accessed in the same thread that they were created in, but I agree that it's probably best to keep save the bulk of our testing for Postgres

rohangpta:
@shiva-menta did we ever have any results on this? Curious about how things match up since we didn't actually seem to have any DB issues in prod this cycle.

PR #610, opened 2024-02-03
Remove archival of application submissions

rm03:
Previously, instead of deleting users' application submissions, we would instead "archive" them and keep them in our database. With the new upsert change, if someone archives an application and tries to submit a new one, the `unique_together` constraint will cause this to fail. However, it seems like this archival feature isn't being used and users aren't able to view archived submissions anyway, so it's safe to remove it.


rohangpta (approved):
LGTM! Feel free to merge. Maybe add some context in the PR about why we are doing this. But overall looks good

PR #611, opened 2024-02-05
Handle Redis connection errors

aviupadhyayula:
Adds flags to `django_redis` cache. Handles Redis connection errors silently, and falls back to DB when the cache is inaccessible. [1] Also adds socket timeouts on initial connection and subsequent requests. [2] This enables fast failures when the cache is down, instead of waiting indefinitely. 

[1] https://github.com/jazzband/django-redis?tab=readme-ov-file#memcached-exceptions-behavior
[2] https://github.com/jazzband/django-redis?tab=readme-ov-file#socket-timeout

### DIFF
--- a/backend/clubs/utils.py
+++ b/backend/clubs/utils.py
@@ -382,7 +382,7 @@ class CacheManager:
         except Exception as e:
             logger.warning("Switching to fallback cache")
             logger.exception(e)
-            return False
+            return self._cache_fallback.set(key, value, timeout)
 
     def get(self, key, default=None):
         try:
@@ -390,7 +390,7 @@ class CacheManager:
         except Exception as e:
             logger.warning("Switching to fallback cache")
             logger.exception(e)
-            return None
+            return self._cache_fallback.get(key, default)
 
     def delete(self, key):
         try:
@@ -398,7 +398,7 @@ class CacheManager:
         except Exception as e:
             logger.warning("Switching to fallback cache")
             logger.exception(e)
-            return False
+            return self._cache_fallback.delete(key)
 
     def clear(self):
         try:
@@ -407,4 +407,4 @@ class CacheManager:
             logger.warning("Switching to fallback cache")
             logger.exception(e)
         finally:
-            return None
+            self._cache_fallback.clear()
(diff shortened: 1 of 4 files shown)


backend/pennclubs/settings/development.py
```diff
@@ -25,4 +25,9 @@
 CHANNEL_LAYERS = {"default": {"BACKEND": "channels.layers.InMemoryChannelLayer"}}
 
 # Caching settings
-CACHES = {"default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"}}
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        "TIMEOUT": 60 * 60,
```
rohangpta, line 31: Can we revert the timeout change? Rather keep the API consistent

backend/clubs/utils.py
```diff
@@ -355,3 +362,49 @@ def get_django_minified_image(url, **kwargs):
     resp = requests.get(url)
     new_image = resize_image(resp.content, **kwargs)
     return ImageFile(io.BytesIO(new_image), name="image.png")
+
+
+class CacheManager:
```
rohangpta, line 367: Do we need this class? Can we just monkeypatch the methods with exception handling/logging? 

Before we do this, can you also confirm that Django does not suppress cache connection exceptions automatically? I'm understanding that we're doing this because of the problems on the main page last night -- but let's confirm that this was because of the cache failing.

The logic in all 3 methods is mostly the same

aviupadhyayula: I'm fairly confident that a Redis connection refusal is fatal, and isn't suppressed by Django. This is reproducible by pointing the development cache to an incorrect Redis host, and accessing a club's information (`/api/clubs/<club_code>`). Attaching a screenshot below:

Screenshot 2024-02-11 at 4 29 46 PM

backend/pennclubs/settings/development.py
aviupadhyayula: Yup, makes sense.

aviupadhyayula:
Verified this works by pointing Django's development cache to an incorrect Redis host, and querying a route that relies on caching. Unsure how to add a unit test for this -- would love to hear ideas.

rohangpta (approved):
LGTM! Thanks for the updates

PR #618, opened 2024-02-20
Streamline django storages config

aviupadhyayula:
Aligns S3 storage config with pre #617 styling, while using the new API.


rm03 (approved):
lgtm

PR #619, opened 2024-02-20
Add back codecov

rm03:
See title

### DIFF
--- a/.github/workflows/build-and-deploy.yaml
+++ b/.github/workflows/build-and-deploy.yaml
@@ -8,7 +8,8 @@ on: push
 jobs:
   backend-check:
     name: "Backend Check"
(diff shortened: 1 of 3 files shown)


aviupadhyayula (approved):
LGTM. Nit: should we consider passing in a value to `GITHUB_HEAD_REF`?[1] Would be useful to see how coverage would change after a PR.

[1] [https://docs.codecov.com/docs/error-reference#missing-head-commit](https://docs.codecov.com/docs/error-reference#missing-head-commit)

rm03:
> LGTM. Nit: should we consider passing in a value to `GITHUB_HEAD_REF`?[1] Would be useful to see how coverage would change after a PR.[2]
> 
> [1] https://docs.codecov.com/docs/error-reference#missing-head-commit [2] https://app.codecov.io/github/pennlabs/penn-clubs/commit/917c0c9d93cf9f270f5232ee36ff8db8731a33a2

Discussed offline, think this will be handled automatically

PR #623, opened 2024-02-27
Switch to Bun

aviupadhyayula:
[bun.sh](https://bun.sh)


frontend/Dockerfile
```diff
@@ -1,15 +1,24 @@
-FROM node:14.14.0-buster-slim
+FROM node:20.11.1-buster-slim
 
 LABEL maintainer="Penn Labs"
 
+RUN apt-get update \
```
rohangpta, line 5: Perhaps we can consider using the docker image instead? https://hub.docker.com/r/oven/bun/tags

rohangpta:
Good idea!

PR #624, opened 2024-03-05
Send club invites with acceptance emails

Porcupine1:
This PR is meant to automatically send club membership invitations with acceptance emails. Invitations expire after days. This is to prevent accepted candidates from having lifetime access.


backend/clubs/views.py
```diff
@@ -4756,6 +4760,15 @@ def send_emails(self, *args, **kwargs):
                 and email_type == "acceptance"
             ):
                 template = acceptance_template
+
+                if send_invites and not dry_run:
```
aviupadhyayula, line 4764: Do we need to check whether it's a dry run here? Seems like we're already doing it further down.

backend/.vscode/settings.json
```diff
@@ -7,7 +7,7 @@
     "[python]": {
         "editor.formatOnPaste": false,
         "editor.codeActionsOnSave": {
-            "source.fixAll": true
+            "source.fixAll": "explicit"
```
aviupadhyayula, line 10: Looks like VSCode transitioned to using enums in the newer release. We might want to keep this as a boolean for now, until support is completely removed. 

https://stackoverflow.com/questions/77637621/vscode-workspace-settings-change-on-its-own

backend/clubs/views.py
```diff
@@ -4679,7 +4679,7 @@ def retrieve(self, *args, **kwargs):
         return Response(data)
 
     @action(detail=True, methods=["post"])
-    def send_emails(self, *args, **kwargs):
+    def send_emails(self, request, *args, **kwargs):
```
aviupadhyayula, line 4682: Does the request object need to be passed in?

aviupadhyayula (commented):
Great work!

backend/clubs/views.py
Porcupine1: I just didn't want to unnecessarily make invites that would otherwise never be sent

backend/clubs/views.py
Porcupine1: You're right, I don't have to. I can use self.request instead without passing in request

backend/clubs/views.py
aviupadhyayula: Good catch. I didn't see that we were committing invites to the DB.

On that note: would it be better to create invites in bulk? After all, we know exactly which invites we're creating even before we go to the DB. Writing everything in one motion might be more efficient. 

This might be a good reference: https://docs.djangoproject.com/en/5.0/ref/models/querysets/#bulk-create

Porcupine1: Noted. Made changes

aviupadhyayula: Great. Also, I think after these changes, it's safe to remove the check for whether we're in a dry run or not.

rm03: +1, you can remove the dry run check here

rm03 (commented):
Great work! Going forward I think we're going to try and write unit tests for new features - do you think you'd be able to write a test for this?

Porcupine1:
> Great work! Going forward I think we're going to try and write unit tests for new features - do you think you'd be able to write a test for this?

Yes, looking into that.

rm03:
Also, you should go ahead and update the frontend to reflect this feature – will be a very small change

backend/tests/clubs/test_commands.py
```diff
@@ -651,3 +651,46 @@ def test_wrong_arguments(self):
...
+        self.user2 = get_user_model().objects.create_user(
+            "tjefferson", "test-application-notif@example.com", "test"
+        )
+
+        # Club
+        self.club1 = Club.objects.create(
+            code="one", name="Club One", active=True, email="test@example.com"
+        )
+
+        # Expired MembershipInvite
```
aviupadhyayula, line 671: nit: comments could be more descriptive. Something like "Create expired membership invite" is generally better practice.

backend/tests/clubs/test_commands.py
```diff
@@ -651,3 +651,46 @@ def test_wrong_arguments(self):
...
+        # Active MembershipInvite
+        self.active_invite = MembershipInvite.objects.create(
+            email=self.user1.email,
+            club=self.club1,
+            creator=self.user1,
+            active=True,
+            expires_at=timezone.now() + datetime.timedelta(days=1),
+        )
+
+    def test_expire_membership_invites(self):
```
aviupadhyayula, line 689: nit: would be great to have a comment here explaining what expected behavior of the command is. It's self-explanatory for this test (and we haven't done a great job of this in the past), but it'd be nice to set the standard going forwards.

backend/tests/clubs/test_views.py
```diff
@@ -1620,6 +1622,95 @@ def test_club_invite_email_resend(self):
...
+        for user in applicants:
+            submission = ApplicationSubmission.objects.create(
+                user=user,
+                application=application,
+                committee=None,
+                status=statuses[i],
+                reason="a reason",
+            )
+            i += 1
+            submission.save()
```
aviupadhyayula, line 1655: Could we do this with a bulk operation? More performant tests are always a plus.

aviupadhyayula, line 1655: In this case, I think something like `for user, status in zip(applicants, statuses)` is more Pythonic.

backend/tests/clubs/test_views.py
```diff
@@ -1620,6 +1622,95 @@ def test_club_invite_email_resend(self):
...
+
+        i = 0
+        for user in applicants:
+            # check application submissions were created
+            self.assertTrue(
+                ApplicationSubmission.objects.filter(
+                    application=application, user=user, status=statuses[i]
+                ).exists()
+            )
+            i += 1
```
aviupadhyayula, line 1687: It's good to be thorough, but I think testing for ancillary objects (i.e. the club application, application submissions) is out of scope for this test. Ideally, a unit test should test a single feature/component, and do it as efficiently as possible.

Since the goal of this test is to check for the creation of membership invites, these checks are unnecessary. In 99.99% of cases, we can expect that calling `ApplicationSubmission.objects.create()` will work as intended.

backend/tests/clubs/test_views.py
```diff
@@ -1620,6 +1622,95 @@ def test_club_invite_email_resend(self):
...
+
+        # Check that invitations were created.
+        # Email test are done in test_send_invite_email
```
aviupadhyayula, line 1690: nit: can remove this line of the comment.

backend/tests/clubs/test_views.py
```diff
@@ -1620,6 +1622,95 @@ def test_club_invite_email_resend(self):
...
+            MembershipInvite.objects.filter(
+                club=application.club, email=self.user1.email
+            ).exists()
+        )
+        self.assertTrue(
+            MembershipInvite.objects.filter(
+                club=application.club, email=self.user2.email
+            ).exists()
+        )
+
```
aviupadhyayula, line 1701: Maybe we can check whether the created `MembershipInvite`s have the intended expiry time? Let me know if that sounds reasonable.

backend/tests/clubs/test_views.py
```diff
@@ -1620,6 +1622,95 @@ def test_club_invite_email_resend(self):
...
+            club=self.club1,
+            application_start_time=now,
+            application_end_time=now + timezone.timedelta(days=1),
+            result_release_time=now + timezone.timedelta(days=2),
+            is_wharton_council=False,
+        )
+        application.save()
+
+        applicants = [self.user1, self.user2, self.user3, self.user4]
+        # 4 = accepted, 2 = rejected, 1 = pending
```
aviupadhyayula, line 1641: nit: line-specific comments can generally go on the same line as the code they're referring to. I usually hit tab, and then leave the comment there.

aviupadhyayula (commented):
Great work, our testing infra has been lacking for a while. This PR's the start of good things 🚀

backend/tests/clubs/test_views.py
aviupadhyayula: Following up on this: I'm pretty sure you can use the enums defined in the `ApplicationSubmission` model, instead of their integer values. Something like `ApplicationSubmission.PENDING`, `ApplicationSubmission.REJECTED_AFTER_WRITTEN`, etc.

backend/tests/clubs/test_views.py
aviupadhyayula, line 1687: nit: already commented above, but something like `for applicant, status in zip(applicants, statuses` is more Pythonic.

rm03 (commented):
Awesome work! Avi's done a great job reviewing this PR, once his comments are addressed everything should be good to go 🚀

backend/tests/clubs/test_views.py
```diff
@@ -1620,6 +1622,95 @@ def test_club_invite_email_resend(self):
         )
         self.assertIn(resp.status_code, [200, 201], resp.content)
 
+    def test_send_invite_with_acceptance_email(self):
+        self.client.login(username=self.user1.username, password="test")
+
+        # Create a club application
+        now = timezone.now()
+        application = ClubApplication.objects.create(
+            name="Send Acceptance Invivites",
```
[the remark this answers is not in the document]
aviupadhyayula: @julianweng's been a bad influence 😔

backend/tests/clubs/test_views.py
Porcupine1: Sure, I don't see why not

rm03 (approved):

backend/tests/clubs/test_views.py
```diff
@@ -1620,6 +1622,99 @@ def test_club_invite_email_resend(self):
...
+        # Compare the expiry dates excluding microseconds
+        self.assertEqual(
+            invite1.expires_at.replace(microsecond=0),
+            expiry_date.replace(microsecond=0),
+        )
+
+        self.assertEqual(
+            invite2.expires_at.replace(microsecond=0),
+            expiry_date.replace(microsecond=0),
+        )
```
aviupadhyayula, line 1699: This should check for a range, instead of exact equality. (Something like "check if the invite expires within 5 minutes plus/minus of the expiry date".) Unit tests should be deterministic. We don't want this failing if the invites get sent out at 0.99 seconds, and the assert statement runs at 1.01.

aviupadhyayula: This ended up causing test failures. Flagged in #662.

PR #626, opened 2024-03-06
Fix admin page access on frontend.

julianweng:
Add frontend check (on top of backend) for relevant permissions to access /admin.


esinx (approved):
LGTM

PR #627, opened 2024-03-10
Optimize assorted Django operations

aviupadhyayula:
Meant to generally optimize our usage of the Django ORM, with a focus on taking full advantage of major upgrades in #605. 

Changelog: 

- Use subqueries to update membership & favorite counts in `update_club_counts`. Shows a ~70% speedup in local benchmarks (n=2000 memberships/favorites). 
- Use bulk updates when updating decision reasons for application submissions.
- Use bulk create when creating activities fairs for registered clubs. 
    - Concerned with how `bulk_create` plays with `AutoField` primary keys (like we have for the `Events` model). But, the documentation mentions that pSQL should be fine.[1] 
    - I'm assuming the order of the returned `Events` objects doesn't matter. Would appreciate a second set of eyes.
 - Use `qs.exists()` to check existence instead of `qs.first() is not None` whenever possible. 
 - Avoid fetching objects from DB into memory whenever possible.
 
[1] https://docs.djangoproject.com/en/5.0/ref/models/querysets/#bulk-create

### DIFF
--- a/backend/clubs/management/commands/update_club_counts.py
+++ b/backend/clubs/management/commands/update_club_counts.py
@@ -1,5 +1,5 @@
 from django.core.management.base import BaseCommand
-from django.db.models import Count, Q
+from django.db.models import Count, OuterRef, Q, Subquery
 
 from clubs.models import Club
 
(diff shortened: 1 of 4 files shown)


backend/clubs/management/commands/update_club_counts.py
rohangpta, line 1: Is this intended to be run as a cronjob?

aviupadhyayula: I think so, AFAI this was part of Julian and Shiva's work. @rm03 might have more clarity here

rohangpta: Yep, looked at the history. We'll need to add a cronjob to `k8s/main.ts`

rm03: https://github.com/pennlabs/penn-clubs/blob/25bb991e3d0695a39203a15c34d8bd4017182aeb/k8s/main.ts#L153

I remember Shiva added it in his PR, but looks like the change was reverted at some point

backend/clubs/views.py
```diff
@@ -4899,36 +4902,36 @@ def update(self, *args, **kwargs):
...
-        )
-        str_start_date = self.request.data.get("start_date").replace("T", " ")
-        str_end_date = self.request.data.get("end_date").replace("T", " ")
-        str_release_date = self.request.data.get("release_date").replace("T", " ")
+        cycle = self.get_object()
+        applications = ClubApplication.objects.filter(application_cycle=cycle)
+
+        new_start = self.request.data.get("start_date").replace("T", " ")
+        new_end = self.request.data.get("end_date").replace("T", " ")
+        new_release = self.request.data.get("release_date").replace("T", " ")
```
aviupadhyayula, line 4910: @julianweng you show up as the last author for these lines -- are these supposed to replace with spaces, or empty strings? Confused how the lines below (`if new_start`) ever evaluates to false if we're replacing with a non-empty string.

aviupadhyayula:
This is stale and most of these changes have been obviated by platform and dependency upgrades. Closing.

PR #637, opened 2024-03-17
Display username in profile page if name is empty

aviupadhyayula:
The logged-in user's display name is empty in some cases, such as when you're a student taking a leave of absence. This change will display the PennKey if the user's name is an empty string.
Screenshot 2024-03-17 at 6 57 46 PM


aviupadhyayula:
:) 
Screenshot 2024-03-17 at 8 05 39 PM

PR #638, opened 2024-03-18
Move submissions from /apply to user profile

owlester12:


aviupadhyayula (changes requested):
Looks like your branch is behind `master` by a couple of commits, specifically #616. I'm not able to get this working on my machine with the newest frontend versions. I think you may have forgotten to `git pull` on `master` before branching off?

You can solve this by merging `master` into this branch. There are a bunch of guides out there on this. Feel free to DM if you have questions!

aviupadhyayula:
Probably should've merged instead of rebase, RIP. The diff tab is messed up because of [GH's 2-dot vs 3-dot compare](https://docs.github.com/articles/about-comparing-branches-in-pull-requests). Here's [the diff with the current head of `master`.](https://github.com/pennlabs/penn-clubs/compare/master..owen/add-application)

frontend/pages/settings.tsx
julianweng, line 1: I would appreciate if you can still use the existing SettingsProps type, e.g. edit it to include submissions and type the props for the Settings page again.

julianweng (changes requested):
Thanks for submitting, left some nits. Would appreciate if you can fix before Friday and we can merge!

frontend/pages/settings.tsx
owlester12: I'm getting a lot of mismatch type errors when I do that– I believe because of getInitialProps. We can resolve during the meeting Friday

frontend/package.json
```diff
@@ -51,6 +51,7 @@
     "react-toastify": "^10.0.4",
     "react-tooltip": "^5.26.3",
     "react-vis": "^1.12.1",
+    "ruff": "^1.5.4",
```
aviupadhyayula, line 54: @owlester12 `ruff` is a Python linter and shouldn't be in `package.json`. Can you remove this please?

cc @rm03

frontend/yarn.lock
```diff
@@ -5547,6 +5547,11 @@ electron-to-chromium@^1.4.668:
   resolved "https://registry.yarnpkg.com/electron-to-chromium/-/electron-to-chromium-1.4.676.tgz#95179dd62f43926c636ca06c555a1da8754073ff"
   integrity sha512-uHt4FB8SeYdhcOsj2ix/C39S7sPSNFJpzShjxGOm1KdF4MHyGqGi389+T5cErsodsijojXilYaHIKKqJfqh7uQ==
 
+emitify@^3.0.1:
+  version "3.1.0"
+  resolved "https://registry.yarnpkg.com/emitify/-/emitify-3.1.0.tgz#fa5a08011b0f1f64e440aa2e7c2713c55f8a5568"
+  integrity sha512-pucUCyaeC8TJoRXniRiNb81Odoqlf2lMaFARA8tci2PaORFIeXBoIQHEzuA3dxPsU1l7AVb7xmFra4iJwVU8wg==
+
```
aviupadhyayula, line 5554: Is there a reason this is bumped?

cc @julianweng

frontend/pages/settings.tsx
```diff
@@ -97,4 +106,19 @@ const Settings = ({ userInfo, authenticated }: SettingsProps) => {
   )
 }
 
+type BulkResp = {
+  submissions: Array<ApplicationSubmission>
+}
+
+Settings.getInitialProps = async (ctx: NextPageContext) => {
+  const data: BulkResp = (await doBulkLookup(
+    ['submissions', '/submissions/?format=json'],
```
aviupadhyayula, line 115: cc @rm03

rm03: Looks like the `?format=json` is not needed as `doBulkLookup()` automatically appends this to the end of any API call. Also, why are we calling `doBulkLookup()` here instead of `doApiRequest()`?

PR #639, opened 2024-03-18
Set (and enforce) TTL on programmatically generated membership invites

aviupadhyayula:
Implements membership invite TTL and creates cronjob to expire stale invites [as per our discussion](https://penn-labs.slack.com/archives/C06NT2MQK96/p1710539985957989). I set the time to expiry at 5 days. Let's tweak this as we receive feedback. 

Because of the cronjob's scheduling, there may be cases where invites are past their TTL but haven't been marked as expired. This shouldn't be a big deal (and in the worst case, is functionally a 24 hour extension). We can solve this by adding checks to membership acceptance, but I don't think it's necessary.


backend/clubs/models.py
```diff
@@ -1305,6 +1305,7 @@ class MembershipInvite(models.Model):
 
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
+    expires_at = models.DateTimeField(null=True, blank=True)
```
aviupadhyayula, line 1308: Pretty sure this should be `null=True, blank=True`. @rm03 thoughts?

Porcupine1 (commented):
lgtm

PR #643, opened 2024-03-23
Notify club owners upon deletion

aviupadhyayula:
We've seen a couple cases where a club is deleted but its owners aren't notified. This PR adds support for sending emails to club owners (and the executors of deletions) upon deletion requests. 

Not sure whether the email template looks fine (and I'm not sure how to check, either). Also unsure whether this should be a method in `models`.


julianweng (approved):
LGTM

rohangpta (approved):
LGTM, discussed design over slack

Issue #630, opened 2024-03-16
Get capture context and feed to frontend


rohangpta:
Use a modified version of this tool: https://github.com/CyberSource/cybersource-rest-samples-python/blob/master/samples/UnifiedCheckout/generate-unified-checkout

To generate the unified checkout capture context, which will be used by the frontend to create the `<iframe>` object. 

Return this on initiating checkout

PR #645, opened 2024-04-12
Capture context generation + local dev setup instructions

rm03:
Addresses #630 and adds instructions on the new local dev setup for ticketing (thanks @esinx for the help), since the Cybersource API requires that the frontend run with HTTPS to generate the capture context.


backend/pennclubs/settings/development.py
```diff
@@ -32,3 +32,13 @@
 
 # Caching settings
 CACHES = {"default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"}}
+
+# Cybersource settings
+CYBERSOURCE_CONFIG = {
```
esinx, line 37: What is this for?

rm03: Test config for the Cybersource API, those credentials are taken directly from their API docs/examples

rm03:
Do we want to move this functionality inside the checkout view or keep it as a separate view? I think the former might make more sense

README.md
```diff
@@ -80,3 +80,13 @@ You will need to set the following environment variables on the frontend:
...
+
+#### Ticketing
+
+To test ticketing locally, you will need to [install](https://github.com/FiloSottile/mkcert?tab=readme-ov-file#installation) `mkcert`, enter the `frontend` directory, and run the following commands:
+
+- `$ mkcert -install`
+- `$ mkcert localhost 127.0.0.1 ::1`
+- `$ export DOMAIN=https://localhost NODE_TLS_REJECT_UNAUTHORIZED=0`
+
+Then, after the frontend is running, run `docker-compose up` **in a new terminal window** and access the application at [https://localhost](https://localhost).
```
aviupadhyayula, line 92: Thanks for the writeup. Might be a dumb question, but what happens if someone doesn't follow these instructions? I'm guessing the Cybersource ticketing routes will still work, but the capture context it generates won't be displayable on the frontend.

Would be ideal if there was a way to do dev work on non-ticketing projects without setting up the HTTPS/nginx bridge.

rm03: Yep, the iframe will not be rendered on the frontend without HTTPS but the backend routes will still work. The dev workflow for non-ticketing tasks should be unaffected by this change

backend/clubs/models.py
```diff
@@ -1840,6 +1840,7 @@ class Ticket(models.Model):
     )
     holding_expiration = models.DateTimeField(null=True, blank=True)
     carts = models.ManyToManyField(Cart, related_name="tickets", blank=True)
+    price = models.DecimalField(max_digits=5, decimal_places=2)
```
rohangpta, line 1843: Lovely to see this finally added.

backend/clubs/views.py
```diff
@@ -4621,42 +4624,47 @@ def cart(self, request, *args, **kwargs):
...
-                        event=ticket.event,
-                        type=ticket.type,
-                        owner__isnull=True,
-                        holder__isnull=True,
-                    )
-                    .first()
+        # Replace in-cart tickets that have been bought/held
+        tickets_to_replace = cart.tickets.filter(
+            owner__isnull=False
+        ) | cart.tickets.filter(holder__isnull=False, holder__ne=self.request.user)
```
rohangpta, line 4630: @aviupadhyayula if you want or behaviour within the ORM...you might be better served with t[`Q` objects](https://docs.djangoproject.com/en/5.0/topics/db/queries/#complex-lookups-with-q-objects)

rohangpta:
So are we getting rid of Carts or not? IIRC last discussion was to go straight to checkout with selected tickets? Is that frontend only and are we keeping the backend abstraction?

frontend/nginx.conf
rohangpta, line 1: What's this for?

rohangpta:
Have you been able to test this at all?

backend/clubs/views.py
aviupadhyayula: Good catch. Should be fixed.

frontend/nginx.conf
rm03: Was using it to handle https traffic, but it's a bit overkill for our purposes. Going to replace with [`local-ssl-proxy`](https://github.com/cameronhunter/local-ssl-proxy)

rm03:
> So are we getting rid of Carts or not? IIRC last discussion was to go straight to checkout with selected tickets? Is that frontend only and are we keeping the backend abstraction?

I think we'll still keep the cart backend since we want to keep track of the tickets the user is attempting to checkout, but no cart page on the frontend

> Have you been able to test this at all?

I tested locally and was able to render the payment iframe using the generated capture context

PR #653, opened 2024-04-14
Add ability to transfer ownership of tickets

rm03:
Addresses #646. Also adds a serializer for ticket creation to simplify validation logic.


backend/clubs/views.py
```diff
@@ -4986,6 +4990,73 @@ def qr(self, request, *args, **kwargs):
         qr_image.save(response, "PNG")
         return response
 
+    @action(detail=True, methods=["post"])
```
rohangpta, line 4993: @transaction.atomic perhaps?

backend/clubs/views.py
```diff
@@ -4986,6 +4990,73 @@ def qr(self, request, *args, **kwargs):
...
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
+        if self.request.user == receiver:
+            return Response(
+                {"detail": "You cannot transfer a ticket to yourself"},
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
+        ticket.owner = receiver
```
rohangpta, line 5052: What do you think about sending email confirmation to both? Perhaps a classmethod on `TicketTransfer`

backend/clubs/models.py
```diff
@@ -1895,6 +1896,42 @@ def send_confirmation_email(self):
...
+
+    ticket = models.ForeignKey(
+        Ticket, related_name="transfers", on_delete=models.CASCADE
+    )
+    sender = models.ForeignKey(
+        get_user_model(), related_name="sent_transfers", on_delete=models.CASCADE
+    )
+    receiver = models.ForeignKey(
+        get_user_model(), related_name="received_transfers", on_delete=models.CASCADE
+    )
```
aviupadhyayula, line 1912: Are we sure `CASCADE` is the intended behavior on deletion? If a ticket's been transferred, it's been bought & exchanged. Allow for its deletion seems a little unsafe. Something like `RESTRICT` or `PROTECT` might work better. 

I guess the same would go for transaction records. Let me know what you think.

backend/clubs/views.py
```diff
@@ -4986,6 +4990,75 @@ def qr(self, request, *args, **kwargs):
...
+                           properties:
+                                detail:
+                                    type: string
+        ---
+        """
+        receiver = get_object_or_404(
+            get_user_model(), username=request.data.get("username")
+        )
+
+        ticket = self.get_object()
```
aviupadhyayula, line 5040: Do we check if the ticket is owned by the sender anywhere?

backend/templates/emails/ticket_transfer.html
```diff
@@ -0,0 +1,27 @@
...
+event_name:
+    type: string
+type:
+    type: string
+-->
+
+{% extends 'emails/base.html' %}
+
+{% block content %}
+<h2>Thanks for using Penn Clubs!</h2>
```
aviupadhyayula, line 17: nit: do we have a thanks message in any of our other email templates? Haven't seen this before, just curious.

backend/templates/emails/ticket_transfer.html
```diff
@@ -0,0 +1,27 @@
...
+
+<p style="font-size: 1.2em">
+    {{ sender_first_name }}, this is confirmation that you have transferred a ticket to {{ receiver_first_name }} ({{ receiver_username }}) for {{ event_name }} with ticket type {{ type }}.
+</p>
```
aviupadhyayula, line 21: nit: would be nice if we could throw in a link to the ticket as an href.

backend/clubs/models.py
```diff
@@ -1846,6 +1846,7 @@ class Ticket(models.Model):
     holding_expiration = models.DateTimeField(null=True, blank=True)
     carts = models.ManyToManyField(Cart, related_name="tickets", blank=True)
     price = models.DecimalField(max_digits=5, decimal_places=2)
+    transferrable = models.BooleanField(default=True)
```
aviupadhyayula, line 1849: Let's add an option to make tickets non-transferrable at event/ticket creation. Would be nice to display this as a frontend toggle that admins can opt into.

aviupadhyayula (commented):
Left some comments, great work 🚀

backend/clubs/models.py
rohangpta: Agree CASCADE is too strong,..transaction records set null on delete

aviupadhyayula: Was actually talking to @rm03 about this -- what if we set it to `PROTECT` for transaction and exchange records? This would disallow ticket deletion without first deleting the associated receipts. 

Existence of a `TicketTransactionRecord` or `TicketTransfer` implies that the ticket has been bought & is in circulation. As long as that is true, IMO the referenced ticket shouldn't be deletable.

backend/clubs/models.py
```diff
@@ -1895,6 +1896,42 @@ def send_confirmation_email(self):
             )
 
 
+class TicketTransfer(models.Model):
```
aviupadhyayula, line 1899: For consistency's sake, can we call this `TicketTransferRecord`? Goes well with `TicketTransactionRecord`.

backend/clubs/models.py
rohangpta: Didn't know this existed: yes, even better. Nice find.

What is the behaviour if someone tries to delete a ticket that's protected by `PROTECT`? An error?

rm03: Reason I used `CASCADE` is that a transfer record is essentially useless if we are missing one of the users or the ticket. Wasn't aware `PROTECT` existed, definitely the best choice here

backend/clubs/views.py
rm03: This is checked in the queryset, and I tested that we return a 404 if the sender tries to transfer a ticket that they don't own

backend/clubs/models.py
aviupadhyayula: Makes sense. I think we can get away with setting it to `NULL` for the recipient or the sender, but the ticket reference should be `PROTECT`.

backend/templates/emails/ticket_transfer.html
rm03: It's in the ticket confirmation email which I copied, doesn't seem as appropriate for ticket transfers though...will remove

backend/templates/emails/ticket_transfer.html
rm03: We also send the new owner of the ticket a confirmation email, which should contain a QR code that has a link to the ticket

aviupadhyayula: Checks out, I'll leave this up to you.

backend/templates/emails/ticket_transfer.html
```diff
@@ -0,0 +1,27 @@
...
+    type: string
+-->
+
+{% extends 'emails/base.html' %}
+
+{% block content %}
+<h2>Thanks for using Penn Clubs!</h2>
+
+<p style="font-size: 1.2em">
+    {{ sender_first_name }}, this is confirmation that you have transferred a ticket to {{ receiver_first_name }} ({{ receiver_username }}) for {{ event_name }} with ticket type {{ type }}.
```
aviupadhyayula, line 20: nit: can we bold the dynamic fields?

backend/clubs/views.py
aviupadhyayula: Great, thanks. I think it'd be useful to add the check, just to make it intentionally clear in the code (useful for reference & future dev). But this is trivial and up to you.

backend/clubs/views.py
```diff
@@ -5056,6 +5058,76 @@ def qr(self, request, *args, **kwargs):
...
+            return Response(
+                {"detail": "You cannot transfer a ticket to yourself"},
+                status=status.HTTP_403_FORBIDDEN,
+            )
+
+        ticket.owner = receiver
+        ticket.save()
+        TicketTransferRecord.objects.create(
+            ticket=ticket, sender=self.request.user, receiver=receiver
+        ).send_confirmation_email()
```
rohangpta, line 5126: At this point, you'll probably need to invalidate the sender's QR, currently built using

`url = f"https://{settings.DOMAIN}/api/tickets/{self.id}"`

This URL was put in as a placeholder and is not actually suitable. We want a URL that is unique to the user/locked behind auth, so that others can't access it. We could use a [hash of ticket.UUID + ticket.owner](https://en.wikipedia.org/wiki/Object-capability_model), since locking behind auth means locking behind Penn access. If we change the way we do the URL, we would need to write another endpoint that validates it.

@julianweng could we co-ordinate with Anthony on the quickest way to get ticket QR code validation into iOS? Or do we want to use a different interface?

backend/clubs/views.py
rohangpta: +1 could you add a comment here? Not immediately obvious

backend/clubs/models.py
```diff
@@ -1896,6 +1877,66 @@ def send_confirmation_email(self):
...
+        null=True,
+        blank=True,
+    )
+    receiver = models.ForeignKey(
+        get_user_model(),
+        related_name="received_transfers",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+    )
```
aviupadhyayula, line 1917: Should these be nullable?

backend/clubs/views.py
```diff
@@ -2616,39 +2620,33 @@ def create_tickets(self, request, *args, **kwargs):
...
+                            items:
+                                type: object
+                                additionalProperties:
+                                    type: array
+                                    items:
+                                        type: string
         ---
         """
         event = self.get_object()
+        quantities = request.data.get("quantities", [])
```
aviupadhyayula, line 2633: should we throw an error if `quantities` is empty?

aviupadhyayula (commented):
LGTM. After cleaning up the frontend lint & addressing the QR stuff it should be good to go  🚀

backend/clubs/models.py
rohangpta: If you want to use `SET_NULL`...

aviupadhyayula: lmao my bad. Was hoping there was a way to disallow null references on creation while using SET_NULL, but idk if there is.

backend/clubs/views.py
rm03: After some discussion I think we're going to put this in a separate PR. I've made the updates locally but more changes will likely be required until we finalize details about QR code scanning.

backend/clubs/views.py
```diff
@@ -2611,6 +2612,8 @@ def create_tickets(self, request, *args, **kwargs):
                                         group_discount:
                                             type: number
                                             required: false
+                                        transferable:
```
aviupadhyayula, line 2615: Are you using this anywhere in `create_tickets`?

backend/clubs/models.py
```diff
@@ -1905,6 +1886,66 @@ def send_confirmation_email(self):
...
 
 
+class TicketTransactionRecord(models.Model):
+    """
+    Represents an instance of a transaction record for an ticket, used for bookkeeping
+    """
+
+    ticket = models.ForeignKey(
+        Ticket, related_name="transaction_records", on_delete=models.PROTECT
+    )
```
aviupadhyayula, line 1896: nit: would be nice to add tests for this

backend/clubs/models.py
```diff
@@ -1905,6 +1886,66 @@ def send_confirmation_email(self):
...
+    buyer_email = models.EmailField(blank=True, null=True)
+
+
+class TicketTransferRecord(models.Model):
+    """
+    Represents a transfer of ticket ownership, used for bookkeeping
+    """
+
+    ticket = models.ForeignKey(
+        Ticket, related_name="transfer_records", on_delete=models.PROTECT
```
aviupadhyayula, line 1911: nit: would be nice to add tests for this

aviupadhyayula (commented):
Great work, thanks Rohan. Left a few comments.

backend/clubs/views.py
rm03: Forgot to add back after the merge, thanks for the catch

backend/templates/emails/ticket_confirmation.html
```diff
@@ -39,7 +39,7 @@ <h2>Thanks for using Penn Clubs!</h2>
 
 <img id="now" style="max-width: 60%; width: auto;" alt="" src="data:image/png;base64,{{ qr}}" />
 
-<p style="font-size: 1.2em"> Note: all tickets issued by us are <b>non-refundable</b> and <b>non-transferable</b>. </p>
+<p style="font-size: 1.2em"> Note: all tickets issued by us are <b>non-refundable</b>. </p>
```
aviupadhyayula, line 42: nit: extra space here

backend/tests/clubs/test_ticketing.py
```diff
@@ -1040,3 +1048,84 @@ def test_complete_checkout_cybersource_fails(self):
...
+    def test_delete_ticket_after_transfer(self):
+        ticket = self.tickets1[0]
+        ticket.owner = self.user2
+        ticket.save()
+        TicketTransferRecord.objects.create(
+            ticket=ticket, sender=self.user1, receiver=self.user2
+        )
+
+        with self.assertRaises(ProtectedError):
+            ticket.delete()
```
aviupadhyayula, line 1131: nit: these shouldn't be in the viewset's test case

aviupadhyayula (approved):
LGTM, feel free to merge. 🚀

PR #654, opened 2024-04-14
Set and enforce order limit on ticket purchases

aviupadhyayula:
Allows for `order_limit` to be set at tickets creation. Defaults to 10 if not specified. Enforces this order limit when the user adds tickets to their cart.


aviupadhyayula:
Messed up the branch naming, sorry for the confusion.

backend/clubs/views.py
```diff
@@ -4780,6 +4788,30 @@ def initiate_checkout(self, request, *args, **kwargs):
...
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        # Check if number of tickets is within event's order limit
+        event = tickets.first().event if tickets.exists() else None
+        if (
+            event.ticket_order_limit is not None
+            and cart.tickets.count() > event.ticket_order_limit
+        ):
```
aviupadhyayula, line 4805: This won't work if there are multiple events represented in a user's cart. IIRC we last discussed passing in the event ID to `initiate_checkout`, and only checking out tickets associated with that event. This allows us to not display a "global cart" in the frontend, but maintain that model in the backend. 

If this sounds reasonable, I can ship that in this PR.

rohangpta: Maybe you can limit in `add_to_cart` instead? That method is keyed on event already..

aviupadhyayula: Will do, but we'll still have to deal with this at some point, right? When users initiate a checkout, they should only be checking out the tickets that are tied to the event they're currently looking at. As it stands, we'd check out every ticket in their cart.

rohangpta (changes requested):
Seems more appropriate to limit in `add_to_cart` instead. That method is on the `EventViewSet`.

backend/clubs/views.py
rohangpta: Moved comment to a change request.

> When users initiate a checkout, they should only be checking out the tickets that are tied to the event they're currently looking at.

Checkouts are now associated with carts (we decided we'd have to stick with carts for a few technical reasons of keeping consistent semantics on frontend/backend), so I don't believe we're limited by this.

aviupadhyayula: Makes sense. Fixed in [3583db1](https://github.com/pennlabs/penn-clubs/pull/654/commits/3583db142af729c4e25fcb18dab0e65593b1e2e7).

rohangpta: Do we want to validate that the user behaved correctly in `initiate_checkout` or perhaps not let them behave incorrectly by adding this validation in `add_to_cart`. We can drop the validation in `initiate_checkout` in that case.

ie: don't let users add more tix to cart than specified by `event.ticket_order_limit`?

aviupadhyayula: Whoops, looks like the commit didn't have everything staged. Agree with moving this validation logic to `add_to_cart`.

Changelog should be up-to-date now. Let me know if it's what you envisioned.

backend/clubs/views.py
```diff
@@ -4780,6 +4800,15 @@ def initiate_checkout(self, request, *args, **kwargs):
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
+        if tickets.count() == 0:
```
rohangpta, line 4803: Nit: could you move this above the check on 4793? Makes a bit more sense

backend/clubs/views.py
```diff
@@ -2599,6 +2615,10 @@ def create_tickets(self, request, *args, **kwargs):
 
         Ticket.objects.bulk_create(tickets)
 
+        order_limit = request.data.get("order_limit", 10)
```
rohangpta, line 2618: I'm looking at this logic and I'm a bit worried about our choice to delete tickets after creating ticket offerings:
We should figure out a way to freeze ticket counts, so that bought tickets can't be deleted. Maybe something like a "ticket drop time" as a time that the tickets will be frozen and not editable?

Not strictly related to your PR but just something that came to mind..

rohangpta (approved):
Thanks Avi, looks great!

backend/clubs/views.py
aviupadhyayula: Agreed, good catch. Resolving this for now -- will fix in another branch.

PR #657, opened 2024-04-15
Check ticket holds before completing checkout

aviupadhyayula:
Resolves edge case where the user initiates checkout, but waits 15 mins (i.e. longer than the hold, but shorter than the capture context expiration).

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -4911,6 +4911,19 @@ class TicketViewSet(viewsets.ModelViewSet):
             Cart.objects.prefetch_related("tickets"), owner=self.request.user
         )
 
+        # Guard against holds expiring before the capture context
+        tickets = cart.tickets.select_for_update(skip_locked=True).filter(
+            holder=self.request.user, owner__isnull=True
+        )
+        if tickets.count() != cart.tickets.count():
+            return Response(
+                {
+                    "success": False,
+                    "detail": "Cart is stale, invoke /api/tickets/cart to refresh",
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
         try:
             _, http_status, transaction_data = TransientTokenDataApi(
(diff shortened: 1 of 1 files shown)


backend/clubs/views.py
```diff
@@ -4911,6 +4911,19 @@ def complete_checkout(self, request, *args, **kwargs):
             Cart.objects.prefetch_related("tickets"), owner=self.request.user
         )
 
+        # Guard against holds expiring before the capture context
+        tickets = cart.tickets.select_for_update(skip_locked=True).filter(
+            holder=self.request.user, owner__isnull=True
+        )
```
aviupadhyayula, line 4917: Not sure if the `select_for_update` is needed here. I'm emulating the check we do in `initiate_checkout` (which has the lock). Is obtaining a lock necessary in both cases?

rohangpta: nope: we don't actually update so we're good :)

rohangpta (approved):

rohangpta:
approved pending removing the lock

backend/clubs/views.py
```diff
@@ -4911,6 +4911,17 @@ def complete_checkout(self, request, *args, **kwargs):
             Cart.objects.prefetch_related("tickets"), owner=self.request.user
         )
 
+        # Guard against holds expiring before the capture context
+        tickets = cart.tickets.filter(holder=self.request.user, owner__isnull=True)
+        if tickets.count() != cart.tickets.count():
```
rohangpta, line 4916: Note: this assertion checks that all the tickets in the cart are held by the user, so someone is not allowed to add more tix to cart without checking out (I'm ok with this)

rohangpta:
Can you also add the update_holds decorator to this function? I omitted it explicitly bc we weren't checking holds but now that we are, may as well update

PR #658, opened 2024-04-15
Add support for free tickets

### LINKED ISSUE #651: Support free tickets
in the view  `initiate_checkout`, we exit early if cart total is 0. 

Let's instead make it check individual tickets: `all(t.price == 0 for t in cart.tickets.all())` and then simply "sell the tickets" to the person and then exit.

This also should be accompanied with a refactor. We have "sell the tickets" logic in `complete_checkout`. Move this to a separate helper function so that both the free ticket purchase path and paid ticket purchase path can use this shared function.

We also need one more refactor: the response returned should indicate that the tickets have been sold and the frontend doesn't need to call `complete_checkout`. This will require changing the response schema.

Porcupine1:
This PR serves to Resolve #651 Support free ticket. If the cart is not empty, not stale, and all tickets in the cart have a price of 0.00, then directly give the tickets to the buyer without going through the payment procedure.

TODO:
*Add tests - done


Porcupine1:
I couldn't make the commit without staging the formatting changes (made automatically) to serializer.py,  but it looks like the changes are causing an error. Will work on this.

backend/clubs/views.py
```diff
@@ -4771,28 +4775,45 @@ def initiate_checkout(self, request, *args, **kwargs):
...
             return Response(
                 {
-                    "success": False,
-                    "detail": "Cart total must be nonzero to generate capture context.",
+                    "success": True,
+                    "detail": "Free tickets sold.",
+                    "sold_free_tickets": True,
                 },
-                status=status.HTTP_400_BAD_REQUEST,
+                status=status.HTTP_200_OK,
```
rohangpta, line 4813: This is the default, so not explicitly needed

rohangpta (approved):
Overall this looks great to me. Thanks Thomas!

- You'll need to undo your changes in `backend/clubs/serializers.py`
- It seems there's a merge conflict: you might want to rebase (HINT: use `git rebase -i ticketing`. A tutorial on rebasing is here https://gist.github.com/todgru/7cff3962e76e1360cc2c).

backend/clubs/views.py
```diff
@@ -4831,16 +4852,58 @@ def initiate_checkout(self, request, *args, **kwargs):
                     reason=f"Received {context} with HTTP status {status}",
                 )
 
-            return Response({"success": True, "detail": context})
+            return Response(
+                {
+                    "success": True,
+                    "detail": context,
+                    "solf_free_tickets": False,
```
aviupadhyayula, line 4859: Looks like there's a typo here.

backend/clubs/views.py
```diff
@@ -4929,34 +4992,8 @@ def complete_checkout(self, request, *args, **kwargs):
                 },
                 status=status.HTTP_500_INTERNAL_SERVER_ERROR,
             )
-
-        # Archive transaction data for historical purposes.
-        # We're explicitly using the response data over what's in self.request.user
         orderInfo = transaction_data["orderInformation"]
```
aviupadhyayula, line 4995: nit: feels like this should be in snake case.

rohangpta:
> I couldn't make the commit without staging the formatting changes (made automatically) to serializer.py, but it looks like the changes are causing an error. Will work on this.

An easy sollution to this is to undo the changes and do `git commit --no-verify`...but bonus points for finding the source of the problem

aviupadhyayula (commented):
Great work, Thomas! Left some minor comments. Should be good to go after addressing Rohan's review.

backend/clubs/views.py
```diff
@@ -4771,28 +4775,45 @@ def initiate_checkout(self, request, *args, **kwargs):
...
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
         holding_expiration = timezone.now() + datetime.timedelta(minutes=10)
         tickets.update(holder=self.request.user, holding_expiration=holding_expiration)
 
-        cart_total = tickets.aggregate(total=Sum("price"))["total"]
-        if not cart_total:
+        # If all tickets are free, we can skip the payment process
+        if all(ticket.price == 0 for ticket in cart.tickets.all()):
```
rm03, line 4794: Thoughts on `if (cart_total := cart.tickets.aggregate(total=Sum("price"))["total"]) == 0:`? Since we're now checking if the ticket count is 0 beforehand, there shouldn't be a case where the cart total is 0 without the cart containing all free tickets

Separately, when we add `price` to `create_tickets`, we will need to add a check for non-negative ticket price

backend/clubs/views.py
```diff
@@ -4831,16 +4852,58 @@ def initiate_checkout(self, request, *args, **kwargs):
...
             return Response(
                 {
                     "success": False,
                     "detail": f"Unable to generate capture context: {e}",
+                    "sold_free_tickets": False,
                 },
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
+    def give_tickets(self, orderInfo, cart, reconciliation_id):
```
rm03, line 4873: nit: `order_info` (snake case)

Can you also move this helper below all other views in the viewset for consistency?

rm03 (commented):
Awesome work 🚀

backend/clubs/views.py
Porcupine1: oh, that's a terrible one. Thanks

backend/clubs/views.py
Porcupine1: I remember reading about the walrus operator, when it was released in 3.7/3.8 but never used it. Cool to use it here.

I see what you mean. Since `aggregate` returns `None` if there are no rows in the query result, `total` is mapped to `None`

Porcupine1:
How is it possible to not catch these failing lint tests locally?

backend/clubs/views.py
```diff
@@ -4801,26 +4806,42 @@ def initiate_checkout(self, request, *args, **kwargs):
...
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
         holding_expiration = timezone.now() + datetime.timedelta(minutes=10)
         tickets.update(holder=self.request.user, holding_expiration=holding_expiration)
 
-        cart_total = tickets.aggregate(total=Sum("price"))["total"]
-        if not cart_total:
+        # If all tickets are free, we can skip the payment process
+        if (cart_total := tickets.aggregate(total=Sum("price"))["total"]) == 0:
```
aviupadhyayula, line 4825: nit: can we not use the walrus operator here? It saves 1 line, but makes the code a lot harder to read. (For reference there are 0 places in the codebase where we use it right now.)

backend/clubs/views.py
aviupadhyayula: Good catch, added a check for non-negative ticket pricing in [065c60c](https://github.com/pennlabs/penn-clubs/pull/659/commits/065c60cfed3f966010ee2bd6d74ff04442ea745a).

backend/clubs/views.py
```diff
@@ -5023,24 +5028,44 @@ def initiate_checkout(self, request, *args, **kwargs):
         )
 
         # Assert that the filter succeeded in freezing all the tickets for checkout
-        if tickets.count() != cart.tickets.count():
+        if tickets.count() != cart.tickets.all().count() or tickets.count() == 0:
```
aviupadhyayula, line 5031: We're checking `if not cart.tickets.exists()` in L5014 -- second condition is unnecessary.

backend/clubs/views.py
```diff
@@ -5078,7 +5103,7 @@ def initiate_checkout(self, request, *args, **kwargs):
             )
             if not context or http_status >= 400:
                 raise ApiException(
-                    reason=f"Received {context} with HTTP status {http_status}",
+                    reason=f"Received {context} with HTTP status {status}",
```
aviupadhyayula, line 5106: Where did `status` come from here?

backend/clubs/views.py
```diff
@@ -5087,17 +5112,22 @@ def initiate_checkout(self, request, *args, **kwargs):
                 cart.save()
 
             # Place hold on tickets for 10 mins
-            holding_expiration = timezone.now() + datetime.timedelta(minutes=10)
-            tickets.update(
-                holder=self.request.user, holding_expiration=holding_expiration
+            self.place_hold_on_tickets(tickets)
```
aviupadhyayula, line 5115: Do we place holds anywhere else in the viewset? Would prefer to keep this logic inside the view if not.

backend/clubs/views.py
```diff
@@ -5023,24 +5028,44 @@ def initiate_checkout(self, request, *args, **kwargs):
...
+                    "reconciliationId": None,
+                    "firstName": self.request.user.first_name,
+                    "lastName": self.request.user.last_name,
+                    "phoneNumber": None,
+                    "email": self.request.user.email,
+                },
+            }
+
+            # Place hold on tickets for 10 mins
+            self.place_hold_on_tickets(tickets)
```
aviupadhyayula, line 5059: Why are we holding here?

backend/clubs/views.py
```diff
@@ -5354,6 +5354,56 @@ def transfer(self, request, *args, **kwargs):
     def get_queryset(self):
         return Ticket.objects.filter(owner=self.request.user.id)
 
+    def give_tickets(self, order_info, cart, reconciliation_id):
```
aviupadhyayula, line 5357: nit: would be nice to make it clear that this is for internal use only.

backend/clubs/views.py
```diff
@@ -5354,6 +5354,56 @@ def transfer(self, request, *args, **kwargs):
...
+        """
+        Helper function that give the tickets to the user/buyer
+        and archive the transaction data
```
aviupadhyayula, line 5360: nit: verb tenses here are wrong.

backend/clubs/views.py
```diff
@@ -5354,6 +5354,56 @@ def transfer(self, request, *args, **kwargs):
...
+        """
+
+        # At this point, we have validated that the payment was authorized
+        # Give the tickets to the user
+        tickets = (
+            cart.tickets.select_for_update()
+            .filter(holder=self.request.user)
+            .prefetch_related("carts")
+        )
```
aviupadhyayula, line 5369: Ah, now I see why we're placing holds even on free tickets. Workaround might be to pass `check_for_hold` as a parameter (defaults to true)? cc @rm03

backend/clubs/views.py
```diff
@@ -5354,6 +5354,56 @@ def transfer(self, request, *args, **kwargs):
...
+                    buyer_first_name=order_info["billTo"]["firstName"],
+                    buyer_last_name=order_info["billTo"]["lastName"],
+                    # TODO: investigate why phone numbers don't show in test API
+                    buyer_phone=order_info["billTo"].get("phoneNumber", None),
+                    buyer_email=order_info["billTo"]["email"],
+                )
+            )
+
+        TicketTransactionRecord.objects.bulk_create(transaction_records)
+        tickets.update(owner=self.request.user, holder=None)
```
aviupadhyayula, line 5390: nit: swap the order of these two lines.

backend/tests/clubs/test_ticketing.py
```diff
@@ -115,6 +121,7 @@ def test_create_ticket_offerings_bad_perms(self):
         self.client.login(username=self.user2.username, password="test")
         qts = {
             "quantities": [
+                {"type": "_free", "count": 30, "price": 0},
```
aviupadhyayula, line 124: Don't need to add this here. Your changes are isolated to `initiate_checkout`. Same goes for all the other changes to tests.

backend/tests/clubs/test_ticketing.py
```diff
@@ -58,10 +58,16 @@ def commonSetUp(self):
...
+        {"type": "free", "count": 30, "price": 0.0},
         {"type": "normal", "count": 20, "price": 15.0},
         {"type": "premium", "count": 10, "price": 30.0},
     ]
 
+    self.tickets0 = [
+        Ticket.objects.create(type="free", event=self.event1, price=0.0)
+        for _ in range(30)
+    ]
+
```
aviupadhyayula, line 70: nit: these will only be useful for the `initiate_checkout` test case for free tickets, would prefer not to add them globally.

backend/tests/clubs/test_ticketing.py
```diff
@@ -794,12 +813,13 @@ def test_get_cart_replacement_required_sold_out(self):
         to_add = set(map(lambda t: str(t.id), tickets_to_add))
         self.assertEqual(len(in_cart & to_add), 0, in_cart | to_add)
 
-    def test_initiate_checkout(self):
+    def test_initiate_checkout_free_and_non_free_tickets(self):
```
aviupadhyayula, line 816: nit: would prefer to create a separate test case for this.

backend/tests/clubs/test_ticketing.py
```diff
@@ -833,16 +855,69 @@ def test_initiate_checkout(self):
...
-        self.assertEqual(held_tickets.count(), 2, held_tickets)
+        self.assertEqual(held_tickets.count(), 3, held_tickets)
+        self.assertEqual(held_tickets.filter(type="normal").count(), 1, held_tickets)
         self.assertEqual(held_tickets.filter(type="normal").count(), 1, held_tickets)
         self.assertEqual(held_tickets.filter(type="premium").count(), 1, held_tickets)
 
+    def test_initiate_checkout_only_free_tickets(self):
+        self.client.login(username=self.user1.username, password="test")
+
+        # Add a few tickets to cart
```
aviupadhyayula, line 866: nit: make it clear that only free tickets are being added.

aviupadhyayula (commented):
Great work Thomas! I imagine we'll be getting a lot of use out of this functionality. LGTM pending minor changes to the view and a few changes to the tests. 🚀

backend/clubs/views.py
aviupadhyayula: nvm, thought about it some more. Holds are how we validate that a ticket was placed in a user's cart before they initiated checkout (even if it's not strictly needed here). Don't like the idea of adding a workaround to that.

backend/clubs/views.py
aviupadhyayula: nit: should also make clear that only held tickets are transferred.

backend/clubs/views.py
Porcupine1: You mean only to be used in `initiate_checkout` and `complete_checkout`?

backend/tests/clubs/test_ticketing.py
Porcupine1: Don't you think if we're going to have free tickets we still want to test how they behave in other instances where the user buys free and non-free tickets to catch bugs we wouldn't otherwise see if free tickets were left out?

backend/tests/clubs/test_ticketing.py
Porcupine1: refer to previous reply

backend/tests/clubs/test_ticketing.py
Porcupine1: Since there are only two routes to purchasing a ticket: `cart_total=0` and `cart_total>0`, including free tickets here doesn't make the route here special since it's the case `cart_total>0`. Would you rather have me revert to the initial name?

backend/clubs/views.py
Porcupine1: Must have slipped through conflict resolution when merging. Will fix it

backend/tests/clubs/test_ticketing.py
aviupadhyayula: Good point -- supporting free tickets means that they should work end-to-end, even if the only code change is in `initiate_checkout`. I still think adding them globally (and integrating them with all current tests) is a bit much. Maybe we can add specific tests (for only free tickets) for all the major routes? i.e. `test_create_ticket_offerings_free_tickets`

backend/clubs/views.py
aviupadhyayula: More in the spirit of `_calculate_cart_total`. [Quick read](https://realpython.com/python-double-underscore/#single-leading-underscore-in-python-names).

backend/tests/clubs/test_ticketing.py
aviupadhyayula: Let's split this into 3 tests -- the original (tests only paid tickets), test only free tickets, and test a mix.

Porcupine1: will do

backend/clubs/views.py
rohangpta: [Less quick read, but still useful](https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name)

backend/tests/clubs/test_ticketing.py
Porcupine1: Even if this means calling `Ticket.objects.create(type="free", ...` every time we want to test free tickets?

aviupadhyayula: That'll get called at each test regardless, whether it's in the setup or in an individual test case. ([`setUp` gets called between each case.](https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUp)) 

For the sake of getting this PR shipped, let's leave the creation in setup for now -- or whatever impl you currently have, provided it works.

Porcupine1: It's ok, I can make the changes right now

backend/clubs/views.py
```diff
@@ -5023,24 +5028,44 @@ def initiate_checkout(self, request, *args, **kwargs):
...
+                    "reconciliationId": None,
+                    "firstName": self.request.user.first_name,
+                    "lastName": self.request.user.last_name,
+                    "phoneNumber": None,
+                    "email": self.request.user.email,
+                },
+            }
+
+            # Place hold on tickets for 10 mins
+            self.place_hold_on_tickets(tickets)
```
aviupadhyayula, line 5059: Can you make this an internal function as well? See our discussion on `_give_tickets`.

backend/clubs/views.py
```diff
@@ -5354,6 +5354,56 @@ def transfer(self, request, *args, **kwargs):
...
+
+        cart.checkout_context = None
+        cart.save()
+
+    def place_hold_on_tickets(self, tickets):
+        """
+        Helper function that places a 10 minute hold on tickets for a user
+        """
+        holding_expiration = timezone.now() + datetime.timedelta(minutes=10)
+        tickets.update(holder=self.request.user, holding_expiration=holding_expiration)
```
aviupadhyayula, line 5405: Does this work? There's no `self.request.user` here, is there?

Porcupine1: yes, passes tests

backend/clubs/views.py
Porcupine1: you're right, thanks

backend/clubs/views.py
aviupadhyayula: What happens if it's called from outside an active route? i.e. TicketViewSet._place_holds (which we'll be doing with _give_tickets in the issue_tickets route)

Porcupine1: I don't quite understand. `_place_hold_on_tickets` and `_give_tickets` are not "static" methods

Porcupine1: Would you rather have me pass `user` as a parameter?

backend/clubs/views.py
```diff
@@ -5354,6 +5354,52 @@ def transfer(self, request, *args, **kwargs):
     def get_queryset(self):
         return Ticket.objects.filter(owner=self.request.user.id)
 
+    def _give_tickets(self, user, order_info, cart, reconciliation_id):
```
aviupadhyayula, line 5357: Do we need to be passing in `reconciliation_id` when it's already part of `order_info`?

backend/clubs/views.py
```diff
@@ -5354,6 +5354,52 @@ def transfer(self, request, *args, **kwargs):
...
+        cart.tickets.clear()
+        for ticket in tickets:
+            ticket.send_confirmation_email()
+
+        Ticket.objects.update_holds()
+
+        cart.checkout_context = None
+        cart.save()
+
+    def _place_hold_on_tickets(self, tickets):
```
aviupadhyayula, line 5396: Should probably add tests for this, like the tests we have for `_calculate_cart_total`.

backend/clubs/views.py
aviupadhyayula, line 5357: Same here. Dedicated tests (now that we've abstracted ownership transfers) would be helpful.

aviupadhyayula (commented):
LGTM

backend/clubs/views.py
```diff
@@ -5027,20 +5032,40 @@ def initiate_checkout(self, request, *args, **kwargs):
...
             )
 
         cart_total = self._calculate_cart_total(cart)
 
+        # If all tickets are free, we can skip the payment process
         if not cart_total:
+            order_info = {
+                "amountDetails": {"totalAmount": "0.00"},
+                "billTo": {
+                    "reconciliationId": None,
```
aviupadhyayula, line 5050: Don't think we need this line.

Porcupine1: Please explain

aviupadhyayula: We're not using `reconciliationId` from `order_info` in `_give_tickets`.

Porcupine1: Yah, good catch. Wasn't clear which line you meant from my end

PR #659, opened 2024-04-15
Ticketing price integration

julianweng:
Adds price field to list/add views for tickets in any given event. Integrates price field into club admin ticket creation page.

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -46,6 +46,7 @@ from django.db.models import (
     DurationField,
     ExpressionWrapper,
     F,
+    Max,
     Prefetch,
     Q,
     Sum,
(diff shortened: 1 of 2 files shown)


aviupadhyayula:
Added a check to disallow negative ticket prices in [065c60c](https://github.com/pennlabs/penn-clubs/pull/659/commits/065c60cfed3f966010ee2bd6d74ff04442ea745a). Would be great if we could throw a failure message in the frontend, when users are creating or modifying ticket pricing.

julianweng:
Sounds good @aviupadhyayula, I just made it impossible to enter negative and fractional numbers for now: should probably make a clearer UI down the line but should be self-explanatory for most users as-is. @esinx are we good to merge? View changes would be useful for the rest of frontend development.

PR #661, opened 2024-04-16
Add support for group discounts

aviupadhyayula:
Adds support for group discounts on a per-ticket-class basis. End-user flow will be: 
1. At ticket creation, admins can choose two parameters for each ticket type: `group_size` and `group_discount`. Either both or neither of these params must be set for each ticket type.
2. When checking out, if a user has `group_size` of the same tickets in their cart, the group discount is automatically applied. Each affected ticket has its price reduced to `(1 - group_discount) * price`.


backend/clubs/views.py
```diff
@@ -2629,14 +2635,50 @@ def create_tickets(self, request, *args, **kwargs):
...
+                {"detail": "Group discount must be between 0 and 1"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        # Min group sizes must be greater than 1
+        if any(item.get("group_size", 2) < 1 for item in quantities):
+            return Response(
+                {"detail": "Min group size must be greater than 1"},
                 status=status.HTTP_400_BAD_REQUEST,
             )
```
aviupadhyayula, line 2658: I'm enforcing that groups must be larger than 1. Do we agree with this?

backend/clubs/views.py
```diff
@@ -4841,7 +4883,21 @@ def initiate_checkout(self, request, *args, **kwargs):
...
+        }
+
+        cart_total = sum(
+            ticket.price * (1 - ticket.group_discount)
+            if ticket.group_discount
+            and ticket.group_size
+            and ticket_type_counts[ticket.type] >= ticket.group_size
+            else ticket.price
+            for ticket in tickets
+        )
```
aviupadhyayula, line 4899: This is pretty ugly, but I couldn't think of a better way to do it.

aviupadhyayula:
It feels a little weird that we're duplicating information across many ticket instances. Attributes like `group_discount`, `price`, `type`, `event`, etc. don't need to be replicated so many times at a per-instance level. 

Should we instead create something like `TicketInformation` (one instance per ticket type), and then use that as a foreign key in each individual ticket? 

(Not super relevant to this PR, but something I had in mind.)

rohangpta:
> It feels a little weird that we're duplicating information across many ticket instances. Attributes like `group_discount`, `price`, `type`, `group_size`, etc. don't need to be replicated so many times at a per-instance level.
> 
> Should we instead create something like `TicketInformation` (one instance per ticket type), and then use that as a foreign key in each individual ticket?
> 
> (Not super relevant to this PR, but something I had in mind.)

I chose to denormalize bc O(num tickets) is likely to get decently big and joining could be expensive (and we'd have to join almost every time we look at tickets). Not super tied to it though.

backend/clubs/views.py
rohangpta: Did you mean <= 1 here?

backend/clubs/views.py
rohangpta: Setting `ticket.group_discount` to default to 0 is probably a good idea here:
- You can skip the `if ticket.group_discount` check
- `1 - ticket.group_discount` won't fail at runtime if the latter is `None`

rohangpta (approved):
Looks good overall. Few small comments. Since we're doing a lot of validation, I wonder if we can merge the logic to one big for loop that iterates over all `item` in `quantities` and asserts whatever. Not a blocker though

backend/clubs/views.py
aviupadhyayula: Yup, thanks for catching.

aviupadhyayula:
> > It feels a little weird that we're duplicating information across many ticket instances. Attributes like `group_discount`, `price`, `type`, `group_size`, etc. don't need to be replicated so many times at a per-instance level.
> > Should we instead create something like `TicketInformation` (one instance per ticket type), and then use that as a foreign key in each individual ticket?
> > (Not super relevant to this PR, but something I had in mind.)
> 
> I chose to denormalize bc O(num tickets) is likely to get decently big and joining could be expensive (and we'd have to join almost every time we look at tickets). Not super tied to it though.

Yeah, makes sense. Let's stick to it for now. As we add on fields (group discounts, promo codes), we can revisit.

rohangpta (approved):
LGTM

aviupadhyayula:
CI's failing because of frontend changes that were forcefully merged into the ticketing main branch. Going to merge for now -- don't want to mess with frontend dev.

Issue #662, opened 2024-04-16
Django tests periodically fail in `test_send_invites_with_acceptance_email`


aviupadhyayula:
Tests merged in #624 periodically fail when the invite's expiry time doesn't exactly match the tested expiry time (down to the second). [We flagged this in the original PR](https://github.com/pennlabs/penn-clubs/pull/624#discussion_r1554533871), but didn't get around to it.

[Here's an example of CI failing in this manner](https://github.com/pennlabs/penn-clubs/actions/runs/8711153597/job/23894635932?pr=661).

PR #664, opened 2024-04-17
Fix intermittent failure of tests introduced in #624

Porcupine1:
This PR aims fix the #662, by checking that that the tested expiry datetime is within 5 minutes of the invite's expires_at datetime


aviupadhyayula (approved):
LGTM! Thanks for the quick turnaround, Thomas.

PR #665, opened 2024-04-17
Fix ticket interface for sellers

julianweng:
Fixes pages for sellers to view all ticketed events and ticket holders for each.

## Issuing Tickets
Screenshot 2024-04-24 at 12 05 55 AM

### Pennkey validation
Screenshot 2024-04-25 at 9 01 28 PM

### Ticket Fine-tuning
Screenshot 2024-04-25 at 9 27 13 PM


frontend/pages/tickets/[[...slug]].tsx
```diff
@@ -85,126 +53,142 @@ const Text = styled.h1`
...
+        {Object.values(tickTypes).map((ticket, i) => (
+          <TicketCard key={i} ticket={ticket} buyersPerm={buyers.buyers} />
         ))}
       </Container>
     </>
   )
 }
 
-const TicketCard = ({ ticket }) => {
+const TicketCard = ({ ticket, buyersPerm }) => {
```
Comment (esinx) [inline frontend/pages/tickets/[[...slug]].tsx:114]: What's the type of `buyersPerm`?

Comment (julianweng) [reply frontend/pages/tickets/[[...slug]].tsx:114]: It's a list of buyer objects returned from the relevant endpoint: was using it being null or not as a stand-in for whether or not user has perms (owns the club etc.) to view ticket holders, but can change the prop to be the result of the null-check to make things cleaner.

frontend/pages/tickets/[[...slug]].tsx
```diff
@@ -85,123 +53,160 @@ const Text = styled.h1`
...
-  // })
-
-  // tickets = []
-  // for (const [_, value] of Object.entries(ticks)) {
-  //   tickets.push(value)
-  // }
+  const tickTypes = {}
+
+  // For given ticket type, get the total and available tickets, as well as buyers
+  for (const ticket of totals) {
```
Comment (joyliu-q) [inline frontend/pages/tickets/[[...slug]].tsx:81]: Duplicate entry of ticket.type and map key

julianweng:
Not great branch etiquette considering these are all over the place but fixed some UI wording and the following bugs here:

1. Items remain in cart after deletion through dedicated button
2. Item quantities only update when pressing enter on actual quantity input rather than the "finish editing" button
3. Would show error toast when successfully adding tickets to cart

frontend/pages/tickets/[[...slug]].tsx
Comment (julianweng) [reply frontend/pages/tickets/[[...slug]].tsx:123]: Not ideal but kept since map is more appropriate for the (hacky) way we populate tickTypes but still need the type later on, so avoids the inevitable intermediate step of feeding the key back into the value.

PR #666, opened 2024-04-17
Ticketing backend tests

rohangpta:


backend/clubs/views.py
```diff
@@ -2671,9 +2692,9 @@ def create_tickets(self, request, *args, **kwargs):
             Ticket(
                 event=event,
                 type=item["type"],
-                price=item["price"],
-                group_discount=item.get("group_discount", None),
-                group_size=item.get("group_size", None),
+                price=item.get("price", 0),
+                group_discount=item.get("group_discount", 0),
+                group_size=item.get("group_size", 0),
```
aviupadhyayula, line 2697: Note: setting this to 0 won't change how cart totals are calculated, since we check `if group_size` before applying the group discount in `initiate_checkout`. However, it's a little unclear, and may trip us up later. What do you think of either 1) defaulting `group_size` to 0 in the `Ticket` model, and checking `if group_size != 0` or 2) reverting this line to `item.get("group_size", None)`?

backend/clubs/views.py
```diff
@@ -4767,15 +4788,17 @@ def cart(self, request, *args, **kwargs):
...
+        )
 
         # Replace in-cart tickets that have been bought/held by someone else
         tickets_to_replace = cart.tickets.filter(
             Q(owner__isnull=False) | Q(holder__isnull=False)
         ).exclude(holder=self.request.user)
 
         # In most cases, we won't need to replace, so exit early
-        if not tickets_to_replace:
+        if not tickets_to_replace.count():
```
aviupadhyayula, line 4801: nit: does `exists()` work here

backend/clubs/views.py
```diff
@@ -4786,19 +4809,20 @@ def cart(self, request, *args, **kwargs):
...
                 owner__isnull=True,
                 holder__isnull=True,
-            ).first()
+            ).exclude(id__in=cart.tickets.all())[: ticket_class["count"]]
 
-            if ticket is not None:
-                replacement_tickets.append(ticket)
-            else:
-                sold_out_count += 1
+            if diff := (ticket_class["count"] - tickets.count()):
```
rm03, line 4823: is this if statement necessary? if `diff` is 0 then we won't be adding anything to `sold_out_count` anyway

rm03 (approved):
:goat:

backend/clubs/views.py
```diff
@@ -4786,19 +4809,20 @@ def cart(self, request, *args, **kwargs):
...
-            ticket = Ticket.objects.filter(
-                event=gone_ticket.event,
-                type=gone_ticket.type,
+            tickets = Ticket.objects.filter(
+                event=ticket_class["event"],
+                type=ticket_class["type"],
                 owner__isnull=True,
                 holder__isnull=True,
-            ).first()
+            ).exclude(id__in=cart.tickets.all())[: ticket_class["count"]]
```
aviupadhyayula, line 4821: nit: I think we can fetch this once, outside the loop. something like `cart.tickets.values_list('id', flat=True)`

backend/clubs/views.py
```diff
@@ -4786,19 +4809,20 @@ def cart(self, request, *args, **kwargs):
...
 
-            if ticket is not None:
-                replacement_tickets.append(ticket)
-            else:
-                sold_out_count += 1
+            if diff := (ticket_class["count"] - tickets.count()):
+                sold_out_count += diff
```
aviupadhyayula, line 4824: Is the check needed here? If `diff` is 0 then `sold_out_count += diff` doesn't do anything.

aviupadhyayula (commented):
🚀

backend/clubs/views.py
rohangpta: Will revert. Looks like we only need to non-nullify `group_discount`?

backend/clubs/views.py
rohangpta: Removed

backend/clubs/views.py
rohangpta: Removed

backend/clubs/views.py
aviupadhyayula: Yup, since it defaults to 0 in the model definition.

rohangpta: Also just noticed that you've defined `group_size` as a PositiveIntegerField so `group_size` = 0 would have also failed at runtime!

PR #667, opened 2024-04-18
Add guards on ticketed event deletion

aviupadhyayula:
Checks for associated tickets that have been bought or checked out before deleting an event. This moves previous, incorrect logic to the right place (general events to club events).


backend/clubs/views.py
```diff
@@ -2836,9 +2836,11 @@ def create(self, request, *args, **kwargs):
 
         return super().create(request, *args, **kwargs)
 
+    @update_holds
```
aviupadhyayula, line 2839: Adding an `update_holds` here since we're checking for holds before destroy.

backend/clubs/views.py
```diff
@@ -2848,6 +2850,18 @@ def destroy(self, request, *args, **kwargs):
                 f"If you would like to do this, email {settings.FROM_EMAIL}."
             )
 
+        if (
+            Ticket.objects.filter(event=event)
+            .filter(Q(owner__isnull=False) | Q(holder__isnull=False))
+            .exists()
+        ):
+            raise DRFValidationError(
```
aviupadhyayula, line 2858: is a `DRFValidationError` appropriate?

rm03: For consistency, I think this is fine. According to the [docs](https://www.django-rest-framework.org/api-guide/exceptions/#validationerror), the API will return status code 400, which is appropriate here

backend/tests/clubs/test_ticketing.py
```diff
@@ -389,6 +389,26 @@ def test_remove_from_cart_extra(self):
...
+        self.tickets1[0].save()
+
+        self.client.login(username=self.user1.username, password="test")
+        resp_held = self.client.delete(
+            reverse("club-events-detail", args=(self.club1.code, self.event1.pk))
+        )
+        self.assertEqual(resp_held.status_code, 400, resp_held.content)
+
+        # Simulate purchase (transfer ticket)
+        self.tickets1[0].owner = self.user1
```
rm03, line 404: nit: can you set the holder to `None` here?

rm03 (approved):
lgtm

backend/tests/clubs/test_ticketing.py
aviupadhyayula: we love the attention to detail, thanks for catching

backend/clubs/views.py
```diff
@@ -2929,6 +2931,18 @@ def destroy(self, request, *args, **kwargs):
                 f"If you would like to do this, email {settings.FROM_EMAIL}."
             )
 
+        if (
+            Ticket.objects.filter(event=event)
+            .filter(Q(owner__isnull=False) | Q(holder__isnull=False))
+            .exists()
```
Porcupine1, line 2937: Was this change made to enable  admin to delete event even though it isn't over? The original code checks `now <= event.end_time`

Porcupine1 (commented):
Everything else looks good

backend/clubs/views.py
aviupadhyayula: Yeah -- if a ticketed event is cancelled & nobody's bought tickets for it, it should be deletable.

PR #668, opened 2024-04-18
Filter Events by active ones (starts after now)

vcai122:


backend/clubs/views.py
```diff
@@ -2426,7 +2426,7 @@ def destroy(self, request, *args, **kwargs):
         return super().destroy(request, *args, **kwargs)
 
     def get_queryset(self):
-        qs = Event.objects.all()
+        qs = Event.objects.filter(start_time__gte=timezone.now())
```
vcai122, line 2429: idk if past events are ever used but the page takes forever to load rn

rm03: Currently past events aren't accessible via the frontend...however with ticketing this will change, so I don't think this should be default behavior. Instead we could add this as a query parameter + an index on start time? @aviupadhyayula feel free to chime in

aviupadhyayula: Makes sense to me. I'm not a huge fan of filtering out past events by default. 

Also, this is probably an issue with our serializer, and not the query.

vcai122: @rm03's suggestion sounds pretty good. For context, Jacky is trying to use this route to render events on iOS but the page is just too massive to load since it contains all events ever

rohangpta: Maybe we have some unnoticed N+1s here?

rm03: > Maybe we have some unnoticed N+1s here?

Doesn't look like it, just checked locally

PR #669, opened 2024-04-20
Add Group Discount to Create Ticket Flow and Auto Scroll Down

joyliu-q:
Create Ticket Changes:
- Initial styling of group discounts (will be using tailwind and stuff to add icons and make it look more clean)

UX Event Changes:
- When you select an event in `/club/[name]/edit/events`, it will autoscroll to the ref for editing event details
- Automatically close the modal if the request to create the ticket was successful 
  - In the future, we may want to redirect this to another page or show something, but for now there's nothing to redirect to or show.
  
Screenshot 2024-04-20 at 6 22 23 PM
Screenshot 2024-04-20 at 6 22 33 PM


frontend/components/ClubEditPage/TicketsModal.tsx
```diff
@@ -129,45 +203,40 @@ type Ticket = {
...
     const ticks = [...tickets]
-    ticks.push({ name: '', count: null, price: null })
+    ticks.push({
+      name: '',
+      count: null,
+      price: null,
+      groupDiscount: null,
+      groupNumber: null,
+    })
     setTickets(ticks)
```
esinx, line 240: prefer immutability:
```typescript
setTickets(t => [...ticks, {name: '', count: null, price: null, groupDiscount: null, groupNumber: null}])
```

frontend/components/ClubEditPage/TicketsModal.tsx
```diff
@@ -176,10 +245,17 @@ const TicketsModal = (props: { event: ClubEvent }): ReactElement => {
       const quantities = tickets
         .filter((ticket) => ticket.count != null)
         .map((ticket) => {
+          const usingGroupPricing = ticket.groupDiscount && ticket.groupNumber
           return {
             type: ticket.name,
             count: parseInt(ticket.count || ''),
```
esinx, line 251: `parseInt` and `parseFloat` on empty string returns `NaN`, so maybe default to 0 instead? Minor note: it's better to use nullish comparison(`??`) in place of `||` since `''` will also cause the fallback to be triggered.

frontend/pages/events/[id].tsx
```diff
@@ -118,6 +125,79 @@ const Divider = styled.hr`
   margin: 20px 0;
 `
 
+const Input = styled.input`
+  border: 1px solid ${ALLBIRDS_GRAY};
+  outline: none;
+  color: ${CLUBS_GREY};
+  flex: '0 0 auto';
```
Comment (esinx) [inline frontend/pages/events/[id].tsx:132]: no need to wrap flex property in `'`

frontend/pages/events/[id].tsx
```diff
@@ -118,6 +125,79 @@ const Divider = styled.hr`
...
+const TicketItem = ({
+  ticket,
+  name,
+  price,
+  max,
+  changeCount,
+  index,
+}): ReactElement => {
+  const [count, setCount] = useState(ticket.count)
+  const handleCountChange = (e: { target: { value: string } }) => {
```
Comment (esinx) [inline frontend/pages/events/[id].tsx:163]: the proper typing would be
```typescript
const handleCountChange = (e: React.ChangeEvent<HTMLInputElement>) => ...
```

frontend/pages/events/[id].tsx
```diff
@@ -118,6 +125,79 @@ const Divider = styled.hr`
...
+`
+
+type Ticket = {
+  type: string
+  price: string
+  max: string
+  count: string | null
+}
+
+const TicketItem = ({
```
Comment (esinx) [inline frontend/pages/events/[id].tsx:154]: Use `React.FC` Type:

```
type TicketItemProps = { ... }
const TicketItem: React.FC<TicketItemProps> = ({ ... }) => ...
```

frontend/pages/events/[id].tsx
```diff
@@ -118,6 +125,79 @@ const Divider = styled.hr`
...
+  ticket,
+  name,
+  price,
+  max,
+  changeCount,
```
Comment (esinx) [inline frontend/pages/events/[id].tsx:159]: convention: use `on...` for sending events to parent.
also try not to rely on index as the children should barely be aware of its own index(it's better to delegate this job to the parent and make the parent take action as it's guaranteed to have context about each item's index)

esinx (changes requested):

frontend/components/ClubEditPage/EventsCard.tsx
```diff
@@ -465,13 +466,19 @@ export default function EventsCard({ club }: EventsCardProps): ReactElement {
...
+            block: 'start',
+          })
           setDeviceContents(obj)
         }}
       />
       <Line />
       <CreateTickets event={event} />
       <Line />
-      <EventPreview event={event} />
+      <div ref={eventDetailsRef as React.RefObject<HTMLDivElement>}>
```
esinx, line 479: Why not change the type of the ref itself?

frontend/pages/events/[id].tsx
```diff
@@ -118,6 +125,86 @@ const Divider = styled.hr`
...
+  onCountChange: (newCount: number) => void
+}
+
+const TicketItem: React.FC<TicketItemProps> = ({
+  ticket,
+  name,
+  price,
+  max,
+  onCountChange,
+}): ReactElement => {
```
Comment (esinx) [inline frontend/pages/events/[id].tsx:168]: No need to specify return type!

esinx (approved):
Could make some minor changes but lgtm!

PR #671, opened 2024-04-21
Use capture context in cart checkout

### LINKED ISSUE #656: Tie a capture context to a cart, and use it to validate the transient token
Pesky edge cases

aviupadhyayula:
Ties the checkout context generated in `initiate_checkout` to the user's cart. Uses said checkout context to decode & verify the transient token. 

Downstream changes: 
1. If the user initiates a second checkout while one is in progress, the cart's capture context will be updated, and the first checkout will fail. 
2. This also guards against users maliciously modifying the cart after the capture context has been generated. Authorization only transfers tickets that are held by the user, and holds are only given when initiating a checkout.

Closes #656.


aviupadhyayula:
Checked that verification of the transient token works as intended by [following this](https://penn-labs.slack.com/archives/C05UD8MF894/p1713044788978199).

backend/clubs/views.py
```diff
@@ -332,19 +332,15 @@ def hour_to_string_helper(hour):
     return hour_string
 
 
-def validate_transient_token(tt: str) -> Tuple[bool, str]:
+def validate_transient_token(cc: str, tt: str) -> Tuple[bool, str]:
     """Validate the integrity of the transient token using
-    the public key (JWK) obtained from the public key endpoint"""
+    the associated capture context"""
```
aviupadhyayula, line 337: Checked that verification of the transient token works as intended by [following this](https://penn-labs.slack.com/archives/C05UD8MF894/p1713044788978199).

backend/clubs/models.py
```diff
@@ -1789,6 +1789,7 @@ class Cart(models.Model):
     owner = models.OneToOneField(
         get_user_model(), related_name="cart", on_delete=models.CASCADE
     )
+    checkout_context = models.CharField(max_length=8297, blank=True, null=True)
```
aviupadhyayula, line 1792: Played around with [Cybersource's code sample](https://github.com/CyberSource/cybersource-rest-samples-python/blob/master/samples/UnifiedCheckout/generate-unified-checkout) for generating a capture context. Returned capture context was always 8297 chars in length. Couldn't find this documented anywhere though.

rohangpta (approved):
Looks great! Thank you for picking this up

backend/clubs/models.py
rohangpta: b64 encoding adds a fixed overhead over what is likely to be a standard format...would be my best guess

Can you please add a comment explaining why it's 8297?

aviupadhyayula: Yup, makes sense. Added.

PR #672, opened 2024-04-21
Add support for ticket drop times

### LINKED ISSUE #655: Add a `ticket_drop_time` to control ticket visibility and editing
`create_tickets` currently deletes and recreates tickets / other parameters around the event every time it is called

This could be very problematic if a bought ticket is deleted since the organizer decides to go in and add a few more tickets.

We should:
- Add a parameter to events specifying a `ticket_drop_time`
- Make tickets publicly available for purchase only after this time
- Freeze ticket offering editing after this time

aviupadhyayula:
Changelog: 
1. Allows admins to specify when tickets will "drop" in `create_tickets`. After this "drop time" has passed, ticket offerings cannot be edited. 
2. Tickets cannot be added to cart until the drop time has elapsed. 
3. Tickets cannot be edited after they've been sold or checked out in a user's cart.

Closes #655.


backend/Pipfile
```diff
@@ -56,6 +56,7 @@ numpy = "*"
 inflection = "*"
 cybersource-rest-client-python = "*"
 pyjwt = "*"
+freezegun = "*"
```
aviupadhyayula, line 59: Used in the unit tests to simulate 12 hours passing after an event's creation. [Here's the library's repo](https://github.com/spulec/freezegun).

rohangpta: FYI, you might have used `pipenv install` which updates packages where possible. For example, ruff is now pinned to 0.4.1 in the lock file (which may be breaking with previous versions). This is probably not your intended behavior, so you might want to revert and do a `pipenv upgrade` instead 

[Reading](https://stackoverflow.com/questions/55592586/how-to-install-update-package-with-pipenv-without-updating-the-rest-of-package)

backend/clubs/views.py
```diff
@@ -2707,6 +2724,11 @@ def create_tickets(self, request, *args, **kwargs):
             event.ticket_order_limit = order_limit
             event.save()
 
+        delay_drop = request.data.get("delay_drop", None)
```
rohangpta, line 2727: Why + 12h here? Seems arbitrary. Can we maybe give the user better control on when exactly their tickets drop?

backend/tests/clubs/test_ticketing.py
```diff
@@ -155,6 +156,43 @@ def test_create_ticket_offerings_bad_data(self):
...
+        # Drop time should be set
+        self.assertIsNotNone(self.event1.ticket_drop_time)
+
+        # Drop time should be 12 hours from initial ticket creation
+        expected_drop_time = timezone.now() + timezone.timedelta(hours=12)
+        diff = abs(self.event1.ticket_drop_time - expected_drop_time)
+        self.assertTrue(diff < timezone.timedelta(minutes=5))
+
+        # Move Django's internal clock 13 hours forward
+        with freeze_time(timezone.now() + timezone.timedelta(hours=13)):
```
rohangpta, line 186: Nice usage of an external lib!

rohangpta (changes requested):

backend/clubs/views.py
aviupadhyayula: No particular reason behind choosing 12. Will modify `delay_drop` to accept an integer value, and use that instead.

rohangpta: Why not just a timestamp?

aviupadhyayula: Like a Unix timestamp? Sure.

rohangpta: Not necessarily a Unix timestamp...there's some use cases around the codebase where we pass in datetime from frontend to backend. For example, we allow club admins to control the `application_start_time`, `application_end_time`, etc. You could use those as guiding examples perhaps

backend/Pipfile
aviupadhyayula: Thanks for the link, was helpful. Fixed.

aviupadhyayula:
Added new feats in later commits (i.e. admins cannot edit tickets after they've been sold or checked out). Re-requesting a review.

aviupadhyayula:
Thinking about this more: it's weird if users see tickets as available in `/event/tickets`, but can't add them to their cart. Thoughts on only displaying tickets as available once the drop time has elapsed?

backend/clubs/views.py
```diff
@@ -2618,6 +2625,10 @@ def create_tickets(self, request, *args, **kwargs):
                             order_limit:
                                 type: int
                                 required: false
+                            drop_time:
+                                type: string
+                                format: date-time
```
rm03, line 2630: nit: can you update the response schema to reflect 403

rm03:
> Thinking about this more: it's weird if users see tickets as available in `/event/tickets`, but can't add them to their cart. Thoughts on only displaying tickets as available once the drop time has elapsed?

Agreed, this would be better

backend/clubs/views.py
aviupadhyayula: Thanks for catching this, @rm03. Your commitment to detail is bot-like. Appreciate your kind review.

rohangpta:
Nice work! Afterthought: would it be cool to show a countdown to ticket drop time on the frontend?

aviupadhyayula:
> Nice work! Afterthought: would it be cool to show a countdown to ticket drop time on the frontend?

Yeah, that's a good idea. cc @julianweng would be cool to put this in.

PR #675, opened 2024-04-21
Fix backend tests for group discounts, new cart API, and more

aviupadhyayula:
Changelog: 
1. Refactors total price summation to an internal helper function. Adds tests for it. (Previous code was failing but wasn't caught, since we mock `initiate_checkout` and `complete_checkout`. Breaking this logic into its own, testable function will hopefully catch future mistakes.)
2. Adds tests for creating tickets with group discounts.
3. Adds tests for summing a cart total with group discounts applied. 
4. Only populates `sold_out` if there are tickets that were not replaced.
5. Aligns tests with changes introduced in ^ and #670.

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -4831,21 +4831,10 @@ class TicketViewSet(viewsets.ModelViewSet):
                                     items:
                                         type: object
                                         properties:
-                                            id:
-                                                type: integer
                                             event:
-                                                type: object
-                                                properties:
-                                                    id:
-                                                        type: integer
-                                                    name:
-                                                        type: string
+                                                type: integer
                                             type:
                                                 type: string
-                                            owner:
-                                                type: string
-                                            price:
-                                                type: number
                                             count:
                                                 type: integer
         ---
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -4832,7 +4832,12 @@ class TicketViewSet(viewsets.ModelViewSet):
                                         type: object
                                         properties:
                                             event:
-                                                type: integer
+                                                type: object
+                                                properties:
+                                                    id:
+                                                        type: integer
+                                                    name:
+                                                        type: string
                                             type:
                                                 type: string
                                             count:
@@ -4873,10 +4878,10 @@ class TicketViewSet(viewsets.ModelViewSet):
                 holder__isnull=True,
             ).exclude(id__in=tickets_in_cart)[: ticket_class["count"]]
 
-            if tickets.count() < ticket_class["count"]: 
+            if tickets.count() < ticket_class["count"]:
                 sold_out_tickets.append(
                     {
-                        **ticket_class, 
+                        **ticket_class,
                         "event": {
                             "id": ticket_class["event"],
                             "name": Event.objects.get(id=ticket_class["event"]).name,
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -4871,6 +4871,7 @@ class TicketViewSet(viewsets.ModelViewSet):
         sold_out_tickets = []
         replacement_tickets = []
         tickets_in_cart = cart.tickets.values_list("id", flat=True)
+
         for ticket_class in tickets_to_replace.values("type", "event").annotate(
             count=Count("id")
         ):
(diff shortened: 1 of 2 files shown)


backend/clubs/views.py
```diff
@@ -4945,21 +4969,7 @@ def initiate_checkout(self, request, *args, **kwargs):
...
-
-        cart_total = sum(
-            (
-                ticket.price * (1 - ticket.group_discount)
-                if ticket.group_size
-                and ticket_type_counts[ticket.type] >= ticket.group_size
-                else ticket.price
-            )
-            for ticket in tickets
-        )
```
aviupadhyayula, line 4962: Was [previously broken](https://stackoverflow.com/questions/45768190/typeerror-manyrelatedmanager-object-is-not-iterable).

backend/clubs/views.py
```diff
@@ -4782,6 +4783,29 @@ class TicketViewSet(viewsets.ModelViewSet):
     http_method_names = ["get", "post"]
     lookup_field = "id"
 
+    def _calculate_cart_total(self, cart) -> float:
```
rohangpta, line 4786: Could we make this a free function? `self` isn't used in the body anywhere.

backend/tests/clubs/test_ticketing.py
```diff
@@ -525,7 +554,48 @@ def test_get_cart(self):
...
+                type="group",
+                event=self.event1,
+                price=10.0,
+                group_size=2,
+                group_discount=0.2,
+            )
+            for _ in range(10)
+        ]
+
+        # Add enough to activate group discount
```
rohangpta, line 588: Could you also assert the correct cart total before hitting the threshold for group discount?

backend/tests/clubs/test_ticketing.py
```diff
@@ -525,7 +554,48 @@ def test_get_cart(self):
         self.assertEqual(len(data["tickets"]), 5, data)
         for t1, t2 in zip(data["tickets"], tickets_to_add):
             self.assertEqual(t1["id"], str(t2.id))
-        self.assertEqual(data["sold_out"], 0, data)
+        num_sold_out = sum(item["count"] for item in data["sold_out"])
```
rohangpta, line 557: Please assert on all keys and values of `data["sold_out"]`, below as well.

backend/tests/clubs/test_ticketing.py
```diff
@@ -741,9 +814,10 @@ def test_complete_checkout(self):
             self.assertEqual(held_tickets.count(), 0, held_tickets)
 
             # Transaction record created
-            TicketTransactionRecord.objects.filter(
+            record_exists = TicketTransactionRecord.objects.filter(
                 reconciliation_id=MockPaymentResponse().reconciliation_id
             ).exists()
+            self.assertTrue(record_exists)
```
rohangpta, line 820: Nice catch!

rohangpta (commented):
Nice work! Looks like these APIs have moved but the tests haven't quite caught up - I've added a few comments around that.

backend/clubs/views.py
```diff
@@ -4802,7 +4826,7 @@ def cart(self, request, *args, **kwargs):
                                     allOf:
                                         - $ref: "#/components/schemas/Ticket"
                                 sold_out:
-                                    type: list
+                                    type: array
```
joyliu-q, line 4829: Oops thanks for the catch haha

joyliu-q (approved):
Looks great, thanks for the catches!

backend/clubs/views.py
aviupadhyayula: Made it a static method.

backend/clubs/views.py
```diff
@@ -4802,25 +4827,14 @@ def cart(self, request, *args, **kwargs):
...
-                                                        type: string
+                                                type: integer
                                             type:
                                                 type: string
-                                            owner:
-                                                type: string
-                                            price:
-                                                type: number
                                             count:
                                                 type: integer
```
aviupadhyayula, line 4839: @joyliu-q looks like you're dealing with the annotated queryset (see L4865), so these fields won't exist. Unsure if this is the behavior you intended.

julianweng: Fixing this shortly

aviupadhyayula: Thanks! Looks like there were some merge conflicts as a result of your changes. Resolved conflicts in [a9054ea](https://github.com/pennlabs/penn-clubs/pull/675/commits/a9054eab62c8e7c2556059ba5b9796fe7bf68d9a) and modified tests in [63ff452](https://github.com/pennlabs/penn-clubs/pull/675/commits/63ff45235b498aaa77026da50e7dd3fa0eb99b04).

PR #679, opened 2024-04-23
Add route for admins to issue tickets to users

aviupadhyayula:
Adds route for club admins to issue tickets to users. Tickets must already be created to be issued.

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -2791,6 +2791,164 @@ class ClubEventViewSet(viewsets.ModelViewSet):
         return Response({"detail": "Successfully created tickets"})
 
     @action(detail=True, methods=["post"])
+    @transaction.atomic
+    @update_holds
+    def issue_tickets(self, request, *args, **kwargs):
+        """
+        Issue tickets to users based on their pennkeys and ticket classes.
+        ---
+        requestBody:
+            content:
+                application/json:
+                    schema:
+                        type: object
+                        properties:
+                            quantities:
+                                type: array
+                                items:
+                                    type: object
+                                    properties:
+                                        username:
+                                            type: string
+                                        ticket_class:
+                                            type: string
+
+        responses:
+            "200":
+                content:
+                    application/json:
+                        schema:
+                            type: object
+                            properties:
+                                detail:
+                                    type: string
+            "400":
+                content:
+                    application/json:
+                        schema:
+                            type: object
+                            properties:
+                                detail:
+                                    type: string
+        ---
+        """
+        event = self.get_object()
+
+        quantities = request.data.get("quantities", [])
+
+        if not quantities:
+            return Response(
+                {"detail": "Quantities must be specified"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        for item in quantities:
+            if not item.get("username") or not item.get("ticket_class"):
+                return Response(
+                    {"detail": "Specify username and ticket class to issue a ticket"},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+
+        usernames = [item.get("username") for item in quantities]
+        ticket_classes = [item.get("ticket_class") for item in quantities]
+
+        # Check that requested usernames are valid
+        invalid_users = set(usernames) - set(
+            get_user_model()
+            .objects.filter(username__in=usernames)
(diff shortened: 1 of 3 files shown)


aviupadhyayula:
This route bypasses `event.ticket_drop_time`, so tickets that aren't publicly available yet (i.e. through `event/tickets`) can still be issued. I think this fits the spirit of why we added the drop time. 

Thoughts?

backend/clubs/views.py
```diff
@@ -2790,6 +2790,145 @@ def create_tickets(self, request, *args, **kwargs):
...
+                                detail:
+                                    type: string
+            "400":
+                content:
+                    application/json:
+                        schema:
+                            type: object
+                            properties:
+                                detail:
+                                    type: string
```
joyliu-q, line 2832: Thank you for this PR! Could the 400 detail also return an array of failed pennid strings?

aviupadhyayula: Sure, added an `errors` attribute to the response schema.

backend/clubs/views.py
```diff
@@ -2790,6 +2790,165 @@ def create_tickets(self, request, *args, **kwargs):
 
         return Response({"detail": "Successfully created tickets"})
 
+    @action(detail=True, methods=["post"])
+    @transaction.atomic
+    @update_holds
+    def issue_tickets(self, request, *args, **kwargs):
```
rohangpta, line 2796: Please lock behind same perms for `create_tickets`.

backend/clubs/views.py
```diff
@@ -2790,6 +2790,165 @@ def create_tickets(self, request, *args, **kwargs):
...
+                                    type: string
+                                errors:
+                                    type: array
+                                    items:
+                                        type: string
+        ---
+        """
+        event = self.get_object()
+
+        quantities = request.data.get("quantities", [])
```
rohangpta, line 2841: Nit: Is quantities an apt name here? We don't really have a count sub field

backend/clubs/views.py
```diff
@@ -2790,6 +2790,165 @@ def create_tickets(self, request, *args, **kwargs):
...
+            ticket = next(
+                ticket
+                for ticket in tickets
+                if ticket.type == ticket_type and ticket.owner is None
+            )
+            ticket.owner = user
+            ticket.holder = None
+
+            transaction_records.append(
+                TicketTransactionRecord(
```
rohangpta, line 2931: Would it be nice to have a "note: manually assigned here"? A null reconciliation id and 0.0 amount does tell us something was up...

backend/clubs/views.py
```diff
@@ -2790,6 +2790,165 @@ def create_tickets(self, request, *args, **kwargs):
...
+                        ),
+                        "errors": [],
+                    },
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+
+            tickets.extend(available_tickets)
+
+        # Hold all selected tickets for 10 mins
+        holding_expiration = timezone.now() + datetime.timedelta(minutes=10)
```
rohangpta, line 2912: Why do we hold here?

rohangpta (commented):
Cleanly written and well done. A few things to iron out but this is looking good to me overall. 

Permissions (and a test validating this) are the most important thing about this PR

backend/clubs/views.py
aviupadhyayula: Forgot that this was wrapped with a `transaction.atomic`, good catch.

backend/clubs/views.py
aviupadhyayula: changed it to `tickets`

backend/clubs/views.py
aviupadhyayula: I don't think it's necessary, but might be helpful (especially with the current impl of free tickets in #658, which also uses a null reconciliation ID & 0.0 amount). Happy to add a field to the model for comments. Can also hack the reconciliation ID, and set it to `issued` instead?

backend/clubs/views.py
aviupadhyayula: thanks for catching

aviupadhyayula:
> Permissions (and a test validating this) are the most important thing about this PR

Merging this in so that frontend dev isn't blocked, but main points should be addressed.

PR #680, opened 2024-04-23
Fix breaking changes to `cart` in #670

aviupadhyayula:
Also replaces previous logic to grab event names with `select_related` as per @julianweng's insistence.


joyliu-q (approved):
LGTM!

PR #683, opened 2024-04-24
Fix empty fields in generated admin reports

aviupadhyayula:
We've gotten complaints from OSA that admin reports are missing certain fields (e.g. `created_at`, `target_majors`, `accepting_members`, etc.). This PR should solve.

[without fix](https://github.com/pennlabs/penn-clubs/files/15103894/unfixed_w_all_fields.xlsx) vs [with fix](https://github.com/pennlabs/penn-clubs/files/15103896/fixed_w_all_fields.xlsx)


rm03 (approved):
Great find Avi! LGTM, tested and working locally.

PR #684, opened 2024-04-25
Add route for attendance tracking

rm03:
Allows for tracking attendance at events. Attendance can only be updated by a club officer for the ticket's associated event.


backend/tests/clubs/test_ticketing.py
```diff
@@ -1312,6 +1313,42 @@ def test_transfer_ticket_to_self(self):
...
+        )
+        ticket.refresh_from_db()
+        self.assertEqual(resp.status_code, 200, resp.content)
+        self.assertTrue(ticket.attended)
+
+    def test_update_attendance_non_officer(self):
+        self.client.login(username=self.user1.username, password="test")
+        ticket = self.tickets1[0]
+        ticket.owner = self.user1
+        ticket.save()
```
Porcupine1, line 1341: I got a little confused for a second why a difference in ticket ownership mattered, but it doesn't. In `test_update_attendance` you create a membership between `user1` and `club1` such that `user1` has officer role. In this function, no membership is given and clubs don't have any membership relations with any user from `comonSetUp`, so here `user1` is not an officer. I think having an explicit comment would prevent future confusion.

Porcupine1 (commented):
LGTM!

backend/clubs/views.py
```diff
@@ -5505,6 +5547,13 @@ def transfer(self, request, *args, **kwargs):
...
 
     def get_queryset(self):
+        if self.action == "partial_update":
+            officer_clubs = Membership.objects.filter(
+                person=self.request.user, role__lte=Membership.ROLE_OFFICER
+            ).values_list("club", flat=True)
+            return Ticket.objects.filter(event__club__in=officer_clubs).select_related(
+                "event__club"
+            )
         return Ticket.objects.filter(owner=self.request.user.id)
```
aviupadhyayula, line 5557: A bit nasty, but I guess it gets the job done. We should revisit later.

backend/clubs/views.py
```diff
@@ -5018,6 +5021,45 @@ def _calculate_cart_total(cart) -> float:
...
+        """
+        attended = request.data.get("attended")
+        if attended is None or not isinstance(attended, bool):
+            return Response(
+                {"detail": "Missing boolean attribute 'attended'."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        ticket = self.get_object()
+        ticket.attended = attended
+        ticket.save()
```
aviupadhyayula, line 5060: Should we allow officers to mark tickets that have been claimed as unclaimed? Don't have strong opinions but would like to hear what you think.

backend/tests/clubs/test_ticketing.py
```diff
@@ -1312,6 +1313,42 @@ def test_transfer_ticket_to_self(self):
...
+
+        resp = self.client.patch(
+            reverse("tickets-detail", args=(ticket.id,)),
+            {"attended": True},
+            format="json",
+        )
+        ticket.refresh_from_db()
+        self.assertEqual(resp.status_code, 404, resp.content)
```
aviupadhyayula, line 1349: Feels weird that this is a 404 instead of a 403 😔

aviupadhyayula (commented):
LGTM, great work Moniz.

backend/clubs/views.py
rm03: I imagine there will be a toggle for attendance for each ticket on the frontend in case officers don't want to scan QR codes, in which case we should allow this

aviupadhyayula: Will there ever be a case where a ticket that's been marked as attended needs to be unmarked? If not, let's add a `dry_run` attribute or something. I don't think we should expose functionality that we never want used.

backend/clubs/views.py
rm03: We'll need something similar for the QR code validation route, since that should also be accessible by club officers + ticket owner

backend/tests/clubs/test_ticketing.py
rm03: Yeah, it should ideally be a 403 but the logic ends with the queryset

backend/clubs/views.py
aviupadhyayula: Should we do something similar to how perms work for `create_tickets`?

PR #690, opened 2024-04-29
Fix ticketing event creation and image on event page

julianweng:
Fixes issue of creating tickets immediately following event creation by calling onChange on object creation for ModelForm. Also fixes image dimension issues on new event page.


esinx (approved):

esinx:
note: We might want to redesign `ModelForm` completely.

It's
- Not typesafe
- Relies on backend data structure rather than frontend explicitly defining the data type. This is bad if we want custom validation logic with awareness of what the type of actual object is.

A good alternative would be a form builder component using [react-hook-form](https://react-hook-form.com/)

Might be a good (chanllenging) newbie project

joyliu-q (approved):
LGTM

Issue #688, opened 2024-04-29
Create indexes and (optionally) filter on event for tickets


rohangpta:
- Create an index on holding expiration and maybe the event to easily filter tickets
- This will make the `update_holds` function substantially cheaper

aviupadhyayula:
Looks like Django creates an index on event [by default](https://stackoverflow.com/questions/5984842/does-django-automatically-generate-indexes-for-foreign-keys-columns), since it's a foreign key.

PR #691, opened 2024-04-29
Move TicketTransactionRecord back to field on Ticket

rm03:
To enforce the constraint that a `Ticket` can't be deleted if an associated `TicketTransactionRecord` or `TicketTransferRecord` exists via `on_delete=models.PROTECT`, we moved `Ticket` to be a foreign key on `TicketTransactionRecord`. However, this results in the relationship between `Ticket`s and `TicketTransactionRecord`s being 1:1 instead of N:1 which doesn't make much sense. 

This PR reverts the change. Instead, to enforce the desired behavior, the `delete()` method for the `Ticket` model and queryset are overridden.


Porcupine1 (commented):
lgtm!

rohangpta (approved):
LGTM. Thanks for adding tests as well.

aviupadhyayula:
Added minor changes to `update_holds`, along with tests to verify functionality. Also added an index on `holding_expiration` (closes #688). 

Sorry for hijacking the PR 🙂

PR #692, opened 2024-04-30
Fix QR codes in confirmation email

julianweng:
Adds qr code image as attachment in confirmation email.

Embeds image using now-public (temporary, since ticket ids are insecure currently regardless) `/tickets/{ticketid}/qr` route, as current method of embedding (data uris) does not work with Gmail clients.

### DIFF
--- a/backend/clubs/models.py
+++ b/backend/clubs/models.py
@@ -111,6 +111,15 @@ def send_mail_helper(name, subject, emails, context, attachment=None):
                 "application/vnd.openxmlformats-officedocument."
                 + "wordprocessingml.document",
             )
+    elif (
+        attachment is not None and "filename" in attachment and "content" in attachment
+    ):
+        msg.attach(
+            attachment["filename"],
+            attachment["content"],
+            # Assume that we just use this for PNGs for now
+            "image/png",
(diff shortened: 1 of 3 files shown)


backend/clubs/models.py
```diff
@@ -1895,19 +1904,20 @@ def send_confirmation_email(self):
         Send a confirmation email to the ticket owner after purchase
         """
         owner = self.owner
+        qr_url = f"https://{settings.DOMAINS[0]}/api/tickets/{self.id}/qr"
 
         output = BytesIO()
         qr_image = self.get_qr()
         qr_image.save(output, format="PNG")
-        decoded_image = base64.b64encode(output.getvalue()).decode("ascii")
+        image = base64.b64encode(output.getvalue())
```
anli5005, line 1912: Is there a reason we're still base64 encoding this...? Wouldn't `msg.attach` abstracts that away in a sane universe?

anli5005:
> Anthony! We aren't doing this PR anymore.

Then why don't you close it?

> I don't know, let me do that right now - @julianweng

julianweng:
#693

PR #695, opened 2024-04-30
Make multiple owned tickets UI more intuitive

julianweng:
image

- Removes action buttons when ticket stack for given event is collapsed.
- Adds numbering to ticket when stack is collapsed and has multiple tickets in it
- Removes numbering for "stacks" with only one ticket in it

Attempts to make it clearer when multiple tickets are owned that the stack is expandable, differentiating it visually from single owned tickets.


aviupadhyayula (commented):
LGTM. Also, would it be difficult to put an "Expand" button in the top right that expands the stack?

julianweng:
Can do

julianweng:
@aviupadhyayula  
image

aviupadhyayula (approved):
Great, thanks. LGTM

PR #697, opened 2024-05-03
Set reply-to to OSA email for approval-related notification emails

julianweng:
The majority of replies to our automated emails related to club approval/re-approval are best handled by OSA (e.g. "When would X club be approved?"). CC'ing them for each email (so recipients can reply all) is unnecessary and asking recipients to email OSA instead of replying might be unreliable: keeping the instruction to reply to the email but setting reply-to to include OSA seems like a reasonable change.

### DIFF
--- a/backend/clubs/models.py
+++ b/backend/clubs/models.py
@@ -52,7 +52,7 @@ def get_mail_type_annotation(name):
     return None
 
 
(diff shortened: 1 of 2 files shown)


aviupadhyayula (approved):
LGTM, thanks Julian.

rm03 (approved):
lgtm, tested and working

PR #703, opened 2024-06-03
Add tests left out from ticketing

### LINKED ISSUE #702: Add tests left out in #658
Test for `_place_hold_on_tickets` and `_give_tickets` were left out in #658. The PR adds these tests.

Porcupine1:
Resolves #702

### DIFF
--- a/backend/tests/clubs/test_ticketing.py
+++ b/backend/tests/clubs/test_ticketing.py
@@ -955,8 +955,7 @@ class TicketTestCase(TestCase):
         # Add a few tickets
         cart, _ = Cart.objects.get_or_create(owner=self.user1)
         tickets_to_add = self.tickets1[:3]
-        for ticket in tickets_to_add:
-            cart.tickets.add(ticket)
+        cart.tickets.add(*tickets_to_add)
         cart.save()
 
         TicketViewSet._place_hold_on_tickets(self.user1, cart.tickets)
@@ -986,8 +985,7 @@ class TicketTestCase(TestCase):
         # Add a few tickets
         cart, _ = Cart.objects.get_or_create(owner=self.user1)
         tickets_to_add = self.tickets1[:3]
-        for ticket in tickets_to_add:
-            cart.tickets.add(ticket)
+        cart.tickets.add(*tickets_to_add)
         cart.save()
 
         order_info = {
--- a/backend/tests/clubs/test_ticketing.py
+++ b/backend/tests/clubs/test_ticketing.py
@@ -947,6 +947,72 @@ class TicketTestCase(TestCase):
         to_add = set(map(lambda t: str(t.id), tickets_to_add))
         self.assertEqual(len(in_cart & to_add), 0, in_cart | to_add)
 
+    def test_place_hold_on_tickets(self):
+        from clubs.views import TicketViewSet
+
+        self.client.login(username=self.user1.username, password="test")
+
+        # Add a few tickets
+        cart, _ = Cart.objects.get_or_create(owner=self.user1)
+        tickets_to_add = self.tickets1[:3]
+        for ticket in tickets_to_add:
+            cart.tickets.add(ticket)
+        cart.save()
+
+        TicketViewSet._place_hold_on_tickets(self.user1, cart.tickets)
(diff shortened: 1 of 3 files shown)


backend/tests/clubs/test_ticketing.py
```diff
@@ -947,6 +947,72 @@ def test_get_cart_replacement_required_sold_out(self):
...
+                delta=timedelta(seconds=10),
+            )
+
+        # Move Django's internal clock 10 minutes forward
+        with freezegun.freeze_time(holding_expiration):
+            Ticket.objects.update_holds()
+            for ticket in cart.tickets.all():
+                self.assertIsNone(ticket.owner)
+                self.assertIsNone(ticket.holder)
+                # TODO: Ask whether to reset holding expiration to None
```
aviupadhyayula, line 980: Can remove this.

aviupadhyayula: Separately, this decision's up to you. We didn't null it because there won't be a case where we rely on `holding_expiration` without checking `holder!=None`, but it is a bit unsafe.

backend/tests/clubs/test_ticketing.py
```diff
@@ -947,6 +947,72 @@ def test_get_cart_replacement_required_sold_out(self):
...
+    def test_place_hold_on_tickets(self):
+        from clubs.views import TicketViewSet
+
+        self.client.login(username=self.user1.username, password="test")
+
+        # Add a few tickets
+        cart, _ = Cart.objects.get_or_create(owner=self.user1)
+        tickets_to_add = self.tickets1[:3]
+        for ticket in tickets_to_add:
+            cart.tickets.add(ticket)
```
aviupadhyayula, line 959: nit: can use `cart.tickets.add(*tickets_to_add)`

backend/tests/clubs/test_ticketing.py
```diff
@@ -947,6 +947,72 @@ def test_get_cart_replacement_required_sold_out(self):
...
+
+    def test_give_tickets(self):
+        from clubs.views import TicketViewSet
+
+        self.client.login(username=self.user1.username, password="test")
+        # Add a few tickets
+        cart, _ = Cart.objects.get_or_create(owner=self.user1)
+        tickets_to_add = self.tickets1[:3]
+        for ticket in tickets_to_add:
+            cart.tickets.add(ticket)
```
aviupadhyayula, line 990: nit: same as above

backend/tests/clubs/test_ticketing.py
```diff
@@ -947,6 +947,72 @@ def test_get_cart_replacement_required_sold_out(self):
...
+                "lastName": self.user1.last_name,
+                "phoneNumber": None,
+                "email": self.user1.email,
+            },
+        }
+
+        TicketViewSet._place_hold_on_tickets(self.user1, cart.tickets)
+        TicketViewSet._give_tickets(
+            self.user1, order_info, cart, reconciliation_id=None
+        )
```
aviupadhyayula, line 1006: Maybe also check for the creation of transaction records? Let me know what you think.

aviupadhyayula (approved):
Great work, left a few comments. Thanks for taking the initiative with this -- tests look super clean 🚀

backend/tests/clubs/test_ticketing.py
Porcupine1: I remember it being tested.
lines 1159 and 1338

aviupadhyayula: Yup, but IMO it's still good practice to move those tests over. Those tests were written back when the "give tickets" functionality wasn't its own function. Now that it is, unit tests for `_give_tickets` should test all its intended functionality. It's not needed here, but something to keep in mind with future unit tests.

backend/tests/clubs/test_ticketing.py
aviupadhyayula: Was the `TODO` comment removed?

backend/tests/clubs/test_ticketing.py
Porcupine1: I see, will apply changes

PR #704, opened 2024-06-06
Disable queue for summer

julianweng:
Per OSA request, disables editing of club fields that would necessitate queue approval as well as daily email notifications for queue approval.

### DIFF
--- a/backend/tests/clubs/test_views.py
+++ b/backend/tests/clubs/test_views.py
@@ -3,6 +3,7 @@ import io
 import json
 import os
 from collections import Counter
+from unittest import mock
 from unittest.mock import MagicMock, patch
 
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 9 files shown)


backend/tests/clubs/test_views.py
```diff
@@ -3,6 +3,7 @@
 import json
 import os
 from collections import Counter
+from unittest import mock
```
rm03, line 6: Is this needed for anything other than `mock.patch()`? If so we already import that

rm03 (approved):
Thanks Julian, good work :)

frontend/utils/branding.tsx
```diff
@@ -252,6 +252,7 @@ const sites = {
 }
 
 export const TICKETING_PAYMENT_ENABLED = false
+export const QUEUE_ENABLED = false
```
aviupadhyayula, line 255: Is it possible to pull this from the backend settings? Not a huge deal if not.

backend/tests/clubs/test_views.py
julianweng: Good catch, thanks!

frontend/utils/branding.tsx
julianweng: Couldn't find a way from first glance. Could potentially use env variables maybe instead of separate settings, but would be more elegant in the long term to implement some sort of global variables table so admin etc. can also toggle queue + other features with a gui.

PR #705, opened 2024-06-20
Add club diff feature

Porcupine1:
This pull request introduces a new feature aimed at enhancing the admin experience by providing a concise display of changes made to a club's information when it is re-submitted for approval. This feature addresses the current challenge where admins have to manually compare and identify changes by reviewing the entire club page, which is not ideal.


backend/clubs/models.py
```diff
@@ -698,6 +698,31 @@ def send_approval_email(self, request=None, change=False):
                 reply_to=settings.OSA_EMAILS + [settings.BRANDING_SITE_EMAIL],
             )
 
+    def get_latest_and_latest_approved_versions(self):
+        """
+        Returns the latest and latest approved versions of club key fields
+        [name, description, image] contents that trigger reapproval.
+        """
+        latest_approved_version = (
+            self.history.filter(approved=True).order_by("-history_date").first()
```
aviupadhyayula, line 707: [Looks like the `history_date` field is indexed](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://django-simple-history.readthedocs.io/en/latest/historical_model.html%23:~:text%3DBy%2520default%2520the%2520history_date%2520field,this%2520behavior%2520using%2520settings.py.&ved=2ahUKEwjdz7njqeuGAxVMMVkFHajuCxEQFnoECBIQAw&usg=AOvVaw3Kwcm4ucYvSbX7gCG3V-OX), in which case it'd be cheaper to order first and filter second. I'm not sure if the ORM does this under the hood, but it's worth looking into the queries being generated. [This could be a good starting point](https://b0uh.github.io/django-show-me-the-sql.html).

backend/clubs/models.py
```diff
@@ -698,6 +698,31 @@ def send_approval_email(self, request=None, change=False):
...
+        )
+        latest_version = self.history.order_by("-history_date").first()
+
+        if (
+            not latest_version
+            or not latest_approved_version
```
aviupadhyayula, line 713: Like Julian mentioned, this won't work if the club is brand new. We still want to display a diff (I think, I'll defer to @julianweng on this) for clubs that are registering for the first time.

backend/clubs/models.py
```diff
@@ -698,6 +698,31 @@ def send_approval_email(self, request=None, change=False):
                 reply_to=settings.OSA_EMAILS + [settings.BRANDING_SITE_EMAIL],
             )
 
+    def get_latest_and_latest_approved_versions(self):
+        """
+        Returns the latest and latest approved versions of club key fields
+        [name, description, image] contents that trigger reapproval.
+        """
```
aviupadhyayula, line 705: Off topic: is it possible to limit how many historical records we store? Creating a new record for every change sounds expensive. I wonder if there's a way to limit it, e.g. only keep the last "approved" version of the club onwards.

backend/clubs/models.py
```diff
@@ -698,6 +698,31 @@ def send_approval_email(self, request=None, change=False):
...
+            self.history.filter(approved=True).order_by("-history_date").first()
+        )
+        latest_version = self.history.order_by("-history_date").first()
+
+        if (
+            not latest_version
+            or not latest_approved_version
+            or (latest_approved_version == latest_version)
+        ):
+            return {}
```
aviupadhyayula, line 716: Can't we just check: "is the club currently not approved, and did there exist a previous version that was approved".

aviupadhyayula (commented):
Great work Thomas! Left a couple comments. We'll definitely need thorough unit tests for this, but should be good to go after that.

backend/clubs/models.py
```diff
@@ -698,6 +698,31 @@ def send_approval_email(self, request=None, change=False):
...
+
+        diff = latest_version.diff_against(latest_approved_version)
```
aviupadhyayula, line 718: If we're only reporting name, description, and image changes, do we need to diff every field? [Hint](https://django-simple-history.readthedocs.io/en/latest/history_diffing.html#history-diffing)

julianweng: Yep, in retrospect I'm not sure if this line is strictly necessary in the first place if we're computing diffs on the frontend (as discussed) anyways. We could compute on the backend but it would require more detailed serialization and not necessarily be much more performant at this scale.

backend/clubs/models.py
julianweng: Was thinking about maybe storing the last k (maybe 3?) approved versions and pruning older ones on insert, since there isn't any need to store previous unapproved versions, but having a brief history of approvals could be marginally useful given that there are currently multiple admins who can approve clubs, and to account for accidental approvals etc.

aviupadhyayula: Sounds reasonable. We're already using the historical records in `ClubListSerializer.to_representation()`, to show the last approved version of a club to non-members. Agree that pruning older records would be great, but we're not introducing extra storage costs with this PR. I think it's safe to push it down the line.

backend/clubs/models.py
aviupadhyayula: Yeah, agreed. I'm sure React has some great diff libraries. I think it'd be cleaner to expose something that fetches the most recently approved version of a club, and we can do diffing on the frontend. 

I'm also not sure where the `pending_clubs` endpoint would be used. We already know which clubs we want to diff (e.g. the ones in the approval queue), and I can't see where we'd use this outside of the queue. 

@Porcupine1 it'd be helpful to iron out the details with Julian further.

backend/clubs/models.py
Porcupine1: Oh, I thought filtering first reduces the amount of data to order, making it cheaper?

backend/clubs/models.py
Porcupine1: I see

backend/clubs/models.py
Porcupine1: I remember we discussed that initially we were just the calling the clubs viewset list function with the approved = false and active = true query/search params. I added `pending_clubs` because we are now returning pending clubs data

backend/clubs/models.py
Porcupine1: Looks like Django filters first

backend/clubs/models.py
```diff
@@ -698,6 +698,37 @@ def send_approval_email(self, request=None, change=False):
...
+        """
+        latest_approved_version = (
+            self.history.filter(approved=True).order_by("-history_date").first()
+        )
+        latest_version = self.history.order_by("-history_date").first()
+
+        # if this is the first time the club is being approved
+        if not latest_approved_version:
+            return {
+                self.code: {
```
rm03, line 714: would prefer to add the club code to the response in the view instead of this helper

backend/clubs/views.py
```diff
@@ -2058,6 +2058,70 @@ def perform_destroy(self, instance):
             context=context,
         )
 
+    @action(detail=False, methods=["GET"])
```
rm03, line 2061: Can we make this an instance-specific route? Ideally, it will be accessed when OSA is viewing a specific club's page and wants to see the diff between the latest and latest approved versions.

rm03 (changes requested):

backend/clubs/models.py
Porcupine1: I think if we make the route instance-specific, we won't require the club code. With this new implementation, it looks like the approval queue list will remain as is, only that the links to a club in the queue are to `club\code\diff`

aviupadhyayula: I don't think we'll point to a new page for clubs in the queue (e.g. right now clicking on a club in the queue goes to `/api/clubs/<code>`). 

Instead, the idea is to expose an endpoint `/diff` that returns some historical comparison (either the last approved version, or the diffs themselves). We'll then add a param when an admin clicks on a club in the queue (e.g. `/api/clubs/<code>?review=True`) that triggers some frontend logic. The frontend logic is what's going to call `/api/clubs/<code>/diff` and display the results in some fashion. 

@julianweng @rm03 let me know if that's what we were thinking here.

Porcupine1: I see, that makes sense. I was actually thinking that there would be code repetition (with `/api/clubs/<code>` route) if implemented the way I described it earlier

aviupadhyayula:
Thanks for your help with this feature Thomas! Closing since #811 incorporates these changes.

PR #707, opened 2024-06-20
Resolve unaddressed nits in #703

Porcupine1:
-Remove TODO
-Test that transaction record create when `_give_tickets()` is called

### DIFF
--- a/backend/clubs/models.py
+++ b/backend/clubs/models.py
@@ -1885,8 +1885,8 @@ class Ticket(models.Model):
     group_size = models.PositiveIntegerField(null=True, blank=True)
     transferable = models.BooleanField(default=True)
     attended = models.BooleanField(default=False)
-    buyable = models.BooleanField(default=True)
     # TODO: change to enum between All, Club, None
+    buyable = models.BooleanField(default=True)
     transaction_record = models.ForeignKey(
         TicketTransactionRecord,
         related_name="tickets",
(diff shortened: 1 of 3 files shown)


backend/clubs/models.py
```diff
@@ -1885,7 +1885,6 @@ class Ticket(models.Model):
     group_size = models.PositiveIntegerField(null=True, blank=True)
     transferable = models.BooleanField(default=True)
     attended = models.BooleanField(default=False)
-    # TODO: change to enum between All, Club, None
```
aviupadhyayula, line 1888: I think you can keep this TODO haha, it's something we'll get around to in the future.

backend/tests/clubs/test_ticketing.py
```diff
@@ -947,6 +947,78 @@ def test_get_cart_replacement_required_sold_out(self):
...
+
+        # Check that tickets are assigned their owner
+        for ticket in cart.tickets.all():
+            self.assertEqual(self.user1, ticket.owner)
+            self.assertIsNone(ticket.holder)
+
+        # Check that the cart is empty
+        self.assertEqual(0, cart.tickets.count())
+
+        # Transaction record created
```
aviupadhyayula, line 1016: Nit: should be "Check that transaction record is created".

aviupadhyayula (commented):
Thanks for circling back to these. In the future, just make a new PR/commit to address comments given after a PR's already been merged.

backend/clubs/models.py
```diff
@@ -1886,6 +1886,7 @@ class Ticket(models.Model):
     transferable = models.BooleanField(default=True)
     attended = models.BooleanField(default=False)
     buyable = models.BooleanField(default=True)
+    # TODO: change to enum between All, Club, None
```
aviupadhyayula, line 1889: Last nit: this should be before L1888, since it's referring to `buyable`.

PR #709, opened 2024-08-17
Remove error details for hidden user profiles

julianweng:
Currently, for signed in users, visiting a user profile set to private yields a slightly different page compared to profiles that don't exist, which implicates the existence or nonexistence of any given Pennkey and associated email at `/user/{pennkey}`. This would theoretically let someone scrape the vast majority of Penn emails in existence through brute force, including those hidden from the Penn directory and email autocompletes.

I'm not sure if this is necessary or outweighs the marginal benefit of the error detail, and we should discuss separately, but am opening a PR since it's a one line change.


Clue88:
I think this makes sense to change, lgtm

Clue88 (approved):

aviupadhyayula (approved):
LGTM, thanks Julian. I remember OSA being concerned about these sort of attacks. This is a great find.

rm03 (approved):

PR #710, opened 2024-08-21
Add support for non-virtual fairs

julianweng:
Conditionally removes copy related specifically to virtual club fairs to support the usage of Penn Clubs fairs for solely informational purposes without using the remaining (mostly virtual) functionality.

Creates default help page for club fairs that uses the provided descriptions instead of hardcoded information.

We should also add the option to toggle between external registration and Clubs-managed registration (opt out of the latter) for the fairs at some point but these changes are higher priority given the incorrect banner on the homepage.


frontend/components/ClubPage/LiveEventsDialog.tsx
```diff
@@ -121,8 +123,8 @@ const LiveEventsDialog = ({
       </Link>
       <LiveTitle>{fairName}</LiveTitle>
       <LiveSub>
-        {liveEventCount === 0 ? (
-          `Get ready for the virtual ${FAIR_NAME} fair! If you need help registering email contact@pennclubs.com`
```
aviupadhyayula, line 125: Why are we taking out the contact email? Can we point them to OSA instead?

frontend/pages/fair.tsx
```diff
@@ -78,29 +74,43 @@ const FairPage = ({
...
-          College Green from 12p-4p each day.
-        </p>
-        <p>
-          To participate, sign-up will be facilitated through Penn Clubs and
-          will coincide with the annual club registration process. Only
-          returning undergraduate student-run groups that were registered in
-          Penn Clubs last year are eligible to sign-up for the Fall Activities
-          Fair.
+          The {fair?.name} sponsored by {fair?.organization}, will be held from{' '}
+          {fair?.start_time.split('T')[0]} to {fair?.end_time.split('T')[0]}!
```
aviupadhyayula, line 78: Nit: can we throw in defaults here, in case `fair` is null?

aviupadhyayula (commented):
Thanks for picking this up Julian, LGTM. If you have time, it'd be nice to throw in a quick unit test for creating a virtual fair.

rm03 (approved):
Looks good to me, thanks Julian

julianweng:
Thanks for the review, I added the private option to the existing fair creation test, but there isn't any difference in backend behavior caused by it, so would like to condition a separate unit test on that being the case.

PR #711, opened 2024-08-24
Move club cards display to masonry layout

aviupadhyayula:
The front page currently displays club cards in two columns, with each "row" of cards beginning at the same point. This creates gaps when one column contains cards with longer lengths than the other. This PR moves to a [masonry layout](https://www.smashingmagazine.com/native-css-masonry-layout-css-grid/).

Here's a before vs after comparison. Before: 
Screenshot 2024-08-23 at 6 32 27 PM
After:
Screenshot 2024-08-23 at 6 32 34 PM


frontend/components/ClubDisplay.tsx
```diff
@@ -50,11 +70,18 @@ const ClubDisplay = ({
 
   if (display === 'cards') {
     return (
-      <div className="columns is-multiline is-desktop is-tablet">
+      <StyledMasonry
+        breakpointCols={{
+          default: 2,
+          960: 1,
```
julianweng, line 76: Ideally use some existing constant to define XL screen sizes

frontend/components/ClubDisplay.tsx
```diff
@@ -50,11 +70,18 @@ const ClubDisplay = ({
...
+        }}
+        className="masonry-grid"
```
julianweng, line 78: Is the classname for the parent Masonry component ever used?

julianweng (changes requested):
Thanks Avi! Using a pure CSS library like this that doesn't re-order to optimize for height is probably the right move from a performance standpoint as well as maintaining rankings. Made some nits.

PR #713, opened 2024-08-28
Don't display duplicate badges and tags

aviupadhyayula:
There are sometimes collisions between tags and badges on a club's main page. (See the attached file for an example.) This PR adds filtering logic on the frontend to 1) only display one badge per unique label 2)  in the event of a conflict between a badge and a tag, display the badge. 

Credit to @jamesdoh0109 for spotting this. 
Screenshot 2024-08-28 at 6 15 32 PM


aviupadhyayula:
Should probably be using some form of caching here (`useMemo` maybe?).

PR #714, opened 2024-08-30
Add the ability to display notification banners for clubs with given badges

julianweng:
Allows organizations such as WC and SAC to display messages on the pages of affiliated clubs.


Porcupine1 (commented):
LGTM!

PR #715, opened 2024-09-01
Fix permission-related display issues after club deactivation

aviupadhyayula:
Deactivated clubs are displayed differently to permissioned vs non-permissioned users. Users without permissions are expected to see the last approved version; users with permissions see the most up-to-date version. Currently, caching in the viewset incorrect caches the non-permissioned state and displays it to permissioned users. This leads to approval messages showing up in the club management dashboard, despite the club not being approved. 

This PR adds a test to check this case (should fail with current code), and adds a fix. It also turns on caching in CI.

### DIFF
--- a/backend/pennclubs/settings/ci.py
+++ b/backend/pennclubs/settings/ci.py
@@ -11,7 +11,7 @@ TEST_OUTPUT_VERBOSE = 2
 TEST_OUTPUT_DIR = "test-results"
 
 # Use dummy cache for testing
-CACHES = {"default": {"BACKEND": "django.core.cache.backends.dummy.DummyCache"}}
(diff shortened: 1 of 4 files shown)


backend/pennclubs/settings/ci.py
```diff
@@ -11,7 +11,7 @@
 TEST_OUTPUT_DIR = "test-results"
 
 # Use dummy cache for testing
-CACHES = {"default": {"BACKEND": "django.core.cache.backends.dummy.DummyCache"}}
+CACHES = {"default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"}}
```
aviupadhyayula, line 14: Caching was effectively turned off in CI/CD tests -- not sure what the original intention was, but I'm reverting it back. We interact with the cache manually in a lot of our views, so feels reasonable that the unit tests also check its correctness.

PR #718, opened 2024-09-05
Add Submissions

owlester12:


frontend/package.json
```diff
@@ -117,4 +117,4 @@
   "engines": {
     "node": "^20.0.0"
   }
-}
\ No newline at end of file
+}
```
julianweng, line 120: Is this necessary to pass lint?

frontend/pages/settings.tsx
```diff
@@ -31,11 +35,18 @@ const Notification = styled.span`
 `
 
 type SettingsProps = {
-  userInfo: UserInfo
+  userInfo?: UserInfo
   authenticated: boolean | null
+  submissions: ApplicationSubmission[]
+  whartonapplications: any
```
julianweng, line 41: Nit: camel case variable name and adding types would be nice (I know the reference code didn't have either but it would be good practice)

frontend/pages/settings.tsx
```diff
@@ -68,6 +79,16 @@ const Settings = ({ userInfo, authenticated }: SettingsProps) => {
...
     },
+    {
+      name: 'submissions',
+      label: 'Submissions',
+      content: <SubmissionsPage initialSubmissions={submissions} />,
+    },
+    {
+      name: 'applications',
+      label: 'Applications',
+      content: <ApplicationsPage whartonapplications={whartonapplications} />,
```
julianweng, line 90: Now that the page is more visible in the profile, would be nice to clarify that it's for Wharton Cycle applications specifically (maybe change it to all eventually but not needed now).

julianweng (changes requested):
Looks good generally, had a couple of nits and things that could be changed from the existing ported implementation. Thanks for taking the time to revisit this!

frontend/package.json
owlester12: Yes lmao

frontend/components/Applications.tsx
```diff
@@ -103,41 +103,44 @@ function ApplicationsPage({ whartonapplications }): ReactElement {
...
-                          __html: application.description,
-                        }}
-                      ></DescriptionWrapper>
-                    )}
-                </Card>
-              </Link>
-            </CardWrapper>
-          ))
+          <div>
+            <Text>Only Wharton applications displayed here!</Text>
```
julianweng, line 107: Maybe rephrase as "Note: only current Wharton applications are displayed on this page"

frontend/pages/settings.tsx
```diff
@@ -87,7 +87,7 @@ const Settings = ({
     {
       name: 'applications',
       label: 'Applications',
-      content: <ApplicationsPage whartonapplications={whartonapplications} />,
+      content: <ApplicationsPage whartonapplications={whartonApplications} />,
```
julianweng, line 90: Could be appropriate to also rename the prop in ApplicationsPage to be camelcase for simplicity

julianweng (changes requested):
Thanks for taking the time to fix the nits, just left two more and it should be ready!

frontend/components/Applications.tsx
owlester12: Made changes

frontend/pages/settings.tsx
owlester12: Made changes

julianweng (changes requested):
Hi, thanks for making the changes, can you fix the type check errors (tsc) with your change? Also, there is probably a better way of handling the rename than declaring and immediately spreading a separate returner variable that would be clearer to read.

owlester12:
Both issues should fixed. Some of the checks are still failing– something about the Codecov report not uploading? I'm unsure what that means.

PR #719, opened 2024-09-06
Display apply button to unauthenticated users

rm03:
Currently, if a user is unauthenticated, the apply button does not display even if the club's application is open (leading to some confusion). Additionally, enforces auth for viewing club applications.


julianweng (approved):
LGTM

PR #721, opened 2024-09-10
Show newest application cycle first in edit club view

Clue88:
## Context
Currently, the Manage Clubs > Applications Page view defaults to the first application cycle created for each club (see below). This is not ideal because when club leaders wish to see their applications page, it is almost always in order to view the most recent application, and creates a slower user flow because we first load all of the applications from the oldest application cycle.

Screenshot 2024-09-09 at 11 31 12 PM

## Changes
This PR reverses the order of applications shown on the Applications Page tab and sets the default application to be the most recent cycle.

Screenshot 2024-09-09 at 11 41 59 PM


aviupadhyayula (approved):
LGTM Chris, thanks for making the change!

PR #722, opened 2024-09-10
Fix staleness bug for application export

julianweng:
Currently, the CSV export for applications relies on its own (Django built-in) cache, which results in staleness. This is especially an issue when club leaders export applications shortly (< 2 hours) after an application is due, as the export would not include many application submissions that were recently submitted.

This PR makes the route use the viewset list() function, which has valid manual caching and related invalidations on application submission. It also refactors list() to cache the non-serialized data to facilitate this.

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -6700,26 +6700,25 @@ class ApplicationSubmissionViewSet(viewsets.ModelViewSet):
         """
 
         app_id = self.kwargs["application_pk"]
+        serializer = kwargs.get("serializer", self.get_serializer_class())
         key = f"applicationsubmissions:{app_id}"
 
         cached = cache.get(key)
         if cached is not None:
-            return Response(cached)
+            return Response(serializer(cached, many=True).data)
         else:
-            serializer = self.get_serializer_class()
             qs = self.get_queryset()
+            cache.set(key, qs, 60 * 60)
             data = serializer(qs, many=True).data
-            cache.set(key, data, 60 * 60)
 
         return Response(data)
 
-    @method_decorator(cache_page(60 * 60 * 2))
     @action(detail=False, methods=["get"])
     def export(self, *args, **kwargs):
         """
         Given some application submissions, export them to CSV.
 
-        Cached for 2 hours.
+        Uses potentially cached data from the list method.
         ---
         requestBody:
             content:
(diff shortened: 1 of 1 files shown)


backend/clubs/views.py
```diff
@@ -6700,26 +6700,25 @@ def list(self, *args, **kwargs):
...
         key = f"applicationsubmissions:{app_id}"
 
         cached = cache.get(key)
         if cached is not None:
-            return Response(cached)
+            return Response(serializer(cached, many=True).data)
         else:
-            serializer = self.get_serializer_class()
             qs = self.get_queryset()
+            cache.set(key, qs, 60 * 60)
```
aviupadhyayula, line 6711: Is there a reason we're caching the queryset instead of the data?

aviupadhyayula (commented):
Thanks for taking the initiative Julian. Left a comment.

backend/clubs/views.py
julianweng: We only get the data after running the queryset through the serializer, while I wanted the cache contents itself to be serializer agnostic given that export uses a different serializer. On second thought though, get_queryset() might be evaluated lazily so I would need to rethink this.

julianweng: Do you think .objects.all() would work @aviupadhyayula?

aviupadhyayula: [From what I've read](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://betterprogramming.pub/understanding-django-database-querysets-and-its-optimizations-1765cb9c36e5%23:~:text%3DPickling%252FCaching,of%2520model%2520objects%2520from%2520cache.&ved=2ahUKEwi4zYmj-biIAxWEEFkFHQbLCCEQFnoECBYQAw&usg=AOvVaw25zhDELVWq2rqCDrZSCpg7), it looks like the Django caching library forces evaluation upon insertion. We can test this later (force Django to log each query evaluation in settings). I'm more concerned about breaking our pattern of not caching querysets anywhere.

Honestly, is there a reason to cache this view? Querying the list of applicants will be done very rarely. Some latency on invocation is preferable to leaving a couple applicants off the UI or downloaded CSV.

julianweng: That's fair, caching definitely isn't necessary, at least for the CSV export (I didn't know that clubs used this feature until very recently). However, the cache invalidation logic was already previously written for the list view, and application-related views also contribute to site-level performance (I'm not sure how strict our transaction isolation levels are if at all, but you could imagine that querying all applications repeatedly, especially while new upsert transactions could still occur, is bad during application season).

julianweng:
Disabling caching entirely in #774

PR #724, opened 2024-09-15
Check for empty responses before application submission

aviupadhyayula:
Currently, users can hit the submit button with questions that don't have an answer. This PR adds a check over all questions in an application before calling the submit route.


aviupadhyayula:
Good catch. Closing as there's no trivial way to check for this without substantively changing the application question creation flow (e.g. club leaders being able to mark certain questions as optional vs not).

PR #726, opened 2024-09-22
Bump React Check for Codecov

joyliu-q:


julianweng:
Missed this my bad, frontend check is unfortunately not a thing anymore 😔

PR #730, opened 2024-09-28
Move graduate_users script to cronjob

rm03:
The `graduate_users` script, which deactivates memberships for users who graduated before the current year, was originally intended to be executed manually via the admin console. Since this task should occur automatically at the beginning of each year, it makes more sense to have the script as a cronjob.

### DIFF
--- a/k8s/main.ts
+++ b/k8s/main.ts
@@ -97,7 +97,7 @@ export class MyChart extends PennLabsChart {
     });
 
     new CronJob(this, 'graduate-users', {
-      schedule: cronTime.everyYearIn(1, 1, 12, 0),
+      schedule: cronTime.everyYearIn(5, 31, 12, 0),
       image: backendImage,
       secret: clubsSecret,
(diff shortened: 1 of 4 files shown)


aviupadhyayula (commented):
LGTM, thanks Rohan. Would love to see a unit test in `test_commands` if possible.

k8s/main.ts
```diff
@@ -95,6 +95,13 @@ export class MyChart extends PennLabsChart {
       secret: clubsSecret,
       cmd: ["python", "manage.py", "expire_membership_invites"],
     });
+
+    new CronJob(this, 'graduate-users', {
+      schedule: cronTime.everyYearIn(1, 1, 12, 0),
```
aviupadhyayula, line 100: I don't think this is right. `graudate_users` graduates all users with year <= the current year, so running this at the beginning of the year will graduate people prematurely (in January instead of May).

PR #732, opened 2024-09-30
Hide personal information for non-authenticated users on club page

julianweng:
* Implements Point of Contact visibility as 3-value enum: admin-only, student/signed in-only, and public (for non-authed), migrates existing `public` boolean to first two.
* Makes responder details anonymous for Club Q&As for non-authenticated calls.

Note to Reviewer: please create a private and a public point of contact for a club in your dev environment prior to pulling these changes and migrating, and make sure that they turn into admin-only and student-only respectively - would like to double-check that the migration would work in prod.


backend/clubs/migrations/0114_alter_advisor_public.py
```diff
@@ -0,0 +1,38 @@
...
+        ),
+        migrations.AddField(
+            model_name="advisor",
+            name="public",
+            field=models.IntegerField(
+                choices=[(1, "Admin Only"), (2, "Signed-in Students"), (3, "Public")],
+                default=2,
+            ),
+        ),
+        migrations.RunPython(migrate_public_to_enum),
```
rm03, line 33: Can you please define the `reverse_code` argument in case we need to undo this migration? See [here](https://docs.djangoproject.com/en/5.1/ref/migration-operations/#runpython)

backend/clubs/models.py
```diff
@@ -1136,7 +1136,18 @@ class Advisor(models.Model):
...
+    PUBLIC_STUDENTS = 2
+    PUBLIC_ALL = 3
+
+    PUBLIC_CHOICES = (
+        (PUBLIC_ADMIN, "Admin Only"),
+        (PUBLIC_STUDENTS, "Signed-in Students"),
+        (PUBLIC_ALL, "Public"),
+    )
+
+    public = models.IntegerField(choices=PUBLIC_CHOICES, default=PUBLIC_STUDENTS)
```
rm03, line 1150: nit: might be more appropriate to name this field something like `visibility`

backend/clubs/serializers.py
```diff
@@ -1249,8 +1249,16 @@ def get_approved_by(self, obj):
         return obj.approved_by.get_full_name()
 
     def get_advisor_set(self, obj):
+        user = self.context["request"].user
+
+        public_filter = (
+            obj.advisor_set.filter(public__gte=Advisor.PUBLIC_STUDENTS).order_by("name")
+            if user.is_authenticated
+            else obj.advisor_set.filter(public=Advisor.PUBLIC_ALL).order_by("name")
+        )
```
rm03, line 1258: nit: could do something like
```suggestion
        visibility_level = Advisor.PUBLIC_STUDENTS if user.is_authenticated else Advisor.PUBLIC_ALL
        queryset = obj.advisor_set.filter(public__gte=visibility_level)
```

backend/clubs/views.py
```diff
@@ -3422,7 +3424,8 @@ def get_queryset(self):
         )
 
         if not self.request.user.is_authenticated:
-            return questions.filter(approved=True)
+            # Hide responder if not authenticated
+            return questions.filter(approved=True).extra(select={"responder": "NULL"})
```
rm03, line 3428: Not sure if this is best practice but I feel that we should avoid using SQL if possible. Feel free to ignore though as this is slightly more verbose:

```suggestion
            return questions.filter(approved=True).annotate(
    responder=Value(None))
```

rm03 (commented):
Great work Julian! Would love to see a couple test cases as well if possible 🙂

backend/clubs/serializers.py
julianweng: Great catch, much cleaner!

backend/clubs/views.py
julianweng: Annotate seems to [not allow overwriting](https://code.djangoproject.com/ticket/28072) of existing fields so extra seems to be a workaround.

backend/clubs/models.py
julianweng: Thanks!

julianweng:
@rm03 Wrote some test cases and also patched a vulnerability related to the separate AdvisorViewSet (non-officers allowed to view all advisors through it), would appreciate if you can take another look

rm03 (approved):
LGTM, thanks Julian

PR #733, opened 2024-10-04
Change how clubs are displayed based on approval status

aviupadhyayula:
Re-hauls serialization to show clubs differently based on their approval status and the current time (whether we're in a renewal period or not). Adds tests to serialization logic and aligns view-related unit tests to this change.

### DIFF
--- a/backend/clubs/serializers.py
+++ b/backend/clubs/serializers.py
@@ -994,30 +994,48 @@ class ClubListSerializer(serializers.ModelSerializer):
 
     def to_representation(self, instance):
         """
-        Return the previous approved version of a club for users
-        that should not see unapproved content.
+        Return appropriate version of club based on user permissions and club status:
+        - Privileged users see current version (or when bypass=True)
+        - For non-privileged users:
+          - During renewal period: show last approved version from past calendar year
+          - Outside renewal period: show last approved version from past calendar year
         """
-        if instance.ghost and not instance.approved:
-            user = self.context["request"].user
+        user = self.context["request"].user
+        can_see_pending = user.has_perm("clubs.see_pending_clubs") or user.has_perm(
+            "clubs.manage_club"
+        )
+        is_member = (
+            user.is_authenticated
+            and instance.membership_set.filter(person=user).exists()
+        )
+        bypass = (
+            self.context["request"].query_params.get("bypass", "").lower() == "true"
+        )
 
-            can_see_pending = user.has_perm("clubs.see_pending_clubs") or user.has_perm(
-                "clubs.manage_club"
-            )
-            is_member = (
-                user.is_authenticated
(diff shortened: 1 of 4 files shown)


backend/tests/clubs/test_serializers.py
```diff
@@ -0,0 +1,152 @@
...
+        self.club.save()
+
+        current_date = self.renewal_end + timedelta(days=1)
+        with mock.patch("django.utils.timezone.now", return_value=current_date):
+            approved_date = self.renewal_start + timedelta(days=1)
+            with mock.patch("django.utils.timezone.now", return_value=approved_date):
+                self.club.name = "Current Cycle Approved Version"
+                self.club.approved = True
+                self.club.approved_on = approved_date
+                self.club.save()
```
Porcupine1, line 108: Unless there is internal logic in `club.save` or `serializer.to_representation` that uses `django.utils.timezone.now`, we might not need the `with` clauses since we are explicitly setting `club.approved_on`

backend/tests/clubs/test_serializers.py
```diff
@@ -0,0 +1,152 @@
...
+        self.renewal_end = now + timedelta(days=30)
+        settings.RENEWAL_PERIOD = (self.renewal_start, self.renewal_end)
+
+        self.serializer = ClubSerializer()
+
+    def test_privileged_user_sees_current_version(self):
+        self.user.user_permissions.add(
+            Permission.objects.get(codename="see_pending_clubs")
+        )
+        request = self.factory.get(reverse("clubs-detail", args=[self.club.code]))
```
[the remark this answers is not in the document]
aviupadhyayula: We don't want to actually send the request to the API -- we just want to create the request object, and then pass it to the serializer. This lets us test the serializer itself instead of the entire view. 

The `user_permissions.add` pattern was unintentional. Changed to `self.user.is_superuser = True` instead.

backend/clubs/serializers.py
```diff
@@ -994,30 +994,48 @@ def get_fields(self):
 
     def to_representation(self, instance):
         """
-        Return the previous approved version of a club for users
-        that should not see unapproved content.
+        Return appropriate version of club based on user permissions and club status:
+        - Privileged users see current version (or when bypass=True)
```
rm03, line 998: What's the use case for `bypass=True`? I don't think clubs should be allowed to show an unapproved version at all

rm03 (approved):
Thanks for taking care of this!

backend/clubs/serializers.py
aviupadhyayula: I think we use `bypass=True` when someone's accepting a membership invite to a club that hasn't been approved yet. (Basing this off the comments in `ClubPermission` from `permissions.py`.)

PR #736, opened 2024-10-12
Admin note frontend

rm03:
Adds a tab under the club management page for site admins to add notes about clubs as per OSA's request. Previously, site admins could only use interact with notes via the backend.

image


julianweng (approved):
LGTM, thanks Rohan!

PR #738, opened 2024-10-13
Rename club description to mission

owlester12:
Change language around club descriptions as per meeting with OSA.


julianweng (approved):

aviupadhyayula (commented):
Are there any other references to club descriptions in the codebase? I know we have a few modals in the registration process that mention it -- would be nice to change to say to "mission statement" (or something similar).

owlester12:
I went through all of the times on the frontend where "description", "Description", and "DESCRIPTION" were mentioned and where appropriate changed them to mission (like on the main club page, editing/creating clubs, but not anything to do with events or ticketing or descriptions of other components).

PR #741, opened 2024-10-20
Move ranking information from separate /rank page to ClubEditPage modal

julianweng:
* Creates toggle-able popup with simplified information about club rankings and how to improve them from a club leader perspective
* Deletes `/rank` page

Help text toggle

Rank modal


frontend/components/ClubEditPage/ClubEditCard.tsx
```diff
@@ -397,6 +399,57 @@ export default function ClubEditCard({
     {
       name: 'General',
       type: 'group',
+      description: (
+        <div className="mb-4">
+          <a onClick={() => setShowRankModal(true)} className="is-italic">
+            How can filling out this information help your club's ranking?
```
aviupadhyayula, line 405: Is there a reason we're italicizing? I can't remember doing it anywhere else on the frontend.

aviupadhyayula (commented):
LGTM, thanks for picking this up Julian!

frontend/components/ClubEditPage/ClubEditCard.tsx
```diff
@@ -397,6 +399,57 @@ export default function ClubEditCard({
...
+          </a>
+          <Modal
+            show={showRankModal}
+            closeModal={() => setShowRankModal(false)}
+            marginBottom={false}
+            width="80%"
+          >
+            <ModalContent className="content mb-4">
+              <h2>Rank Algorithm</h2>
```
aviupadhyayula, line 414: Nit: can we make this title a little better? Maybe something like "How club rankings are calculated" (you get the idea).

frontend/components/ClubEditPage/ClubEditCard.tsx
aviupadhyayula: Taking another look at this: we should probably make this less overt, since OSA will see this whenever they open the club management view. Let's change this to something like "How does filling out this information affect your club's page?" -- anything that doesn't involve the word "rank".

julianweng:
Thanks for the feedback @aviupadhyayula!

PR #745, opened 2024-11-06
Set `ghost` to false on club approval decision

aviupadhyayula:
Currently, approval decisions patch the `approved` and `approved_by` fields, but the `ghost` field remains unchanged. This causes the "changes to this club are pending approval" modal to show up for clubs that have been approved.

### DIFF
--- a/frontend/components/ClubPage/ClubApprovalDialog.tsx
+++ b/frontend/components/ClubPage/ClubApprovalDialog.tsx
@@ -265,6 +265,7 @@ const ClubApprovalDialog = ({ club }: Props): ReactElement | null => {
                           body: {
                             approved: true,
                             approved_comment: comment,
+                            ghost: false,
                           },
                         })
                           .then(() => router.reload())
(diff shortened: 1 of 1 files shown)


julianweng:
It looks like we do have logic to set ghost to be false within `ClubSerializer`, [here](https://github.com/pennlabs/penn-clubs/blob/3a002c404617cf5fabb468b8d1eeb04ecd2b18d2/backend/clubs/serializers.py#L1580). The edge case which prompted this PR was likely a result of an admin approving a club manually from Django Admin, which does not use the save function within the serializer.

This is unrelated but we may have a separate bug / edge case at L1557 where we use the nullity of `approved` to signify revoking approval, but we use `approved=false` elsewhere including on the frontend, which might have led to other issues seen previously: curious to know your thoughts on it?

aviupadhyayula:
Good catch, looks like the serializer makes this PR unnecessary. I'm less sure that Rodney approving the club through the API admin is what caused this. Both the `approved_by` and `approved_on` fields are set (and dated to 8-28), and it seems implausible that he'd know to do this. 

That's definitely a problem. IIRC `approved=None` signifies clubs that are seeking reapproval, while `approved=False` is reserved for clubs that have had their most recent approval request denied.

aviupadhyayula:
Going to close this, and take a look at how we use `approved` on the frontend in another PR.

PR #746, opened 2024-11-11
Approval history UI

julianweng:
Allows site admins and associated leaders to view a club's approval history.

Club home page list of approvals/rejections/submissions

Collapsed view

Expanded reason modal


backend/clubs/views.py
```diff
@@ -1276,6 +1277,49 @@ def upload_file(self, request, *args, **kwargs):
...
+                                        format: date-time
+                                        description: >
+                                            The time in which the specific version
+                                            of the club was saved at.
+        ---
+        """
+        club = self.get_object()
+        return Response(
+            ApprovalHistorySerializer(
+                club.history.order_by("history_date"),
```
aviupadhyayula, line 1317: Should this be ordered in reverse, so that recent approvals are first?

backend/clubs/views.py
```diff
@@ -1276,6 +1277,49 @@ def upload_file(self, request, *args, **kwargs):
 
         return file_upload_endpoint_helper(request, code=club.code)
 
+    @action(detail=True, methods=["get"])
+    def history(self, request, *args, **kwargs):
```
aviupadhyayula, line 1281: Do we need to add this route to the `ClubPermission` permission class? If not, I think the default is only members & above can see routes that aren't explicitly handled.

aviupadhyayula (commented):
This is a sick feature, thanks for taking the initiative on this. LGTM outside of a couple comments.

backend/clubs/views.py
rm03: I think Julian's handling it on the frontend, but would prefer to handle it here.

backend/clubs/serializers.py
```diff
@@ -2975,6 +2975,37 @@ class Meta:
...
 
+class ApprovalHistorySerializer(serializers.ModelSerializer):
+    approved = serializers.BooleanField()
+    approved_on = serializers.DateTimeField()
+    approved_by = serializers.SerializerMethodField("get_approved_by")
+    approved_comment = serializers.CharField()
+    history_date = serializers.DateTimeField()
+
+    def get_approved_by(self, obj):
+        user = self.context["request"].user
```
rm03, line 2986: Instead of casing on permissions here, it might be better use a permission class on the route (as Avi suggested) since only officer+ will be able to view this information.

rm03 (commented):
Thanks Julian, good work

rm03 (approved):

PR #748, opened 2024-11-23
Restrict event creation to approved clubs

aviupadhyayula:
Prevents clubs that aren't approved from creating events. This shouldn't change any existing behavior: events by non-approved clubs currently 404. (Should also resolve OSA's concerns about only approved clubs being able to use ticketing.)


aviupadhyayula:
Closing as #750 handles these changes.

PR #749, opened 2024-11-23
Add support for (ticket) email blasts

aviupadhyayula:
Adds support for email blasts from event owners to ticket holders. I'm envisioning using this for an admin feature as well (ex: OSA wants to send a message to all club leaders, like they did at the start of this semester). 

Here's what an event-related email might look like:
Screenshot 2024-11-23 at 12 42 07 AM

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -3067,6 +3067,17 @@ class ClubEventViewSet(viewsets.ModelViewSet):
                                     type: string
                                     description: A message indicating how many
                                         recipients received the blast
+            "400":
+                description: Content field was empty or missing
+                content:
+                    application/json:
+                        schema:
+                            type: object
+                            properties:
+                                detail:
+                                    type: string
+                                    description: Error message indicating content
+                                        was not provided
             "404":
                 description: Event not found
                 content:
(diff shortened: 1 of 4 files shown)


backend/clubs/views.py
```diff
@@ -3039,6 +3039,75 @@ def issue_tickets(self, request, *args, **kwargs):
...
+        officer_emails = event.club.get_officer_emails()
+        emails = list(holder_emails) + list(officer_emails)
+
+        send_mail_helper(
+            name="blast",
+            subject=f"Update on {event.name} from {event.club.name}",
+            emails=emails,
+            context={
+                "sender": event.club.name,
+                "content": request.data.get("content"),
```
aviupadhyayula, line 3097: Should we check if this is empty?

julianweng: Yea, probably strip whitespace and check if empty after.

julianweng (changes requested):
Thanks for taking the lead on this: you might have to add the action name to the relevant code in `EventPermission` in order to ensure that only club leaders use it but looks good otherwise. Also, would it be possible to take a first crack at the UI? Shouldn't be more than 20 or so loc in the club-specific ticket page.

aviupadhyayula:
> Also, would it be possible to take a first crack at the UI? Shouldn't be more than 20 or so loc in the club-specific ticket page.

Sure, but I think the plan is to have a designer take a crack at the UI before we make any changes or additions to it. We're going to need to change how the ticket dashboard looks to club leaders a lot (add refunds, email blasts, revenue metrics, question answers, etc.). Let me know if that sounds reasonable.

backend/clubs/views.py
```diff
@@ -3088,6 +3099,13 @@ def email_blast(self, request, *args, **kwargs):
         officer_emails = event.club.get_officer_emails()
         emails = list(holder_emails) + list(officer_emails)
 
+        content = request.data.get("content").strip()
```
julianweng, line 3102: Might error out if no content is provided before being able to return the response, maybe request.data.get("content", "")?

julianweng (approved):
LGTM other than one nit.

PR #750, opened 2024-11-23
Ticket drop time UI support, unapproved club event visibility changes

julianweng:
* Add UI support for ticket drop times
* Add `ticket_drop_time` validation in event serializer based on event end time
* Prevent retroactive editing of ticket drop time after tickets have been sold (plus handling of edge case where ticket has been added to cart but not bought before ticket drop time is pushed back)
* Prevent unapproved clubs from selling tickets (add_to_cart)
* Prevent clubs which have been archived from selling existing tickets
* Allow leaders of unapproved clubs to create and view events (with the warning that such events are not visible to the public until the club is approved), allowing for "draft" functionality (also currently, the event just disappears after submission on the frontend for unapproved clubs...)
* Standardize spelling of "publicly" in the codebase (alternative being "publically")
* Add button to view public event page within event editing page
* Show all clubs in `/directory` which have been approved at some point

Event page with unreleased tickets

Unapproved club event (displayed to officers only)

Note for events without a ticket_drop_time

User viewport after clicking on link to ticket_drop_time field within note

### DIFF
--- a/backend/clubs/serializers.py
+++ b/backend/clubs/serializers.py
@@ -457,9 +457,17 @@ class ClubEventSerializer(serializers.ModelSerializer):
         end_time = data.get(
             "end_time", self.instance.end_time if self.instance is not None else None
         )
+        ticket_drop_time = data.get(
+            "ticket_drop_time",
+            self.instance.ticket_drop_time if self.instance is not None else None,
+        )
         if start_time is not None and end_time is not None and start_time > end_time:
             raise serializers.ValidationError(
                 "Your event start time must be less than the end time!"
+            )
+        if ticket_drop_time is not None and ticket_drop_time > end_time:
+            raise serializers.ValidationError(
+                "Your ticket drop time must be before the event ends!"
             )
         return data
 
@@ -507,6 +515,7 @@ class ClubEventSerializer(serializers.ModelSerializer):
             "location",
             "name",
             "start_time",
+            "ticket_drop_time",
             "ticketed",
             "type",
             "url",
--- a/backend/clubs/serializers.py
+++ b/backend/clubs/serializers.py
@@ -465,7 +465,7 @@ class ClubEventSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError(
                 "Your event start time must be less than the end time!"
             )
-        if ticket_drop_time is not None and ticket_drop_time > end_time:
+        if ticket_drop_time is not None and ticket_drop_time >= end_time:
             raise serializers.ValidationError(
                 "Your ticket drop time must be before the event ends!"
             )
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -3216,7 +3216,10 @@ class ClubEventViewSet(viewsets.ModelViewSet):
         event = self.get_object()
         if (
             "ticket_drop_time" in request.data
-            and parse(request.data["ticket_drop_time"]) >= timezone.now()
+            and (
+                event.ticket_drop_time is None  # case where sales immediately start
+                or event.ticket_drop_time <= timezone.now()
+            )
             and Ticket.objects.filter(event=event, owner__isnull=False).exists()
         ):
             raise DRFValidationError(
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -1762,7 +1762,7 @@ class ClubViewSet(XLSXFormatterMixin, viewsets.ModelViewSet):
         """
         serializer = ClubMinimalSerializer(
             Club.objects.all()
-            .exclude(Q(approved=False) | Q(archived=True))
+            .exclude((~Q(approved=True) & Q(ghost=False)) | Q(archived=True))
             .order_by(Lower("name")),
             many=True,
         )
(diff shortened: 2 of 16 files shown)


backend/clubs/views.py
```diff
@@ -5335,6 +5393,9 @@ def initiate_checkout(self, request, *args, **kwargs):
         # are locked, we shouldn't block.
         tickets = cart.tickets.select_for_update(skip_locked=True).filter(
             Q(holder__isnull=True) | Q(holder=self.request.user),
+            Q(event__ticket_drop_time__lt=timezone.now())
+            | Q(event__ticket_drop_time__isnull=True),
+            event__club__archived=False,
```
aviupadhyayula, line 5398: Same note as above here. Will there ever be a case where a ticket is in a user's cart, but the drop time hasn't elapsed?

aviupadhyayula (commented):
Thanks for picking this up (and sorry for getting to this so late). Left some comments on the backend changes. Frontend changes LGTM!

backend/clubs/views.py
```diff
@@ -2464,6 +2464,24 @@ def add_to_cart(self, request, *args, **kwargs):
...
+        club = Club.objects.filter(code=event.club.code).first()
+        # As clubs cannot go from historically approved to unapproved, we can
+        # check here without checking further on in the checkout process
+        # (the only exception is archiving a club, which is checked)
+        if not club:
+            return Response(
+                {"detail": "Related club does not exist", "success": False},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+        elif not club.approved and not club.ghost:
```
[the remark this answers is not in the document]
julianweng: Yep, if a club has no approved value, then it's either new, in which case ghost would also be none and the condition should evaluate to true (e.g. the club shouldn't be able to sell tickets), or approved previously but recently changed, in which case the club should be able to sell tickets, ghost would be true, and this condition would be false.

backend/clubs/views.py
```diff
@@ -5203,8 +5254,13 @@ def cart(self, request, *args, **kwargs):
 
         tickets_to_replace = cart.tickets.filter(
             Q(owner__isnull=False)
+            | Q(event__club__archived=True)
             | Q(holder__isnull=False)
             | Q(event__end_time__lt=now)
+            | (
+                Q(event__ticket_drop_time__gt=timezone.now())
+                & Q(event__ticket_drop_time__isnull=False)
+            )
```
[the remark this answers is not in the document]
julianweng: This is handling the edge case where people have added tickets to their cart under a certain ticket drop time, but the drop time has since been updated (assuming that existing cartholders shouldn't be "grandfathered" into the old ticket drop time).

backend/clubs/views.py
julianweng: Responded above.

backend/clubs/serializers.py
```diff
@@ -457,10 +457,18 @@ def validate(self, data):
...
+            self.instance.ticket_drop_time if self.instance is not None else None,
+        )
         if start_time is not None and end_time is not None and start_time > end_time:
             raise serializers.ValidationError(
                 "Your event start time must be less than the end time!"
             )
+        if ticket_drop_time is not None and ticket_drop_time >= end_time:
+            raise serializers.ValidationError(
+                "Your ticket drop time must be before the event ends!"
+            )
```
aviupadhyayula, line 471: Thinking about this again: would it make more sense for the ticket drop time to be before the event's start time? I can't think of a scenario where the event would start but tickets would drop midway through it.

backend/clubs/views.py
```diff
@@ -3193,24 +3208,60 @@ def destroy(self, request, *args, **kwargs):
...
+        """
+        event = self.get_object()
+        if (
+            "ticket_drop_time" in request.data
+            and Ticket.objects.filter(event=event, owner__isnull=False).exists()
+        ):
+            raise DRFValidationError(
+                detail="""Ticket drop times cannot be edited
+                        after tickets have been sold."""
+            )
```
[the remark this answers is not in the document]
aviupadhyayula: I thinking only checking this once the ticket drop time has elapsed makes sense.

backend/clubs/views.py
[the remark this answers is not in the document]
aviupadhyayula: I see, that makes sense.

backend/clubs/views.py
```diff
@@ -5241,6 +5297,8 @@ def cart(self, request, *args, **kwargs):
                 continue
 
             available_tickets = Ticket.objects.filter(
+                Q(event__ticket_drop_time__lt=timezone.now())
```
aviupadhyayula, line 5300: Nit: should be less than or equal to.

backend/tests/clubs/test_ticketing.py
```diff
@@ -605,6 +647,50 @@ def test_add_to_cart_before_ticket_drop(self):
         # Tickets should not be added to cart before drop time
         self.assertEqual(resp.status_code, 403, resp.content)
 
+    def test_add_to_cart_unapproved_club(self):
```
aviupadhyayula, line 650: It'd be nice to test the case where a user adds a ticket that's dropped, the ticket drop time gets pushed back, and the user attempts to check out. But maybe it's extra, feel free to skip.

aviupadhyayula (commented):
LGTM Julian! Left some comments, thanks for picking this PR back up.

aviupadhyayula (approved):
Backend LGTM! If you feel good about the frontend feel free to merge.

PR #751, opened 2024-11-24
Migrate from Babel to SWC

julianweng:
* Migrate to SWC for faster compilation performance
* Use [built-in Next.js SWC Styled Components support](https://nextjs.org/docs/14/architecture/nextjs-compiler#styled-components)
* Replace Istanbul Babel plugin with [SWC plugin ](https://github.com/kwonoj/swc-plugin-coverage-instrument) (for continuity, not sure if this works or is needed since we don't really have tests or a frontend check step anymore)


aviupadhyayula (approved):
LGTM, thanks for taking the lead on this Julian. Don't have much context here but trust that you've got this handled.

PR #752, opened 2024-12-17
Wharton Dashboard fixes

owlester12:
I fixed the wharton/status page. I know you want me to fix the cycles page. However, I don't yet have access to the cycles page on Penn clubs website and the one on localhost just points to a create cycle component.

### DIFF
--- a/frontend/components/Settings/WhartonApplicationStatus.tsx
+++ b/frontend/components/Settings/WhartonApplicationStatus.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement, useState } from 'react'
-import { DiscreteColorLegend, Hint, RadialChart } from 'react-vis'
+import { Hint, RadialChart } from 'react-vis'
+import styled from 'styled-components'
 
 import { WHITE } from '~/constants/colors'
 import { ApplicationStatus } from '~/types'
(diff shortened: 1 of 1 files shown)


julianweng (changes requested):
Thanks for picking this up! I don't think there's anything left to fix in the Cycles tab so the Status page is more than enough.

The legend fix looks great!

Wondering if there's a more elegant way than the triple-nested loop. Maybe `Object.groupBy` could be used to create applicationsCommittees since from my understanding it's just grouping by name then one final iteration to project into applicationsGrouped?

Also, I'm not sure if the Application Name is a reliable indicator for similar applications across different cycles since they are often different in practice (e.g. "Penn Memes Application Fall 2022"), at least in production. Maybe the club name would be a better groupby attribute to fix the issues we discussed around applications across cycles not being consolidated?

unconsolidated bar charts of application

owlester12:
I changed it so that we use status.club instead of status.name . We can create applicationCommittees using groupby, but the for loop allows us to create both applicationCommittees and applicationTotals at the same time (and it isn't nested at all). They are the same asymptotically and is most like the original code. With regards to the triple for loop:

Applications committee is a dictionary object that looks like 
{club : { 
    committee: {
       status: number,
    }, ...
}, ...}

It is triple nested, so to make sure we get all of the key value pairs, we have to loop through all of the clubs and then all of the committees within each club and then all of the statuses in each committee. GroupBy works on arrays, not dictionaries. So, I think that, although not that elegant, the triple nested for loop captures all of the key value pairs as efficiently as possible.

julianweng (approved):
Looks great to me, thanks for picking this up Owen!

Issue #754, opened 2024-12-24
Allow club leaders to choose who pays transaction fees


aviupadhyayula:
Give club leaders the option to choose whether buyers will pay the transaction fee or not.

Allow club leaders to choose whether buyers will cover the transaction fee or not.

aviupadhyayula:
Addressed in #767.

aviupadhyayula:
Reopening as ticketing-related development is moving off of #767 to `master`.

PR #762, opened 2025-01-02
Add option to pass transaction fees onto buyers

aviupadhyayula:
Refactors out event-wide settings for ticketed events into a new `TicketSettings` model. Adds backend support for club leaders to pass transaction fees onto buyers. (This isn't reflected in calculated cart totals yet. We're waiting on Penn Finance to determine how transaction fees will be collected before implementing this.)

Addresses #754.


backend/clubs/models.py
```diff
@@ -1821,6 +1823,23 @@ class Cart(models.Model):
...
+class TicketSettings(models.Model):
+    """
+    Configuration settings for events that have tickets.
+    Only created when an event has associated tickets created.
+    """
+
+    event = models.OneToOneField(
+        Event, on_delete=models.CASCADE, related_name="ticket_settings"
+    )
+    order_limit = models.IntegerField(null=True, blank=True)
```
aviupadhyayula, line 1835: Removed the default order limit of 10 -- I don't see a reason why we should have it enabled by default. Happy to change it back if you guys disagree.

backend/clubs/models.py
```diff
@@ -1821,6 +1823,23 @@ class Cart(models.Model):
...
+    drop_time = models.DateTimeField(null=True, blank=True)
+    fee_charged_to_buyer = models.BooleanField(default=False)
```
aviupadhyayula, line 1837: Setting transaction fees to be shouldered by the club by default.

backend/clubs/models.py
Porcupine1: Agreed, but I think it was more of a safety net for say club officers that wouldn't know why they might want to limit orders per customer. Having a default limit might force them to think about this and opt out after making an informed decision. Also, since we will be opening ticket purchases to non-penn people, we have a larger pool of potential bad actors

backend/clubs/views.py
```diff
@@ -2772,27 +2798,59 @@ def create_tickets(self, request, *args, **kwargs):
         """
         event = self.get_object()
 
-        # Tickets can't be edited after they've dropped
-        if event.ticket_drop_time and timezone.now() > event.ticket_drop_time:
+        # Tickets can't be edited after they've been sold or checked out
+        if event.tickets.filter(
+            Q(owner__isnull=False) | Q(holder__isnull=False)
+        ).exists():
```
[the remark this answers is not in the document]
aviupadhyayula: Good point, but I think I want to leave it in for now. If something changes with the redesign of checkout (e.g. we use holds in a different context), it'll be a pain to go back and add these checks to all the previous PRs. Once we commit to a specific implementation checkout, we can just go through and remove all the holding logic in one go.

backend/clubs/views.py
```diff
@@ -2772,27 +2798,59 @@ def create_tickets(self, request, *args, **kwargs):
...
-                {"detail": "Tickets cannot be edited after they have dropped"},
+                {
+                    "detail": "Tickets cannot be edited after they have been "
+                    "sold or checked out"
+                },
                 status=status.HTTP_403_FORBIDDEN,
             )
 
-        # Tickets can't be edited after they've been sold or held
+        ticket_settings, _ = TicketSettings.objects.get_or_create(event=event)
```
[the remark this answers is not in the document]
aviupadhyayula: What happens if we're creating an event for the first time? Then there is no `ticket_settings` associated with the `event`, and the check below (if `event.ticket_settings.drop_time` is set) throws an exception, since `TicketSettings` is a one-to-one relationship.

backend/clubs/models.py
aviupadhyayula: Good point. Since there'll eventually be ticketed events with free tickets, letting non-Penn users check out all of them would be disastrous. Let's re-implement the limit.

PR #768, opened 2025-01-18
Add support for viewing and refunding transactions

### LINKED ISSUE #756: Add support for refunding transactions

aviupadhyayula:
Creates new viewset, serializer, and permission class to interact with transaction records. Superusers can see all transaction records. Non-privileged users must specify an event, and can only view transactions for that event if they're appropriately privileged. 

Also contains minor modifications to attributes in `TicketTransactionRecord`.

TODO: still need to integrate with Cybersource's API for refunds. 

Closes #756.


rohangpta:
Seems like a blocker to launch?

PR #772, opened 2025-01-29
Prevent duplicate committees

aviupadhyayula:
Clubs sometimes mess up during creating applications, and [have multiple committees with the same name](https://penn-labs.slack.com/files/U07NEDHL7RB/F089TTDBBSA/screenshot_2025-01-23_at_12.48.58___pm.png). 

This PR adds serializer-level checks to prevent duplicate committees tied to the same application. Also adds a DB-level constraint. 

The `merge_duplicate_committees` command retroactively fixes existing applications by merging committees together. This should be run in prod, and then removed from this PR.


aviupadhyayula:
Made redundant by #814

PR #776, opened 2025-02-09
Add frontend health route

sritanmotati:
/health route


julianweng (changes requested):
Hi, thanks for taking the initiative on this, can you split this PR into multiple if you intend on doing all of these things (editing OSA cronjob time, non-health related prettier changes, and health): would like to keep this focused on just health.

k8s/main.ts
```diff
@@ -83,7 +83,7 @@ export class MyChart extends PennLabsChart {
     })
 
     new CronJob(this, 'osa-perms-updates', {
-      schedule: cronTime.every(1).minutes(),
+      schedule: cronTime.every(5).minutes(),
```
aviupadhyayula, line 86: Is there a reason we're moving this back to 5 minutes? We've noticed perms being reset upon logout and have given OSA directions to wait 1 minute before trying again. Would be nice if we could stick to that.

aviupadhyayula (commented):
Thanks for taking the initiative on this! Agree with Julian that it'd be nice to break this up into separate PRs -- feel free to remove the `k8s` changes and just push those to another branch.

k8s/main.ts
sritanmotati: Yeah I'll remove this change, we originally wanted to try moving to 5 minutes because we thought this was a potential cause of the deployment issues, but that wasn't the case. I'll keep it at 1.

PR #778, opened 2025-02-11
Allow Wharton Council officers to manage associated clubs

julianweng:
Enable WC officers to change club settings and provisioned applications for WC clubs, avoiding access issues involving non-WC clubs.

### DIFF
--- a/backend/clubs/permissions.py
+++ b/backend/clubs/permissions.py
@@ -163,6 +163,10 @@ class ClubPermission(permissions.BasePermission):
         # user must be in club or parent club to perform non-view actions
         membership = find_membership_helper(request.user, obj)
         if membership is None:
+            if obj.is_wharton:
+                return WhartonApplicationPermission.check_wharton_council_officer(
+                    self, request
+                )
             return False
         # user has to be an owner to delete a club, an officer to edit it
         if view.action in {"destroy"}:
@@ -338,15 +342,14 @@ class WhartonApplicationPermission(permissions.BasePermission):
     Grants permission if the user is an officer of Wharton Council
     """
 
-    WHARTON_COUNCIL_CLUB_CODE = "wharton-council"
-
     def check_wharton_council_officer(self, request):
+        WHARTON_COUNCIL_CLUB_CODE = "wharton-council"
         if not request.user.is_authenticated:
             return False
         user = get_user_model().objects.filter(username=request.user).first()
         if user is not None:
             membership = Membership.objects.filter(
-                club__code=self.WHARTON_COUNCIL_CLUB_CODE, person=user
+                club__code=WHARTON_COUNCIL_CLUB_CODE, person=user
             ).first()
             if membership is not None:
                 return membership.role <= Membership.ROLE_OFFICER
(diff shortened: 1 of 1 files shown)


backend/clubs/permissions.py
```diff
@@ -338,15 +349,14 @@ class WhartonApplicationPermission(permissions.BasePermission):
...
-
     def check_wharton_council_officer(self, request):
+        WHARTON_COUNCIL_CLUB_CODE = "wharton-council"
         if not request.user.is_authenticated:
             return False
         user = get_user_model().objects.filter(username=request.user).first()
         if user is not None:
             membership = Membership.objects.filter(
-                club__code=self.WHARTON_COUNCIL_CLUB_CODE, person=user
+                club__code=WHARTON_COUNCIL_CLUB_CODE, person=user
```
aviupadhyayula, line 359: Just to confirm: we're moving this inside because non `WhartonApplicationPermission` instances will now be using this method, correct?

backend/clubs/permissions.py
```diff
@@ -298,7 +302,14 @@ def has_permission(self, request, view):
                 return True
             obj = Club.objects.get(code=view.kwargs["club_code"])
             membership = find_membership_helper(request.user, obj)
-            return membership is not None and membership.role <= Membership.ROLE_OFFICER
+            return (
+                membership is not None and membership.role <= Membership.ROLE_OFFICER
+            ) or (
+                obj.is_wharton
+                and WhartonApplicationPermission.check_wharton_council_officer(
+                    self, request
```
aviupadhyayula, line 310: This feels a little messy somehow -- thoughts on making `check_wharton_council_officer` a static class method? This gets around the passing of `self` into this call.

aviupadhyayula (commented):
Thanks for taking the initiative on this Julian -- pretty cool way to get around the WC permissions stuff. Left a couple comments.

backend/clubs/permissions.py
julianweng: Yea

backend/clubs/permissions.py
```diff
@@ -163,6 +163,8 @@ def has_object_permission(self, request, view, obj):
         # user must be in club or parent club to perform non-view actions
         membership = find_membership_helper(request.user, obj)
```
rohangpta, line 164: Skimming this and memory not the best of the code base... but isn't this already supported? MUSE for example has WC as a parent org and the tree traversal in `find_membership_helper` should catch this case (are parent orgs synced?) 

image

rohangpta: Old PR https://github.com/pennlabs/penn-clubs/pull/241

julianweng: Thanks for the catch, definitely is supported (sync apparently wasn't a cron: pennlabs/penn-clubs@f608990 but that's a separate story) ... will repurpose this PR for the minor change to approved club version viewing and adding a test for the `find_membership_helper` behavior

julianweng:
@aviupadhyayula we now sync badges to parent-child relationships on editing badges of a club as discussed: it should ensure consistency as long as 1. badge organizations aren't editable and 2. the badge being edited is a result of a direct parent-child relationship as opposed to an indirect descendant, can you verify if this is true and if so, what kind of workarounds we can implement to support these?

PR #781, opened 2025-02-15
Application submission management clarity

julianweng:
Some QoL improvements to the application submission management flow for club officers, aiming at reducing user error and increasing clarity for the application decision sending process:

* Add warning toast when sending decisions while having selected applications (route send decisions to all submissions regardless of selection on frontend, so can result in wrong user assumptions about who the updates are sent to)
* Wording changes to emphasize the above fact and warn users against the resend notifications feature
* Route to send application updates now has detail breaking down why certain users were not sent new emails
* (not that related) Putting in non-existent club/application ids into the `/application` route returns a proper X not found screen instead of an empty one

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -6619,17 +6619,17 @@ class ClubApplicationViewSet(viewsets.ModelViewSet):
                 invite.send_mail(self.request)
 
         dry_run_msg = "Would have sent" if dry_run else "Sent"
-        skip_msg = ""
+        skip_msg = []
         if skip["already_notified"]:
-            skip_msg += f" {skip['already_notified']} already notified"
+            skip_msg.append(f" {skip['already_notified']} already notified")
         if skip["pending"]:
-            skip_msg += f" {skip['pending']} still marked as pending"
+            skip_msg.append(f" {skip['pending']} still marked as pending")
         if skip["no_reason"]:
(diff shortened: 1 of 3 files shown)


backend/clubs/views.py
```diff
@@ -6617,10 +6619,17 @@ def send_emails(self, *args, **kwargs):
...
+            skip_msg += f" {skip['already_notified']} already notified"
+        if skip["pending"]:
+            skip_msg += f" {skip['pending']} still marked as pending"
+        if skip["no_reason"]:
+            skip_msg += f" {skip['no_reason']} have no reason provided"
         return Response(
             {
                 "detail": f"{dry_run_msg} emails to {n} people, "
-                f"skipping {skip} due to one of (already notified, no reason, no email)"
+                f"{skip_msg if skip_msg else ''}",
```
aviupadhyayula, line 6632: Not sure how this message shows up on the frontend (guessing it's a toast), but are these going to render as on the same line? Might be good to add newlines or some sort of delimiter if so.

aviupadhyayula (commented):
LGTM, thanks for taking the initiative on this Julian 🐐🚀

julianweng:
(precommit formatting was a little messed up and led to un-related changes so skipping commit checks for this one change)

PR #782, opened 2025-02-16
Migrate to uv

clay53:
Currently seems to fail unless using uv from python 3.11


aviupadhyayula (changes requested):
Thanks for taking the initiative on this. Quick question: is there a reason we're migrating now? The Clubs team has discussed this change before, and our understanding has been that `uv` isn't quite mature enough. 

Also, it looks like CI/CD is failing. Changing the `Dockerfile` to use `uv` instead of `pip` should solve.

clay53:
Sorry, I meant to submit this as a draft for now. This is my newbie project. I think it's because it'll be better to use in waypoint.

> Thanks for taking the initiative on this. Quick question: is there a reason we're migrating now? The Clubs team has discussed this change before, and our understanding has been that `uv` isn't quite mature enough. Has anything changed on that front?
> 
> Also, it looks like CI/CD is failing. Changing the `Dockerfile` to use `uv` instead of `pip` should solve.

aviupadhyayula:
Closing as per #800.

PR #785, opened 2025-02-22
Restore frontend check, bump action versions

julianweng:
* Add back frontend check for lint purposes
* Bump deprecated action versions across CICD

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -110,6 +110,7 @@
     "nyc": "^15.1.0",
     "prettier": "^3.5.1",
     "swc-plugin-coverage-instrument": "^0.0.25",
+    "typescript": "5.5.4",
     "wait-on": "^7.2.0"
   },
   "engines": {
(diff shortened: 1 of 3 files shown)


frontend/package.json
```diff
@@ -110,6 +110,7 @@
     "nyc": "^15.1.0",
     "prettier": "^3.5.1",
     "swc-plugin-coverage-instrument": "^0.0.25",
+    "typescript": "5.5.4",
```
aviupadhyayula, line 113: Why are we adding this?

aviupadhyayula (approved):
LGTM, thanks for shipping all over the codebase 🙂

frontend/package.json
julianweng: I removed it as part of https://github.com/pennlabs/penn-clubs/commit/9a18193243377a028ba86cc8ea13a18d3a8c7dbd, but the dependency is needed for frontend check specifically.

PR #786, opened 2025-02-23
Add email blast feature to admin dashboard

### LINKED ISSUE #783: Add admin support for email blasts
As per conversation with OSA, add feature in admin console to email Clubs users. Should support specific targeting (e.g. club leaders, officers, etc.)

aviupadhyayula:
Adds backend route to email selected populations (e.g. club leaders, officers, or all members). This route's only usable by admins. Also adds frontend UI to admin dashboard.

Closes #783.

### DIFF
--- a/backend/clubs/permissions.py
+++ b/backend/clubs/permissions.py
@@ -171,6 +171,11 @@ class ClubPermission(permissions.BasePermission):
             return membership.role <= Membership.ROLE_OFFICER
 
     def has_permission(self, request, view):
+        if view.action in {"email_blast"}:
+            return request.user.is_authenticated and (
+                request.user.is_staff or request.user.has_perm("clubs.manage_club")
+            )
+
         if view.action in {
             "children",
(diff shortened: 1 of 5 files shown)


aviupadhyayula:
Cursor can't compare to the frontend goat 🙁

I'm a little confused by what you mean about integration -- which viewset do you see this fitting in with? Or are you speaking from a UI perspective? Definitely agree with the idea though, it'd be nice to consolidate a lot of this admin lookup/bulk action stuff into a singular interface (both on the frontend, and on the backend).

julianweng:
Referring to both, mostly UI but there could also be some consolidation on the backend (with whichever viewset generates reports right now). Given that both reports and email blasts (ideally) allow admins to filter and find users and clubs by various attributes, with the only difference being whether to send an email or add to a CSV row a given user/club at the end, they should be able to be abstracted as "actions" to be taken on a more generic analytics feature.

Issue #784, opened 2025-02-21
Implement stricter access controls


aviupadhyayula:
As per conversation with OSA, implement stricter access controls to Clubs directory and club pages.

aviupadhyayula:
Closed by #788.

PR #788, opened 2025-03-01
Add frontend auth gate for all pages

julianweng:
Addresses #784 by adding frontend redirects from all pages to login page if not authed.


aviupadhyayula (approved):
Didn't test locally but looks right. LGTM, thanks for picking this up Julian.

PR #791, opened 2025-03-11
Add functionality to control registration queue

### LINKED ISSUE #789: Allow admins to toggle registration queue from dashboard
As per OSA's request, we'd like to support the ability to open and close the registration queue from the admin dashboard. Currently, the queue has two tiers: one for existing groups, and one for new ones. Both queues should be toggle-able from the dashboard.

aviupadhyayula:
Migrates control of the registration queue from environment variables to a singleton model. Adds protected route to patch queue settings as needed. (Closes #789.)


aviupadhyayula:
@julianweng do you mind adding a frontend component for this to the PR? I was thinking we could throw something under the Approval Queue tab of the admin dashboard.

backend/tests/clubs/test_views.py
```diff
@@ -3569,3 +3556,63 @@ def test_health(self):
...
+        self.client.login(username="user", password="password")
+        resp = self.client.get(reverse("queue-settings"))
+        self.assertEqual(resp.status_code, 403, resp.content)
+
+        self.client.login(username="super", password="password")
+        resp = self.client.get(reverse("queue-settings"))
+        self.assertEqual(resp.status_code, 200, resp.content)
+        data = resp.json()
+        self.assertIn("reapproval_queue_open", data)
+        self.assertIn("new_approval_queue_open", data)
```
Porcupine1, line 3589: I think it would be nice to have comments here:
- not a user
- not a super-user
- a super user

backend/tests/clubs/test_views.py
```diff
@@ -3569,3 +3556,63 @@ def test_health(self):
...
+        )
+        self.assertEqual(resp.status_code, 200, resp.content)
+        data = resp.json()
+        self.assertFalse(data["reapproval_queue_open"])
+        self.assertFalse(data["new_approval_queue_open"])
+
+        resp = self.client.get(reverse("queue-settings"))
+        data = resp.json()
+        self.assertFalse(data["reapproval_queue_open"])
+        self.assertFalse(data["new_approval_queue_open"])
```
Porcupine1, line 3618: same thing here (comment wrt user)

Porcupine1 (approved):
I like this idea! PR looks good, just need one or two tests for make coverage happy.

backend/clubs/models.py
```diff
@@ -2086,3 +2086,44 @@ def user_create(sender, instance, created, **kwargs):
...
+        self.pk = 1
+        super().save(*args, **kwargs)
+
+    @classmethod
+    def get(cls):
+        """
+        Get or create singleton instance of QueueSettings.
+        """
+        obj, created = cls.objects.get_or_create(pk=1)
+        return obj
```
Porcupine1, line 2126: Maybe add some test to check that this model is singleton

julianweng: Agreed, should probably override the create method as well and check pk=1 for it, save, and get,

backend/clubs/models.py
```diff
@@ -2086,3 +2086,44 @@ def user_create(sender, instance, created, **kwargs):
...
+        null=True,
+        blank=True,
+        related_name="modified_queue_settings",
+    )
+
+    def save(self, *args, **kwargs):
+        """
+        Override save method to ensure this is a singleton.
+        """
+        self.pk = 1
```
julianweng, line 2117: Would prefer to explicitly check for self.pk != 1 and error if true rather than enforce it silently

julianweng: Also would it be possible to create a constant SINGLETON_PK or something to that effect for readability purposes?

julianweng (approved):
LGTM, agree with Thomas' comments and left a couple of my own. I'll integrate it into the admin interface soon.

backend/clubs/models.py
julianweng: Also ideally make it undeletable unless there's a reason for it to be deletable

gabeweng (approved):
Everything looks good to me. I really don't have many comments, but I agree with what Thomas and Julian said! Eventually it could be good to have singleton behavior in an abstract model (like the django-solo package) but not really a concern now when we only have this.

julianweng:
frontend support

@aviupadhyayula

backend/clubs/serializers.py
```diff
@@ -3112,3 +3114,6 @@ class Meta:
             "updated_by",
         ]
         read_only_fields = ["updated_at", "updated_by"]
+
+    def get_updated_by(self, obj):
+        return obj.updated_by.get_full_name()
```
aviupadhyayula, line 3119: Good catch, but this fails when there isn't an existing record. Throwing in a non-null check should fix.

aviupadhyayula:
> frontend support
> 
> @aviupadhyayula

This is great, thanks Julian. Not sure how much work it would be, but do you think we can make a separate component for the queue management (like how the components in the Club Management tab are arranged)? Ideally something separate that explicitly notes the status of each queue, and then color-coded buttons to open/close each queue below that. (I think the current design's a little prone to human error.)

julianweng:
revised design
@aviupadhyayula

backend/clubs/models.py
aviupadhyayula: Yup, good call.

Issue #725, opened 2024-09-18
Add option for club applications to require graduation year


julianweng:
Currently, as Penn Clubs does not require users to self-identify their graduation year, many club applications collect the information either through a separate question on the application or elsewhere in the process. This would add the option for clubs to mandate that people fill out their school year on the Penn Clubs platform before being able to submit their application.

This would benefit clubs, as there are less points for error (e.g. clubs forgetting to update their grad year question for new years) and reduces work for them. This would also benefit Clubs (and other clubs) as more users would be nudged to fill out their school year and increasing data availability for other applications.

There could be edge privacy cases where users would not want to expose their graduation year to other clubs or via their profile, but the former would seem rare (and impossible in practice for any club that would care about grad year) and the latter can be solved through hiding their profile. Open to discussion on this point, however.

aviupadhyayula:
@julianweng I'm thinking about this more -- is there a reason we can't just mandate this (e.g. if you want to submit an application, you need to fill out the major & grad year fields on your profile page). I have a feeling that most club leaders won't discover this option (since a lot of them just duplicate past applications). I also think it's reasonable to require these -- club leaders obviously benefit, and the knowledge is public anyway (you can just look someone up in the directory). 

I'm envisioning a serializer-level change to the application submission flow, where we check if the user has set their grad year, major, and school. If they haven't, we throw a validation error.

julianweng:
I think directory is technically voluntary (forget if it's opt-in or opt-out): this reasoning makes sense and I wouldn't be opposed to it but we should probably double-check with stakeholders before going down this path just in case.

I was thinking that we would run a one-time script that would scan current-cycle (Spring) applications and toggle the option on if certain keywords are found in the associated questions and question options (e.g. "Freshman", "2026", "graduation year", "class"), but there is the edge case where clubs might still care about grad year but ask it later on in their application process (off-platform).

aviupadhyayula:
How about we just implement it as "always-on" in the backend for now (e.g. we require the info to be set before the application submission is created)? And we can check in with club leaders and OSA at our next meeting. This will let @yuviji get started.

yuviji:
Will implement it as @aviupadhyayula suggested with the "always-on" setting:
- First-time Clubs users on the `/welcome` page will be required to enter grad year AND school AND major
- Returning users that press "Apply" on a club's page without having their details in will be redirected to `/profile` and on `/profile` will get a toast that tells them that they need to add their details in order to be able to apply to clubs

PR #792, opened 2025-03-21
Require completed profile for application submissions

yuviji:
#725 --> force users to set grad year, major, school in order to apply to application-based clubs


yuviji:
i think it's more appropriate to add validation in the view itself before we scan through the rest of the stuff, and stop it right there if the user profile is incomplete. from there, i return a 400 and redirect the user to the settings/profile on the frontend.

backend/clubs/views.py
```diff
@@ -6392,8 +6392,25 @@ def question_response(self, *args, **kwargs):
...
+        if (
+            not user.profile.graduation_year
+            or not user.profile.school.exists()
+            or not user.profile.major.exists()
+        ):
+            return Response(
+                {
+                    "success": False,
+                    "detail": """You need to set your graduation year, major, and
+                    school before you can apply to a club!""",
```
aviupadhyayula, line 6407: Nit: is there a reason this is in triple quotes?

aviupadhyayula (commented):
Thanks for tackling this Yuvraj. Left a nit but otherwise backend changes LGTM.

backend/clubs/views.py
yuviji: Just copied Gupta's format 💀 can make it single quotes 👍🏼

aviupadhyayula: Sounds good, no need to change it.

aviupadhyayula (approved):
LGTM, good work Yuvraj. Looks like there's a merge conflict – once it's resolved, feel free to push this out.

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
aviupadhyayula, line 4: Just a heads up to delete this before you merge the PR in.

PR #794, opened 2025-03-23
Remove repo artifacts

aviupadhyayula:
Removes various artifacts from previous development from the codebase.


aviupadhyayula:
Closing -- sounds like there are plans to use dev container moving forwards.

PR #797, opened 2025-03-23
Fix hardcoded outdated language in renewal email

julianweng:
Currently, renewal emails and renewal reminder emails reference SAC fair registration at a specific time as well as assume the queue is open. This is bad for multiple reasons, as the templates have to be manually changed every year according to circumstances, and SAC has also largely moved off of our platform for fair registration as the SAC fair is no longer virtual.

This PR adds an argument to the deactivate script which sends renewal emails to optionally add when the approval queue should be re-opened by (for use at the beginning of the school year) as well as removes language claiming that Penn Clubs registration is sufficient for SAC fair registration (it is only necessary). It also removes the renewal reminder email, as reminder emails and miscellaneous emails in general without side-effects should be handled through the new dedicated email blast functionality.


aviupadhyayula (commented):
LGTM, thanks for taking the initiative on this Julian. Adding onto this: should we make the yearly deactivation a cron-job, and move deactivation from a manual event to an automatic one? I think this could create a more defined cadence for renewals.

backend/clubs/management/commands/deactivate.py
```diff
@@ -26,14 +26,22 @@ def add_arguments(self, parser):
...
+            help="If this parameter is specified, then the approval queue is not "
+            "currently open and clubs will be notified of the specified open date "
+            "in plain text (e.g. put the string 'August 24, 2025').",
+        )
 
         parser.add_argument(
             "action",
             default="all",
-            choices=["deactivate", "emails", "all", "remind"],
+            choices=["deactivate", "emails", "all"],
```
aviupadhyayula, line 41: Is there a reason we're keeping only emails as an option? If we're removing reminders, it makes sense to me to remove emails as well. (Might as well overhaul the script if we're cleaning it up.)

julianweng: Yeah I would be open to that, I think the email option is sort of nice because it has a button that redirects to a specific club's renewal link but I think that we can have emails be a binary flag instead of a 3-way option (since I'm not sure if sending just emails would be a common usecase that can't be handled through the email blast feature).

julianweng:
I think making it a cron would be a good idea, I'm a little concerned about lack of visibility (e.g. forgetting that the cron exists) but as long as we proactively set the cron ahead of every renewal cycle with OSA feedback this should be good.

aviupadhyayula:
Fair, let's hold off on the cron for now then. We'll ask OSA about it next meeting. I think the ideal would be to choose a date before the school year starts (e.g. August 15 or something), and every year all clubs are deactivated then.

PR #798, opened 2025-03-26
Implement discount (promo) codes

julianweng:
* Implement auto-generation of discount code
* Fix implementation of group discounts
* Misc mobile responsivity improvements

adding discounts

seeing selling information for a given ticket

checkout


aviupadhyayula (commented):
Took a look at the backend and left a couple comments. At a high level: I'm not a huge fan of moving pricing-related stuff (group discounts, promo codes) under the `Ticket` class. It feels crowded and sort of against the principles of OOP. 

The way I was envisioning we'd implement this is as a separate model (e.g. `TicketPromoCode`), with a link back to the ticket type. When a user checks out (or when we call the `cart` route), they apply as many promo codes as they want through the frontend. We check if the promo codes are valid, and then use them in the cart total calculation. 

This feels cleaner to me, and also unlocks more capability for event owners. With the current approach, event owners can't create multiple promo codes per ticket. 

I think the work Rohan was doing with the `Ticket` and `TicketClass` models is kind of what we want. I don't know what the best course of action here is: should we wait until the checkout flow rewrite's done, and then migrate this work over to that branch?

Let me know what you think. (One last nit: we'll be merging this PR into `ticketing-v2`, so we'll need to change the target branch in the PR.)

PR #799, opened 2025-03-27
Use built-in Next.js servers, bump Node/Bun/Next versions, use Bun for React check

julianweng:
Something in the Bun migration made the development server very slow. This attempts to fix it by bumping Node/Bun/Next versions and moving off of custom `server.js` (which doesn't do much anymore) in favor of Next's built-in dev and prod servers.

Also migrates off of shared actions in order to use Bun for React check.


aviupadhyayula (approved):
LGTM, tested locally. Thanks for shipping all over the codebase goat.

PR #800, opened 2025-03-28
Use uv

julianweng:
Implements uv as the backend package manager similar to CFA. #782 seems to be blocked on waypoint / shared-actions related issues and frontend CI/CD is already decoupled from shared-actions so this makes sense for the time being. During onboarding this semester, first-time Pipenv installation seemed really slow (felt like a regression qualitatively from that of previous Python versions), so switching to a modern package manager should hopefully help with that.


.github/workflows/build-and-deploy.yaml
```diff
@@ -7,16 +7,54 @@ on: push
...
-      ruff: true
+    name: Backend Check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Cache
+        uses: actions/cache@v4
+        with:
+          path: .venv
+          key: v0-${{ hashFiles('./uv.lock') }}
```
aviupadhyayula, line 18: Do we still need to cache dependencies? `uv`'s fast enough that we can clean re-install each time. I remember a broken actions cache being a problem a while ago.

.github/workflows/build-and-deploy.yaml
```diff
@@ -7,16 +7,54 @@ on: push
...
+    env:
+      DATABASE_URL: postgres://postgres:postgres@postgres:5432/postgres
+    services:
+      postgres:
+        image: postgres:17
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_DB: postgres
+          POSTGRES_PASSWORD: postgres
+        options: "--health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5"
```
aviupadhyayula, line 57: Is there a reason we have to include this DB config stuff?

aviupadhyayula (approved):
LGTM, thanks for taking the initiative on this Julian. Left a couple of questions I had. Feel free to close out #782 once this hits main.

.github/workflows/build-and-deploy.yaml
julianweng: So true Avi

PR #793, opened 2025-03-22
Support sub-events: EventGroup approach

Porcupine1:
This PR resolves issue #755

PR #802, opened 2025-03-29
Refactor events model to support sub-events

### LINKED ISSUE #755: Add support for sub-events on different dates
Certain events will have multiple "showings" (such as a performance that happens on both Friday and Saturday). When checking out, buyers should be able to choose which date they're purchasing a ticket for. 

To do this, we'll have to change how events work on the backend. (A first pass might be to make a parent class representing events, and then a one-to-many relationship from an event to its distinct showings.)

Porcupine1:
Refactors how events are represented in the backend to allow support for sub-events (necessary for ticketing). Introduces model `EventGroup` to stand in for what's currently `Event`, and breaks out sub-events into their own, new `Event` model.

Closes #755.


julianweng:
Closing for #793

PR #806, opened 2025-04-05
Add (and migrate to) event showings (V2)

aviupadhyayula:
(Second attempt at migrating over to the `EventShowing` model. Addresses failure from 64f4c4f by keeping `ZoomMeetingVisit` foreign key'ed on `Event`.)

Introduces an `EventShowing` model to abstract out start times, end times, and locations from `Event`. This allows for multiple showings of the same event (e.g. a play that has showings on Friday and Saturday).

Other changes: 
1. Removes `RecurringEvent`, which can now instead be represented as an `Event` with multiple `EventShowing` instances. Confirmed that there are 0 recurring events in prod currently.
2. Adds serializers, permissions, and urls specific to `EventShowing`.
3. Since users can't filter events directly by start and end times anymore, adds fields to the events queryset to support filtering (`earliest_start_time`, `latest_start_time`, `earliest_end_time`, `latest_end_time`). 
4. Changes `Ticket` from pointing to `Event` to now pointing to `EventShowing`. Given this change, all ticketing-related logic has been moved from `ClubEventViewSet` to `ClubEventShowingViewSet`.
5. Refactors unit tests and commands to comply with changes..

I've made the event showings API available in two places: `/api/events/<event_id>/showings`, and `/api/clubs/<club_id>/events/<event_id>/showings`. New showings and updates must occur through the second (enforced by permissions). This is in line with how events work currently.

Since this is a breaking change to the DB, deprecated attributes (e.g. `Event.start_time`) haven't been removed. Once this PR is merged in, we'll populate `EventShowing` objects for all past events, and then remove the deprecated attributes from the schema. 

**(To reviewers)** I'd recommend reviewing this PR in the following order: models (specifically the new `EventShowing` model and changes to `Event` and `Ticket`), serializers, permissions, and views. Ignore "alignment" changes, and focus only on `EventShowing` and `Event`. Afterwards, examine ticketing-related changes. Finally, review the refactor of other models, views, commands, and tests.


julianweng (approved):
Thanks lgtm!

PR #807, opened 2025-04-05
Complete migration to event showings

aviupadhyayula:
Cleans up legacy attributes from schema (as per #806). I've run management commands in prod to create an `EventShowing` for each existing `Event` in the DB, and to migrate all tickets in the DB to the corresponding showing -- so all data should be migrated correctly.


backend/clubs/migrations/0121_remove_event_parent_recurring_event_and_more.py
```diff
@@ -0,0 +1,50 @@
...
+            name='ticket_order_limit',
+        ),
+        migrations.RemoveField(
+            model_name='ticket',
+            name='event',
+        ),
+        migrations.AlterField(
+            model_name='zoommeetingvisit',
+            name='event',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='visits', to='clubs.event'),
```
julianweng, line 45: Is this a formatting or a github rendering issue?

julianweng (commented):
LGTM! Can you add some color on what we're doing with zoommeetingvisit? Seems like we made it nullable in the last pr and are making it non-nullable here and was wondering if we did anything with it in the migration script.

backend/clubs/migrations/0121_remove_event_parent_recurring_event_and_more.py
aviupadhyayula: It looks like `ZoomMeetingVisit.event` was first updated to be nullable and then non-nullable in #806, but the migration file wasn't updated. Running `makemigrations` with this PR's changes causes that change to be applied now instead of in the last PR.

PR #812, opened 2025-04-13
Ownership request status

gabeweng:
We want to keep track of ownership requests that are accepted or denied as opposed to deleting them when they are handled. This is done by replacing the "withdrawn" field with a "status" field, which can be PENDING, WITHDRAWN, ACCEPTED,  or DENIED.


backend/clubs/views.py
```diff
@@ -4389,20 +4391,33 @@ def create(self, request, *args, **kwargs):
...
-            requester=request.user,
-            defaults={"withdrawn": False, "created_at": timezone.now()},
-            create_defaults=create_defaults,
-        )
-
-        if created:
-            obj.send_request(request)
+        request_instance = OwnershipRequest.objects.filter(
+            club__code=club, requester=request.user
+        ).first()
```
aviupadhyayula, line 4396: Nit: would prefer `obj`.

backend/clubs/views.py
```diff
@@ -4389,20 +4391,33 @@ def create(self, request, *args, **kwargs):
...
+        elif request_instance.status == OwnershipRequest.PENDING:
+            return Response(
+                {"detail": "Ownership request to this club has already been sent"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        else:
+            return Response(
+                {"detail": "Request has already been handled"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
aviupadhyayula, line 4418: Should we let requesters resubmit after some period of time? I can see a world in which someone's ownership request is denied initially but approved a year later (e.g. when a club goes defunct).

backend/clubs/views.py
```diff
@@ -4484,9 +4500,16 @@ def accept(self, request, *args, **kwargs):
             defaults={"role": Membership.ROLE_OWNER},
         )
 
-        request_object.delete()
+        request_object.status = OwnershipRequest.ACCEPTED
+        request_object.save(update_fields=["status"])
```
aviupadhyayula, line 4504: Just realized: do we update requesters when the status of their request is updated? Feels like we should do that (if we don't already).

aviupadhyayula (commented):
LGTM, thanks for taking the initiative on this Gabe. Left a couple comments. 

Also, is this compatible with @beladelgado2's work in #815? If not, let's add the necessary frontend changes to this PR before merging.

backend/clubs/views.py
gabeweng: I fixed it 180 days for now.

backend/tests/clubs/test_models.py
```diff
@@ -284,3 +286,210 @@ def test_send_mail_retry_logic(self):
                 )
 
             self.assertEqual(mocked_send.call_count, 3)
+
+
+class OwnershipRequestTestCase(TestCase):
```
gabeweng, line 291: Honestly I'm not sure how necessary it is to test the business logic in both model unit tests and view tests. If I get the go-ahead I will probably just delete all of these model tests.

gabeweng:
@Porcupine1 can you take a look when you get the chance?

backend/tests/clubs/test_models.py
Porcupine1: @julianweng thoughts?

backend/clubs/models.py
```diff
@@ -1205,9 +1205,56 @@ class OwnershipRequest(JoinRequest):
...
+    STATUS_TYPES = (
+        (PENDING, "Pending"),
+        (WITHDRAWN, "Withdrawn"),
+        (DENIED, "Denied"),
+        (ACCEPTED, "Accepted"),
+    )
+    status = models.IntegerField(choices=STATUS_TYPES, default=PENDING)
+
+    class Meta:
+        unique_together = ()
```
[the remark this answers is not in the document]
Porcupine1: refering to:
```
class Meta:
        unique_together = ()
```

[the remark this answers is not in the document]
gabeweng: Hi Thomas, basically all it does is override the uniqueness constraint that the parent abstract model JoinRequest has in it's Meta class. I can change it so that JoinRequest no longer has a uniqueness constraint and MembershipRequest (another subclass of JoinRequest) specifically has a uniqueness constraint.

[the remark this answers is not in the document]
Porcupine1: I see. In that case, I think this is fine. Maybe just add a comment to make this clear for the next person

PR #813, opened 2025-04-14
Ticketing checkout flow rewrite

rm03:
This PR implements a major refactor of the ticketing checkout flow. Key changes include allowing anonymous (guest) users to checkout, simplifying the checkout flow by removing cart holds, adding the `CartItem` and `TicketClass` models for a more modular representation of ticketing data.

TODO:
- [ ] Finish manually testing integration with `EventShowing`
- [ ] Update tests
- [ ] Determine which Cybersource API we need to use (Unified Checkout vs. Secure Acceptance)
- [ ] Add cronjob to clean up old carts


backend/clubs/serializers.py
```diff
@@ -1896,8 +1901,9 @@ class TicketSerializer(serializers.ModelSerializer):
     """
 
     owner = serializers.SerializerMethodField("get_owner_name")
-    showing = EventShowingSerializer()
-    event = EventSerializer(source="showing.event")
+    event = EventSerializer(source="ticket_class.showing.event", read_only=True)
```
Porcupine1: here, wouldn't something more specific like event showing be more desirable?

PR #815, opened 2025-04-25
Add Ownership Request Flow for Inactive Clubs

beladelgado2:
Student-Side Changes:
- Adds a “Request Ownership” button to all inactive club pages.
- Submits a request via /api/requests/ownership/ using authenticated user info.
- Displays confirmation message upon successful submission.

Admin-Side Changes:
- Adds “Pending Ownership Requests” section to /admin/queue:
- Displays club name, requester name, request date, and approve/reject buttons.
- Approving sends a POST to the backend to make requester an owner.
- Rejecting deletes the request and updates the queue.

PR #816, opened 2025-05-05
Fix route for listing all ownership requests

gabeweng:
I noticed that the frontend PR for ownership requests #815 was calling the  `/api/requests/ownership` endpoint when listing all ownership requests for admins to see. The route that actually provides this functionality is `/clubs/any_club_code/ownershiprequests/all`.  As a result, ownership requests made by users other than the currently logged-in user were not appearing in the admin dashboard. This PR so far just updates the frontend to use the correct route so all relevant ownership requests are displayed.

### DIFF
--- a/frontend/components/Settings/QueueTab.tsx
+++ b/frontend/components/Settings/QueueTab.tsx
@@ -501,7 +501,7 @@ const QueueTab = (): ReactElement => {
     if (canApprove) {
       refetchClubs()
 
-      doApiRequest('/requests/ownership/?format=json')
+      doApiRequest('/clubs/any/ownershiprequests/all/?format=json')
         .then((resp) => resp.json())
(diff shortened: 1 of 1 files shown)


gabeweng:
Having the route for listing all ownership requests under `/clubs/any_club_code/ownershiprequests/all` feels unintuitive since it is nested under `/clubs/` but does not pertain to any single club. I am considering defining a separate View for this functionality, rather than having it grouped together with club-level ownership request management views under `OwnershipRequestManagementViewSet` as we do now.

aviupadhyayula:
I think that makes sense. I think it makes sense to have `/clubs/<code>/ownership-requests/` return all the ownership requests submitted to the `<code`> club, and have a separate route that lists all ownership requests (for admin to view). (AFAIK we do something similar with events.)

PR #817, opened 2025-05-06
Fix Ownership Requests

gabeweng:
This PR is aimed at getting Ownership Requests to a completely functional state.

1) #816 was incomplete since the frontend expects `club_name` as a field that is returned for each club and `all_requests` didn't have that. This PR ensures that this expectation is met by adding a new `club_name `field in `OwnershipRequestSerializer`

Before:
ownershipRequestQueueBefore

After:
ownershipRequestQueueAfter

2) Every time an ownership request is made for the first time, an email is sent to current club owners letting them know of the request and linking to `/clubs/{club_name}/edit/members`. This was done with the assumption that owners of clubs would be able to approve or deny requests in the frontend (like with membership requests). Since this does not exist , this PR removes the mail notification from ownership requests.

### DIFF
--- a/backend/clubs/views.py
+++ b/backend/clubs/views.py
@@ -4518,9 +4518,6 @@ class OwnershipRequestViewSet(viewsets.ModelViewSet):
             create_defaults=create_defaults,
         )
 
-        if created:
-            obj.send_request(request)
-
         serializer = self.get_serializer(obj, many=False)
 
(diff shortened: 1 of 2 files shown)


aviupadhyayula (commented):
Thanks for taking the initiative on this Gabe. Left a comment but otherwise LGTM.

PR #819, opened 2025-07-05
Use ASGI server exclusively in production on Django backend

julianweng:
In the spirit of the July 4th long weekend, Penn Clubs is declaring independence from uwsgi and wsgi more generally. Having a dedicated asgi server specifically for websockets and wsgi for everything else, especially when the number of replicas for each is quite small (e.g. one asgi instance) and websockets aren't used very much in the current feature set, leads to unnecessary resource usage from this lack of granularity. Maintaining multiple endpoints also leads to more complexity in the codebase, and separately, the act of installing uwsgi by itself in development during onboarding has wasted dozens of manhours 😬.

This PR performs the following:

* Streamlines old routing logic to expose a single ASGI endpoint for both http and websocket requests
* Reconfigures CI/CD to serve ASGI instances only using Uvicorn workers with Gunicorn
* Adds the [Blacknoise](https://github.com/matthiask/blacknoise) library to handle static asset serving ~efficiently in place of uwsgi's built-in functionality (I believe this is mostly just used for hosting css/js for the Django admin page)
* Removes uwsgi dependencies and WSGI endpoint scaffolding


Porcupine1 (approved):
had to read up on what asgi & wsgi is lol
LGTM

PR #832, opened 2025-08-07
Mandate Constitution Upload

### LINKED ISSUE #825: Enforce stricter constraints on certain fields
Limit club mission statement length (and perhaps include more strongly-worded language that it should focused on club mission)

Make uploading a club constitution mandatory

Porcupine1:
This PR mandates that clubs upload their constitutions. Uploading a constitution file can be done via resource tab, in which case the file name should at least contain `constitution` as a substring. The file upload section could potentially be used to upload files that aren't intended to be the constitution. One can also upload a constitution file via the `ClubEditCard` on `/create/` `.../renew/` `.../edit/` routes where we assume the user definitely intends to upload a constitution.

On the resource tab, you'll see some warning text that is conditionally rendered for clubs that don't satisfy the constitution requirement.

NOTE: club mission word limit was implemented in an earlier commit not in this PR
Closes #825


julianweng (changes requested):
Thanks for your work Thomas. Left some nits. On a higher level, do you think it could be more elegant to treat constitutions as a foreign key for the club model vs an attribute on a given file upload? It seems like a useful invariant that a club can have only at max one constitution, and there are likely performance benefits as well (it's also easy to `prefetch_related` or treat this similarly to club photo in terms of our business logic). In addition, especially if we move to this alternative, constitution uploads should be made visually distinct in the UI from normal, non-constitution file uploads and the experience shouldn't feel different from, say, the club photo requirement or field in club edit.

We could also potentially kill the file upload feature altogether in favor of this since I'm not sure if it has any purpose other than constitutions, but that's likely a question for later and we'd want to give clubs time to download their existing files and reupload them as constitutions.

Also, have not gotten a chance to test this locally, backend or frontend (travelling), so will trust that you have!

Porcupine1:
addressed in #875

Issue #568, opened 2023-01-08
application.description is vulnerable to XSS


rohangpta:

julianweng:
Seems like we [already sanitize](https://github.com/pennlabs/penn-clubs/blob/837219bf02ca2eaf27e78cb70546fc9728b87f6f/backend/clubs/serializers.py#L1654) any html fields, is this more related to this sanitization step being insufficient or the nature of `dangerouslySetInnerHTML`?

rohangpta:
Looks like we have some sanitization but not for the entire surface area. GPT 5 via the Cursor CLI fixed the rest in #842

PR #842, opened 2025-09-15
fix: harden HTML sanitization for applications, fairs, and imported events (XSS mitigation)

rohangpta:
#568 

- Sanitize HTML for content rendered via `dangerouslySetInnerHTML` by cleaning on the backend.
- Add allowlist-based sanitization for Club Applications and Club Fairs.
- Sanitize third-party imported event descriptions (SNF Paideia) before saving.


julianweng (approved):
LGTM, thanks for looking into this

PR #846, opened 2025-10-31
New interface for the main page (newbie project)

LautaroJBeck:
### **New interface for my newbie project**
This UI is meant to give more visibility to the Events of PennClubs
I renamed the old "index.tsx" file to "clubs.tsx". I modified the Header and added a new link to redirect users to /clubs.
Screenshot 2025-10-31 at 5 16 25 PM

### DIFF
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -162,6 +162,7 @@ const InternalSearchContainer = styled.div`
   background-color: ${WHITE};
   border: 1px solid #e0e0e0;
   box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+  overflow: hidden;
 `
 const SearchInput = styled.input`
   width: 100%;
@@ -462,8 +463,6 @@ const SmallLogo: React.FC<{ image?: string; name?: string }> = ({
   return <ClubLogo>{name?.charAt(0)?.toUpperCase() || 'C'}</ClubLogo>
 }
 
-const ClubResultsContainer = styled.div``
-
 const NoResults = styled.div`
   padding: 1rem 1.25rem;
   color: ${CLUBS_GREY_LIGHT};
@@ -492,9 +491,17 @@ const ClubResult = styled.div`
     background-color: ${LIGHTER_BLUE};
     transform: translateY(-1px);
   }
+`
 
-  &:last-child {
+const ClubResultsContainer = styled.div`
+  /* only apply rounding/styling to the actual last result element */
+  & > *:last-child ${ClubResult} {
     border-bottom: none;
+    border-radius: 0 0 16px 16px;
+  }
+
+  & > *:last-child ${ClubResult}:hover {
+    transform: none;
   }
 `
 


frontend/pages/index.tsx
```diff
@@ -1,648 +1,655 @@
...
+  max-width: 800px;
+  padding: 0 2rem;
+`
+const InternalSearchContainer = styled.div`
+  border-radius: 25px;
+  background-color: ${WHITE};
+  border-radius: 25px;
+  border: 1px solid #e0e0e0;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+  &::placeholder {
```
julianweng: Double check that this does something since `InternalSearchContainer` is a `div` and not an `input`

frontend/pages/index.tsx
```diff
@@ -1,648 +1,655 @@
...
+    color: ${CLUBS_GREY_LIGHT};
   }
 `
-
-const Divider = styled.div`
+const SearchInput = styled.input`
   width: 100%;
-  height: 3px;
+  padding: 1rem 3rem 1rem 1rem;
+  border: 0px;
```
julianweng: `border: none`

frontend/pages/index.tsx
```diff
@@ -1,648 +1,655 @@
...
-              }}
-            />
-
-            {(preFair || fairIsOpen) && (
-              <LiveEventsDialog
-                isPreFair={!!preFair}
-                isFair={!!fairIsOpen}
-                isVirtual={!!fairIsVirtual}
-              />
+            {isSearchFocused && clubs.length > 0 && (
```
julianweng: We should show some sort of text when the search query returns but with zero clubs to show the user that the query is done and hasn't errored, but just doesn't return any valid club.

frontend/pages/index.tsx
```diff
@@ -1,648 +1,655 @@
...
+          <Logo src={SITE_LOGO} alt={`${SITE_NAME} Logo`} />
+          <Title>{SITE_NAME}</Title>
+        </LogoContainer>
+        <Subtitle>
+          Student Organizations at the University of Pennsylvania
+        </Subtitle>
+        <SearchContainer>
+          <InternalSearchContainer>
+            <SearchInput
+              placeholder="Explore Your Favorite Clubs Here"
```
julianweng: Add an `aria-label` for accessibility / screen readers

frontend/pages/index.tsx
```diff
@@ -1,648 +1,655 @@
...
-            {SHOW_SEARCHBAR_TOP && (
-              <>
-                <TopSearchBar
-                  onChange={(value) =>
-                    setSearchInput((inpt) => ({ ...inpt, search: value }))
-                  }
-                />
-                <ScrollTopButton />
-              </>
+            <SearchIcon>
```
julianweng: This should be a button (or wrapped in one) for accessibility if clicking on it does anything

julianweng (changes requested):
This is a great first PR! I left some inline comments to address, also will leave some general design comments below

* In mobile, consider putting "See More" on its own line slightly below Upcoming Events.
* Same for retry and "Search failed. Please try again."
* The layout for individual events looks sort of weird (image seems very short with a lot of whitespace while the description is squished) for mobile, maybe we should rethink this or use some sort of card format with the image inscribed on top, something like [this](https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRvJ4u3uA0sfQFU2WnRbe-01rNuWecs-4G7NQ&s) maybe)
* Clubs have a lot of badges and tags on the actual site: the tags might look weird / stretch too far down vertically if not limited somehow in mobile
* The club names are center aligned in mobile which looks kind of weird, especially when viewed with the club logo and tags aligned left.
* The search bar and events section are sort of narrow on mobile and can take up more width (at least that of the "Student organizations..." tagline)
* The blue color when hovering on clubs is a little faint
* Ideally don't show the clubs section if nothing is typed into the search bar
* Hitting enter should navigate to the club's specific page if there's only one search result for the query and to the clubs search page with query prepopulated otherwise. Hitting enter while the query is blank shouldn't do anything.

frontend/pages/index.tsx
```diff
@@ -1,648 +1,766 @@
...
-      window.clearTimeout(searchTimeout.current)
-    }
-    setSearchValue(value)
-    searchTimeout.current = window.setTimeout(() => {
-      searchTimeout.current = null
-      onChange(value)
-    }, 100)
+const RetryButton = styled.button`
+  background: transparent;
+  border: 0px;
```
julianweng: I missed this last time I think but border should be `none` here as well

PR #848, opened 2025-11-02
Fix infinite loading on applications page for clubs without applications

dkwncho:
Currently, I have it so that we show an infinite loading screen upon api error. 
Probably worth showing some error text instead, but I didn't bother, lmk if I should


julianweng (approved):
LGTM

Issue #835, opened 2025-08-19
Add Option to set Club Status on Approval


Porcupine1:
The Frontend has `ClubApprovalDialog.tsx` where and admin with the right permission can approve a club. On this dialog, we want to allow the approver to set the `status` the club should have after it is approved. The point of this to make it easier to set the `status` field because currently we can only set the status on the Club Edit Page and on `/api/admin/`.

It might be a good idea to leave out `Under Review` as an option as a club no longer in the queue should never have this status value.

This issue is fullstack, so you'll need to make some backend and frontend changes/additions.

PR #851, opened 2025-11-11
Moving towards a status-based approval system

yuviji:
Addresses #835 (and a bit more?)

We set a club's `status` during approval and auto set status to “Under Review” on queue entry. The Admin Queue can set status in bulk as well. This PR moves us towards using the `status` field as a single source of truth, conditioning, and validation - currently, we scatter our judgement across `active`, `approved`, and `archived` fields for a Club.

**Backend**
- Add Status listing endpoint and status-driven mapping in serializers:
  - Officers can set Status to “Under Review” for their club renewals only; admins can set any Status.
- Status mapping to state, making many assumptions with guidance from the PennClubs Descriptors Excel file:
  - FULL/PRELIMINARY/PROVISIONAL → approved=True, active=True
  - SUSPENDED/INACTIVE → approved=False, active=False
  - TEMPORARY → approved=False, active=True
  - UNDER REVIEW → approved=None, active=True; clears approved_by/approved_on
  - DEFUNCT → archived_on set, active=False
- Allow approvers to update clubs:
  - Update object permissions so users with clubs.approve_club can PATCH clubs.
- Serializer fields: added read-only `status` everywhere; introduced `status_id` (write-only) in AuthenticatedClubSerializer and AdminClubSerializer; made `approved`, `approved_comment` writable in AdminClubSerializer

**Frontend**
- Club Approval Dialog (`ClubApprovalDialog.tsx`):
  - Change from Accept, Reject, Delete buttons to a single a Status dropdown (excluding “Under Review”).
  - Button labeling/styles/icon dynamically reflect the chosen status (Approve/Reject/Delete).
  - “Revoke Approval” sets Status to “Under Review” and records a reason.
- Admin Approval Queue (`Settings/QueueTab.tsx`):
  - New “Change Status” bulk modal with templates + comment, and a Status dropdown (excluding “Under Review”).
  - Applies status and comment to all selected clubs; refreshes table.
- Status UI helpers (`constants/status.ts`):
  - Centralize allowed status categories, the “UNDER REVIEW” constant, and UI action mapping (label, button class, icon).

### DIFF
--- a/backend/clubs/serializers.py
+++ b/backend/clubs/serializers.py
@@ -2702,7 +2702,9 @@ class AuthenticatedClubSerializer(ClubSerializer):
                     # archive and deactivate the club
                     if self.instance:
                         self.validated_data["archived"] = True
-                        self.validated_data["archived_by"] = request.user if request is not None else None
+                        self.validated_data["archived_by"] = (
+                            request.user if request is not None else None
+                        )
                         self.validated_data["archived_on"] = timezone.now()
                         self.validated_data["active"] = False
                 elif status_name in ["FULL", "PRELIMINARY", "PROVISIONAL"]:


julianweng:
Thanks Yuvraj, here is what Codex says about your PR

---

**High-Level Findings**
- `approved` is writable for admins but not treated as an approval field; an admin can `PATCH {"approved": ...}` without `status_id`, bypassing status mapping and leaving `status/approved/active` inconsistent (backend/clubs/serializers.py:2752-2774; backend/clubs/views.py:2096-2109). Make `approved` read-only for admins and/or include it in `approval_fields`.
- Leaving `DEFUNCT` status doesn’t clear `archived`/`archived_on` (and possibly approval), so a club stays archived after setting status to Full/Preliminary/etc. Either disallow exiting Defunct or clear archive/approval when transitioning away (backend/clubs/serializers.py:2700-2727).
- Officer “Under Review” flows silently no-op if the status is missing: “Request Review” and creation completion send an empty PATCH and reload without error when lookup fails (frontend/components/ClubPage/ClubApprovalDialog.tsx:497-517; frontend/components/ResourceCreationPage.tsx:349-370). Surface an error and skip the PATCH, or guarantee the status exists via migration/seed.

**Nits / Other Comments**
- `getStatusActionInfo` can throw on empty/undefined; add a guard or document non-empty input (frontend/constants/status.ts).
- Bulk modal button falls back to “Approve” when no status is selected; prefer a disabled/neutral state without implying an action (frontend/components/Settings/QueueTab.tsx).
- Repeated “Under Review” fetch logic should be centralized (renew.tsx, ResourceCreationPage.tsx, ClubApprovalDialog.tsx).
- Remove the commented-out `validate_active` block (backend/clubs/serializers.py ~1750).
- Approval error message uses a set; sort for deterministic ordering (backend/clubs/views.py).
- The IIFE around `handleAction` in ClubApprovalDialog adds noise; define handlers normally (frontend/components/ClubPage/ClubApprovalDialog.tsx).
- Clarify the approved-status mapping comment to mention approval metadata is set in the parent save (backend/clubs/serializers.py ~2710).

backend/tests/clubs/test_views.py
```diff
@@ -1265,23 +1220,384 @@ def test_club_approve(self):
...
+            self.club1.refresh_from_db()
+            self.assertEqual(self.club1.status.id, status.id)
+            if expected_approved is not None:
+                self.assertEqual(self.club1.approved, expected_approved)
+            else:
+                self.assertIsNone(self.club1.approved)
+            self.assertEqual(self.club1.active, expected_active)
+            self.assertEqual(self.club1.archived, expected_archived)
+
+    def test_status_full_maps_to_approved_active(self):
```
Porcupine1: `test_admin_can_set_any_status` already checks this

backend/tests/clubs/test_views.py
```diff
@@ -1265,23 +1220,384 @@ def test_club_approve(self):
...
+            content_type="application/json",
+        )
+        self.assertIn(resp.status_code, [200, 201], resp.content)
+
+        self.club1.refresh_from_db()
+        self.assertFalse(self.club1.approved)
+        self.assertFalse(self.club1.active)
+        self.assertEqual(self.club1.status.name, "Suspended")
+
+    def test_status_under_review_maps_to_pending_active(self):
```
Porcupine1: `test_admin_can_set_any_status` already checks this

backend/tests/clubs/test_views.py
```diff
@@ -1265,23 +1220,384 @@ def test_club_approve(self):
...
+            content_type="application/json",
+        )
+        self.assertIn(resp.status_code, [200, 201], resp.content)
+
+        # Verify status is set to "Under Review"
+        self.club1.refresh_from_db()
+        self.assertEqual(self.club1.status.name, "Under Review")
+        self.assertIsNone(self.club1.approved)
+
+    def test_admin_can_modify_status_id_alone(self):
```
Porcupine1: this test unnecessary. previous tests should be sufficient

backend/tests/clubs/test_views.py
```diff
@@ -1265,23 +1220,384 @@ def test_club_approve(self):
...
+            reverse("clubs-detail", args=(self.club1.code,)),
+            {"status_id": self.status_full.id},
+            content_type="application/json",
+        )
+        self.assertIn(resp.status_code, [200, 201], resp.content)
+
+        self.club1.refresh_from_db()
+        self.assertEqual(self.club1.status.id, self.status_full.id)
+
+    def test_admin_can_modify_approved_comment_alone(self):
```
Porcupine1: unnecessary test

backend/tests/clubs/test_views.py
```diff
@@ -1265,23 +1220,384 @@ def test_club_approve(self):
...
+            reverse("clubs-detail", args=(self.club1.code,)),
+            {"approved_comment": "This is a test comment"},
+            content_type="application/json",
+        )
+        self.assertIn(resp.status_code, [200, 201], resp.content)
+
+        self.club1.refresh_from_db()
+        self.assertEqual(self.club1.approved_comment, "This is a test comment")
+
+    def test_admin_can_modify_both_approval_fields_together(self):
```
Porcupine1: unnecessary test

backend/tests/clubs/test_views.py
```diff
@@ -2626,11 +2876,11 @@ def test_club_sensitive_field_renew(self):
         club.save(update_fields=["approved"])
         resp = self.client.patch(
             reverse("clubs-detail", args=(club.code,)),
-            {"approved": None},
+            {"status_id": self.status_under_review.id},
             content_type="application/json",
         )
         self.assertIn(resp.status_code, [400], resp.content)
-        club.approved = True
+        club.approved = None
```
Porcupine1: might be missing something here, but this changes doesn't seem to address anything

backend/tests/clubs/test_views.py
```diff
@@ -2644,6 +2894,10 @@ def test_club_sensitive_field_renew(self):
         queue_settings.reapproval_queue_open = True
         queue_settings.save()
 
+        # Reset club to approved state before checking history
+        club.approved = True
+        club.save(update_fields=["approved"])
+
```
Porcupine1: reverting change in line 2883, makes this unnecessary

backend/tests/clubs/test_views.py
Porcupine1: why was this removed?

backend/clubs/serializers.py
```diff
@@ -1751,27 +1761,27 @@ def validate_youtube(self, value):
...
+    #         return value
 
-        # otherwise, can't edit this field
-        raise serializers.ValidationError(
-            "You do not have permissions to change the active status of the club."
-        )
+    #     # otherwise, can't edit this field
+    #     raise serializers.ValidationError(
+    #         "You do not have permissions to change the active status of the club."
+    #     )
```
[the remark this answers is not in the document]
Porcupine1: can delete since we no longer receive  `active` from user, but now derived from `status`. this logic should now be moved to `status` with adjustments like officers can't set their own status, but it's automatically set for them on renewal

backend/clubs/serializers.py
Porcupine1: this would also become deadcode. `approved` in `read_only` for non-admin. `approved` would only ever be in `self.validated_data` if an admin included it in a request, would this would not trigger a reapproval since admin has approval perms

Porcupine1: add test for this to ensure change is non-breaking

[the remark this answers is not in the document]
Porcupine1: add test for this to ensure change is non-breaking

Issue #843, opened 2025-09-23
Fix Zoom webhook verification


julianweng:
As [Zoom Webhook verification tokens](https://developers.zoom.us/changelog/platform/verification-changes-webhook/) have been deprecated and then sunset in favor of alternate logic (new secret token + header), Clubs' webhook verification logic must be [updated](https://developers.zoom.us/docs/api/webhooks/#verify-webhook-events) accordingly.

This is not urgent as the Zoom functionality is currently unused (no virtual fairs anymore) but this issue should still be fixed for correctness' sake (and also if there is another pandemic this could be useful). In addition, Penn's Zoom service manager gets periodic emails about fixing the "failing verification" error so this would be good to do.

PR #876, opened 2026-07-15
Fix Zoom webhook verification to use new secret token + HMAC scheme

ajboyle806:
The old verification token (Authorization header) was deprecated by Zoom and sunset in June 2025. Replace it with the new HMAC-SHA256 signature scheme using x-zm-signature and x-zm-request-timestamp headers.

Also add handling for the endpoint.url_validation challenge-response check (CRC) that Zoom sends on initial setup and every 72 hours to revalidate the webhook endpoint. This fixes the periodic 'failing verification' emails sent to Penn's Zoom service manager.

Settings: ZOOM_VERIFICATION_TOKEN -> ZOOM_WEBHOOK_SECRET_TOKEN

Fix to #843


Porcupine1 (approved):
thanks! can merge after the token is added. platform might help with this

PR #883, opened 2026-09-08
Keep the status banner and navbar pinned together on scroll

### LINKED ISSUE #881: Bugged UI for Status bar and Navbar
# Covers half of the search bar

Image

# since status bar does not scroll theres a gap that remains at the top

Image

InsideousBot:
## Problem

The Penn Labs status banner and the navbar come apart as soon as you scroll.

The banner isn't rendered by this repo. `frontend/pages/_document.tsx` loads `https://status.pennlabs.org/banner.js`, which injects

```html
<div id="pennlabs-status-banner" style="position: relative; ...">
```

as the **first child of `<body>`** — before the Next.js root, in normal document flow.

Meanwhile `Nav` in `frontend/components/Header/index.tsx` is `position: fixed` with **no `top` declaration**, so it falls back to its static position. The in-flow banner pushes that static position down by the banner's height, and a fixed box never re-resolves it on scroll.

The result: the banner scrolls away, the navbar stays pinned ~39px down, and page content scrolls through the gap above it. Measured on `master` at `scrollY: 600` — banner at `top: -600`, navbar still at `top: 39.2`.

The same root cause clipped the `/clubs` filter sidebar, which pins to the viewport independently of the navbar.

## Fix

Pin the banner to the top of the viewport and let the navbar sit directly beneath it, so the two behave as one block at any scroll position.

The banner's height can't be hardcoded — it's third-party, injected asynchronously after hydration, dismissible, and wraps to multiple lines on narrow viewports (39px at 1440px wide, 76px at 375px). So `frontend/hooks/useStatusBannerHeight.ts` measures it and publishes the value as a `--status-banner-height` CSS variable on the document root:

- a `MutationObserver` on `document.body` catches the async insertion and the removal when the user dismisses it
- a `ResizeObserver` tracks line-wrap changes
- writes are deduped, so it can't trip the "ResizeObserver loop" error

Everything that anchors to the navbar consumes it through a new `STATUS_BANNER_HEIGHT` constant in `constants/measurements.ts`:

```ts
export const STATUS_BANNER_HEIGHT = 'var(--status-banner-height, 0px)'
```

The `0px` fallback matters — it's a plain string baked into the styled-components CSS at module-eval time, identical on server and client, so there's no hydration mismatch and **the layout is unchanged when no banner is present** (no banner, adblock, script 404, or dismissed).

Consumers updated:

| File | Change |
|---|---|
| `components/Header/index.tsx` | `Nav` gets `top`; `NavSpacer` and `ImageHead` grow by the offset |
| `components/SearchBar.tsx` | sidebar `Wrapper`, mobile `Content`, `MobileToggle` |
| `renderPage.tsx`, `components/BaseLayout.tsx`, `components/common/Container.tsx` | `min-height: calc(100vh - NAV_HEIGHT)` |

Those three `min-height` sites aren't incidental: pinning the banner takes it out of flow, so without subtracting it every page would gain a ~39px scrollbar.

The hook also lowers the banner's `z-index` from the script's `99999` to `1001`. Without that, a pinned banner would float over every modal's shade — a pre-existing bug that only showed at `scrollY: 0`, now fixed as a side effect.

## Verification

`tsc --noemit` and eslint clean. Measured with Playwright against the real banner script (there's a live `degraded` incident, so it loads on localhost too):

| Case | Banner | Navbar | Gap |
|---|---|---|---|
| Desktop, scrolled | 0–39.2 (fixed) | 39–91 | ~0 |
| `/clubs`, scrolled 600 | 0–39.2 | 39–91 | ~0 |
| Mobile 375, scrolled | 0–75.6 (3 lines) | 76–128 | ~0 |
| Dismissed via `×` | gone, var → `0px` | 0–52 | — |

Also checked: content still starts below the navbar at `scrollY: 0`, the `/clubs` sidebar Search field clears the navbar, and dismissing the banner mid-scroll doesn't jump the navbar.

Not exercised locally: the `fyh` hub build (`ImageHead` / `LogoBackground` render only when `SITE_ID === 'fyh'`). Those offsets were updated for consistency and are covered by the same `0px` fallback, but a hub reviewer may want to eyeball them.

## Alternative considered

Making the navbar `position: sticky` and letting the banner scroll away needs no measurement at all, and is the more common announcement-bar pattern. Rejected because the desired behavior here is for the banner to stay visible while scrolled — during an incident, that's the point of it.

---

https://claude.ai/code/session_01Hzzw8TjPF98orwVN3STPpZ

### DIFF
--- a/frontend/constants/measurements.ts
+++ b/frontend/constants/measurements.ts
@@ -50,6 +50,13 @@ export const CARD_HEADING = MEASUREMENTS[SITE_ID].CARD_HEADING
 export const BANNER_HEIGHT = MEASUREMENTS[SITE_ID].BANNER_HEIGHT
 export const FULL_NAV_HEIGHT = MEASUREMENTS[SITE_ID].FULL_NAV_HEIGHT
 
+/**
+ * Height of the Penn Labs status banner injected by the script in
+ * `pages/_document.tsx`, published by `useStatusBannerHeight`. Falls back to
+ * `0px` when no banner is present, which is the default layout.
+ */
+export const STATUS_BANNER_HEIGHT = 'var(--status-banner-height, 0px)'
+
 export const BORDER_RADIUS = '4px'
 export const BORDER_RADIUS_LG = '8px'
 
--- a/frontend/components/BaseLayout.tsx
+++ b/frontend/components/BaseLayout.tsx
@@ -5,7 +5,7 @@ import styled from 'styled-components'
 import AuthPrompt from '~/components/common/AuthPrompt'
 import Footer from '~/components/Footer'
 import Header from '~/components/Header'
-import { NAV_HEIGHT, SNOW } from '~/constants'
+import { NAV_HEIGHT, SNOW, STATUS_BANNER_HEIGHT } from '~/constants'
 import { RenderPageWrapper, ToastStyle } from '~/renderPage'
 import { PermissionsContext } from '~/utils'
 import { createBasePropFetcher } from '~/utils/getBaseProps'
@@ -17,7 +17,7 @@ type BaseLayoutProps = Awaited<
 }
 
 export const Wrapper = styled.div`
-  min-height: calc(100vh - ${NAV_HEIGHT});
+  min-height: calc(100vh - ${NAV_HEIGHT} - ${STATUS_BANNER_HEIGHT});
   background: ${SNOW};
 `
 
--- a/frontend/components/Header/index.tsx
+++ b/frontend/components/Header/index.tsx
@@ -3,6 +3,7 @@ import { useRouter } from 'next/router'
 import { ReactElement, useEffect, useState } from 'react'
 import styled from 'styled-components'
 
+import { useStatusBannerHeight } from '~/hooks/useStatusBannerHeight'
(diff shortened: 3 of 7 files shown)


frontend/constants/measurements.ts
```diff
@@ -50,6 +50,13 @@ export const CARD_HEADING = MEASUREMENTS[SITE_ID].CARD_HEADING
 export const BANNER_HEIGHT = MEASUREMENTS[SITE_ID].BANNER_HEIGHT
 export const FULL_NAV_HEIGHT = MEASUREMENTS[SITE_ID].FULL_NAV_HEIGHT
 
+/**
+ * Height of the Penn Labs status banner injected by the script in
+ * `pages/_document.tsx`, published by `useStatusBannerHeight`. Falls back to
+ * `0px` when no banner is present, which is the default layout.
+ */
```
Porcupine1, line 57: leaving this here, but applies across the entire repo and future PRs:

I appreciate Claude's explanation, but note to keep comments/documentation concise and clear, only accepting verbosity where you think it's necessary

Porcupine1 (approved):
thanks for taking this up @InsideousBot!
lgtm
left a comment. should be good to go once addressed

PR #885, opened 2026-09-08
Club subscription groups (#867)

InsideousBot:
Closes #867.

Replaces the one-click subscribe button with configurable **subscription forms**. A `SubscriptionGroup` is an ordered set of questions a student answers; submitting subscribes them to the club, and the club gets source attribution plus listserv-ready exports.

As #867 puts it, subscriptions today are "a button that students press … after which their email is incremented onto a list", with no source attribution and no way to collect anything else. This makes them useful enough that clubs might actually rely on them.

## What's in it

**Models** — `SubscriptionGroup`, `SubscriptionQuestion`, `SubscriptionMultipleChoice`, `SubscriptionSubmission`, `SubscriptionQuestionResponse`, plus `Club.default_subscription_group` and `Subscribe.subscription_group`.

**Officer UI** — a Subscriptions tab on the club edit page with four panes: Overview (settings, set-default, archive), Questions, Subscribers, Distribution.

**Student-facing** — a public form page and a magic-link landing page. When a form asks nothing beyond auto-collected profile info, opening its link subscribes the user outright instead of showing an empty form.

**Distribution** — signed magic links and QR codes, for flyers and activities fair booths.

**Exports** — six shapes: CSV, TSV, plain email list, Mailchimp, Google Groups, and XLSX. Mailchimp and Google Groups are CSV files carrying the exact headers those tools expect, so there's no column-mapping step on import.

**Attribution** — every subscription records where the student came from (club page, search results, activities fair, magic link).

**Soft delete** — archiving hides a form and stops new submissions while retaining subscriber data, per the issue's requirement that deletion be visual only when submissions exist.

## Design decisions worth reviewing

**Defaults are opt-in.** A club keeps one-click subscribe until an officer explicitly sets a default form. The issue's wording — *"if one is set, hitting the subscription button would redirect to the form instead of instantly subscribing"* — reads as opt-in, and the alternative would silently change subscribe behaviour for every club on the platform.

**`?fmt=` selects the export shape, not `?format=`.** DRF reserves `format` for renderer content negotiation, so `?format=csv` would fail negotiation before reaching the view. `?format=xlsx` is left to DRF's `XLSXRenderer`, matching how `clubs-subscription` already serves spreadsheets.

**Magic links use `DEFAULT_DOMAIN` in production.** A printed QR code has to stay valid. Under `DEBUG` the link instead follows the calling frontend's origin so it's usable on localhost — but only if that origin is already in `CSRF_TRUSTED_ORIGINS`, so a caller can't aim links at an arbitrary host.

**`workflows.py` is shared, not speculative.** `BaseFormWorkflow` holds the submit-and-respond sequence. `ClubApplication` is the other `FormDefinitionBehaviorMixin` implementor and an `ApplicationFormWorkflow` subclass is the natural follow-up. #867 floated a shared abstract *model* for both; that would mean migrating `ClubApplication`'s table against a 5,800-line suite for no user-facing gain, so this shares the behaviour instead.

## Reuse

The officer UI is built on what already exists rather than new components: `ModelForm` (drag-reorder, confirm-delete), `common/Table`, `BaseCard`, `QRCodeCard`, `CreatableMultipleSelectField`, and `XLSXFormatterMixin` on the backend.

`StyledHeader`, `FormWrapper`, `ModalContainer` and `ScrollWrapper` were duplicated verbatim between `ApplicationsPage` and the new page; they now live in one `FormManagementStyles.ts` that both import. `QRCodeCard` gained optional `apiPath` / `title` props so it can serve an endpoint outside the `/{type}/{id}/qr` convention — existing club and ticket usages are unchanged.

## Testing

**454 tests pass** (full suite, run as CI runs it). 97 are new — 60 integration, 37 unit.

Integration covers officer permissions on every endpoint, the entry/instant-vs-form decision, submission (auth, required questions, closed windows, idempotency, attribution), question and multiple-choice CRUD, reordering, all six export formats (content type, filename, header row, exact data row), and magic-link round trips including tampered tokens.

Unit tests cover `is_currently_open` across the window boundaries, `auto_subscribe_eligible`, the workflow contract, and `ClubApplication` as the mixin's other implementor.

The frontend has no component-test framework (`"test"` is `tsc --noemit`), so verification there is typecheck, ESLint, and a production build — all clean.

## Migrations

Rebased onto current `master`, so the three migrations are numbered `0144`–`0146` and chain off `0143_split_club_reject_permission`. The branch is exactly one commit ahead of `master` with nothing behind.

`0146_cleanup_auto_subscription_groups` is data-only and reversible. It removes `SubscriptionGroup` rows named `"Subscribe"` that have **zero questions and zero submissions** — artifacts of an earlier iteration that auto-created a default form on every club save. Any form with real content is untouched, and `Club.default_subscription_group` is cleared before deletion.

## Note for reviewers

`populate.py` includes one line unrelated to this feature — `Club.objects.update(visible_to_public=True)` at the end of the seed command, so the anonymous club list matches `test_club_list_filter`. It predates this work and is carried along rather than dropped; happy to split it out.

## Not included

- Mailchimp OAuth sync and mailing-list sending — both marked TBD in #867 pending an SES budget discussion. The Mailchimp *export* is here; the API integration is not.
- Anonymous (logged-out) submissions. Submitting requires login; the magic-link page routes logged-out users through the standard `AuthPrompt`. True anonymous subscribe would need a nullable `Subscribe.person`, email verification, and rate limiting.

https://claude.ai/code/session_015BxU89wnYsLpU14zTHcusY

### DIFF
--- a/backend/clubs/admin.py
+++ b/backend/clubs/admin.py
@@ -55,6 +55,11 @@ from clubs.models import (
     Status,
     StudentType,
     Subscribe,
+    SubscriptionGroup,
+    SubscriptionMultipleChoice,
+    SubscriptionQuestion,
+    SubscriptionQuestionResponse,
+    SubscriptionSubmission,
     Tag,
     TargetMajor,
     TargetSchool,
@@ -693,3 +698,8 @@ admin.site.register(Type, TypeAdmin)
 admin.site.register(Status, StatusAdmin)
 admin.site.register(Classification, ClassificationAdmin)
 admin.site.register(GroupActivityOption, GroupActivityOptionAdmin)
+admin.site.register(SubscriptionGroup)
+admin.site.register(SubscriptionQuestion)
+admin.site.register(SubscriptionMultipleChoice)
+admin.site.register(SubscriptionSubmission)
+admin.site.register(SubscriptionQuestionResponse)
--- a/backend/clubs/management/commands/populate.py
+++ b/backend/clubs/management/commands/populate.py
@@ -1188,4 +1188,8 @@ class Command(BaseCommand):
                         Ticket.objects.create(showing=showing, type="Premium", price=i)
                     )
 
+        # Anonymous club list API filters to visible_to_public;
+        # match test_club_list_filter.
+        Club.objects.update(visible_to_public=True)
+
         self.stdout.write("Finished populating database!")
--- a/backend/clubs/migrations/0144_subscription_group_models.py
+++ b/backend/clubs/migrations/0144_subscription_group_models.py
@@ -1,0 +1,246 @@
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("clubs", "0143_split_club_reject_permission"),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
+
+    operations = [
+        # 1. SubscriptionGroup (no FK to Club yet — Club will gain the FK later)
+        migrations.CreateModel(
+            name="SubscriptionGroup",
+            fields=[
+                (
+                    "id",
+                    models.AutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("name", models.CharField(blank=True, max_length=255)),
+                ("description", models.TextField(blank=True)),
+                ("is_active", models.BooleanField(default=False)),
+                ("opens_at", models.DateTimeField(blank=True, null=True)),
+                ("closes_at", models.DateTimeField(blank=True, null=True)),
+                ("created_at", models.DateTimeField(auto_now_add=True)),
+                ("updated_at", models.DateTimeField(auto_now=True)),
+                (
+                    "club",
+                    models.ForeignKey(
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name="subscription_groups",
+                        to="clubs.club",
+                    ),
+                ),
+            ],
+        ),
+        # 2. SubscriptionQuestion
+        migrations.CreateModel(
+            name="SubscriptionQuestion",
+            fields=[
+                (
+                    "id",
+                    models.AutoField(
+                        auto_created=True,
+                        primary_key=True,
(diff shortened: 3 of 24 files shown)


Porcupine1 (commented):
thanks for taking of this @InsideousBot.
had codex review your changes. have a look at these in the meantime

<html><head></head><body>

Severity | Location (file:line) | Finding
-- | -- | --
P1 | [backend/clubs/serializers.py (line 4019)](/Users/lingo/Documents/PennLabs/penn-clubs/backend/clubs/serializers.py:4019) | Editing questions erases answers. Saving a multiple-choice question deletes and recreates its options—even for a wording-only edit. Existing selections become NULL, losing answers from subscriber details and exports. Preserve existing option IDs.
P2 | [backend/clubs/views.py (line 5613)](/Users/lingo/Documents/PennLabs/penn-clubs/backend/clubs/views.py:5613) | Required questions accept blank answers. Validation checks question IDs, not answer contents. Submitting the normal UI with empty required text and unselected choices succeeds.
P2 | [backend/clubs/serializers.py (line 1554)](/Users/lingo/Documents/PennLabs/penn-clubs/backend/clubs/serializers.py:1554) | Default forms can subscribe students to the wrong club. Club A’s officer can assign club B’s form as A’s default. Clicking Subscribe on A then creates a subscription to B. Validate that the default form belongs to the edited club.
P2 | [backend/clubs/serializers.py (line 4197)](/Users/lingo/Documents/PennLabs/penn-clubs/backend/clubs/serializers.py:4197) | Answers accept unrelated questions and choices. Submissions can reference another club’s questions or another question’s options. These inconsistent answers are saved and displayed in subscriber details.
P2 | [backend/clubs/serializers.py (line 4116)](/Users/lingo/Documents/PennLabs/penn-clubs/backend/clubs/serializers.py:4116) | Students don’t see the saved question order. The editor updates precedence, but public forms never sort by it. Dragging questions or instructions into order has no effect on the student-facing form.
P2 | [backend/clubs/views.py (line 1633)](/Users/lingo/Documents/PennLabs/penn-clubs/backend/clubs/views.py:1633) | Debug logging adds database work to every club listing. Three extra count queries execute before attempting to write to a hardcoded developer log path. Remove this block from the directory request path.

</body></html>

PR #886, opened 2026-09-08
Backfill majors missing from the Penn catalog

InsideousBot:
The `Major` table has never been synced against Penn's course catalog, so it has drifted well behind. Most visibly, **none of the Artificial Intelligence programs exist** — students in the Singh Program can't select their own major, and clubs can't target them.

This adds the 184 catalog programs that are missing. It changes nothing else.

## What's added

Comparing `catalog.upenn.edu/programs` (641 programs) against the table (615 unique names), 184 are genuinely absent. Among them:

```
Artificial Intelligence, BSE
Artificial Intelligence, MSE
Artificial Intelligence, Certificate
Artificial Intelligence for Business, BS
Artificial Intelligence for Business, MBA
Data Science and Artificial Intelligence, MSE
Learning Analytics and Artificial Intelligence, MSEd
Biology, BA
Architecture, MEBD
Applied Economics and Data Science, MEDS
…
```

Names use the catalog's own `Program, DEGREE` form, which is what the existing rows already use.

## Why it's additive only

`Major` is M2M'd to `Profile.major` and reached through `TargetMajor`, whose FK is `on_delete=CASCADE`. Deleting a row would silently strip that major from every student profile **and** delete club targeting rules, with no audit trail. So this migration only inserts, and its reverse is a deliberate `noop` rather than a delete.

It's also idempotent — it filters against existing names before `bulk_create`, so re-running is a no-op.

## On duplicates

`Major.name` has no unique constraint (`name = models.TextField()`, no index), so nothing at the database level prevents a duplicate. That matters here because the catalog and the table disagree on where a concentration goes:

```
catalog:      Africana Studies, BA: African American Studies
this table:   Africana Studies: African American Studies, BA
```

A naive diff reports ~100 false additions from that alone. The candidate list was normalised to `(program, concentration, degree)` before comparing, and a second pass caught 13 further clashes in programs whose *names* contain commas — for example `Operations, Information & Decisions, BS: General Track`, which already exists as `Operations, Information & Decisions: General Track, BS`. Those 13 were excluded too.

Net: 197 raw candidates → 184 after removing everything already present in another form. A test asserts no duplicate names exist after the migration runs.

## What this deliberately does not do

- **No renames.** ~171 existing rows use the older concentration convention or reflect programs the catalog has since renamed. Realigning them would rewrite rows that student profiles already point at, so it's left alone.
- **No deletions.** Retired programs stay, for the cascade reasons above.
- **No automated sync.** The catalog has no JSON feed and returns `HTTP 202` with an empty body to non-browser agents, so any recurring sync means scraping HTML with a spoofed User-Agent — fragile enough to want its own design discussion. A dry-run monitor that reports drift without writing would be the safer next step.

## Testing

12 new tests in `tests/clubs/test_majors.py` covering: the list is non-empty, internally duplicate-free, and well-formed; a freshly migrated database contains exactly these rows; re-running adds nothing; only genuinely missing entries are inserted; existing majors are neither renamed nor deleted; and `Profile.major` / `TargetMajor` relations survive a re-run.

Full suite passes at 369.

## Note on merge order

This is numbered `0144`, branched off current `master`. #885 also adds `0144`–`0146`. Whichever merges second needs its migrations renumbered — happy to rebase this one if #885 lands first.

https://claude.ai/code/session_015BxU89wnYsLpU14zTHcusY


backend/clubs/migrations/0144_add_missing_majors.py
```diff
@@ -0,0 +1,225 @@
+from django.db import migrations
+
+
+# Programs listed in the Penn course catalog (catalog.upenn.edu/programs) that
+# were absent from the Major table. Penn adds programs each year — notably the
```
Porcupine1, line 5: pretty sure this isn't true. should rarely happen

Porcupine1 (commented):
gabe left a great review on slack
@gabeweng wanna leave it here too for git history?

backend/clubs/migrations/0144_add_missing_majors.py
gabeweng: last few years in particular have had substantial changes in degrees, majors, and concentrations (additions, renames), so we should probably just be on the lookout for any changes to the catalog if Penn continues to adapt to workforce demands. let's assume that majors, concentrations, and even degrees could change on a semesterly basis.

gabeweng:
I like the normalization approach that we take; it's very helpful that you recognized that the current Major table follows an older convention from the catalog than the current catalog. I'll highlight potential areas of concern with the list that you ended up with.

Could we get a full list of

1. Majors on Website, not on Penn Clubs, alongside the normalized tuples
2. Majors on Penn Clubs, not on Website, alongside the normalized tuples


Then once we see what we have, we can determine next steps, but my preliminary thoughts are 

A) Just rename majors when they are really the same thing and people already use the old names on Penn Clubs
B) For majors that don't exist anymore, like NETS or the Biology concentrations, just have a field like active or listed

backend/clubs/migrations/0144_add_missing_majors.py
```diff
@@ -0,0 +1,225 @@
...
+    "Architecture, MEBD",
+    "Architecture, MSD: Property Development and Design",
+    "Architecture, MSD: Robotics and Autonomous Systems",
+    "Artificial Intelligence for Business, BS",
+    "Artificial Intelligence for Business, MBA",
+    "Artificial Intelligence, BSE",
+    "Artificial Intelligence, Certificate",
+    "Artificial Intelligence, MSE",
+    "Biochemistry, Biophysics and Chemical Biology, PhD",
+    "Biology, BA",
```
gabeweng: Biology no longer has concentrations so while we are adding Biology, BA, we still would have all the concentrations on Penn Clubs. We shouldn't delete them, but consider if we actually need to do anything about them (maybe add a new field that says if a major is active or not)

backend/clubs/migrations/0144_add_missing_majors.py
```diff
@@ -0,0 +1,225 @@
...
+    "Nutrition Science, MSNS",
+    "Nutrition, Certificate",
+    "One Health, Certificate",
+    "Oral Biology, MSOB",
+    "Oral Health Sciences, MOHS",
+    "Oral and Population Health, MSOPH",
+    "Pediatric Dentistry, Certificate",
+    "Periodontics and Prosthodontics, Certificate",
+    "Philosophy, BA: Moral and Political Philosophy",
+    "Philosophy, BA: Philosophy of Science",
```
gabeweng: Philosophy: Political and Moral Philosophy, BA got renamed to Philosophy: Moral and Political Philosophy, BA

Philosophy, BA: Philosophy and Science got renamed to Philosophy, BA: Philosophy of Science

backend/clubs/migrations/0144_add_missing_majors.py
```diff
@@ -0,0 +1,225 @@
...
+    "Middle Eastern Languages & Cultures, PhD: Egyptology",
+    "Middle Eastern Languages & Cultures, PhD: Hebrew and Judaic Studies",
+    "Middle Eastern Languages & Cultures, PhD: Mesopotamian Civilization",
+    "Middle Eastern Languages & Cultures, PhD: Middle Eastern Literatures & Societies",
+    "Middle Eastern Languages & Cultures: Ancient Middle East, Minor",
+    "Middle Eastern Languages & Cultures: Arabic & Hebrew Studies, Minor",
+    "Middle Eastern Languages & Cultures: Arabic Language and Literature, Minor",
+    "Middle Eastern Languages & Cultures: Cultures and Societies of the Middle East and North Africa, Minor",
+    "Middle Eastern Languages & Cultures: Hebrew & Judaic Studies, Minor",
+    "Middle Eastern Languages & Cultures: Persian Language & Literature, Minor",
```
gabeweng: Should consider just renaming the existing 'Near Eastern Languages and Civilizations" majors

backend/clubs/migrations/0144_add_missing_majors.py
```diff
@@ -0,0 +1,225 @@
...
+    "Urban Education, MSEd",
+    "Visual and Interactive Computing, BSE",
+    "Writing, BAAS",
+    "Yiddish, Minor",
+]
+
+
+def add_missing_majors(apps, schema_editor):
+    """
+    Purely additive. Existing rows are never renamed or deleted: Major is
```
gabeweng: We might want to consider renaming. In many cases it could be better than having two versions of the same major

backend/clubs/migrations/0144_add_missing_majors.py
```diff
@@ -0,0 +1,225 @@
...
+    referenced by Profile.major (M2M) and by TargetMajor with on_delete=CASCADE,
+    so removing a row would silently strip it from student profiles and delete
+    club targeting rules.
+    """
+    Major = apps.get_model("clubs", "Major")
+    existing = set(Major.objects.values_list("name", flat=True))
+    Major.objects.bulk_create(
+        [Major(name=name) for name in MISSING_MAJORS if name not in existing]
```
gabeweng: good

